PR #1, opened 2015-08-19
Working Plugin

alexhersh:


Makefile
```diff
@@ -1,10 +1,19 @@
...
 
-binary: dist/calico
+BUILD_DIR=build_calico_rkt
+BUILD_FILES=Dockerfile requirements.txt
 
-dist/calico:
-	# Build docker container
-	cd build_calico_rkt; docker build -t calico-rkt-build .
+default: all
+all: binary test
```
TrimBiggs: binary isn't required here.  You can follow the path of test and see that it makes sure the binary is installed before continuing.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
-import sys
+#!/usr/bin/env python
```
TrimBiggs: This should have the Metaswitch copyright at the top.  Also I don't think this line here is required.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+input_json = json.loads(input_)
+
+# Append to existing env, to avoid losing PATH etc.
+# TODO-PAT: This shouldn't be hardcoded
+env = os.environ.copy()
+ETCD_AUTHORITY_ENV = 'ETCD_AUTHORITY'
+env[ETCD_AUTHORITY_ENV] = 'localhost:2379'
+
+# PROFILE_LABEL = 'CALICO_PROFILE'
+# ETCD_PROFILE_PATH = '/calico/'
```
TrimBiggs: Is this still needed?

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+from pycalico import netns
+from pycalico.ipam import IPAMClient, SequentialAssignment
+from pycalico.netns import Namespace
+from pycalico.datastore_datatypes import Rules
+from pycalico.datastore import IF_PREFIX, DatastoreClient
+from pycalico.datastore_errors import PoolNotFound
+
+print_stderr = functools.partial(print, file=sys.stderr)
+
+
```
TrimBiggs: Extra space here.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
 def main():
-    stdin_raw_data = sys.stdin.read()
+    mode = env['CNI_COMMAND']
+
+    if mode == 'init':
+        print_stderr('No initialization work to perform')
+    elif mode == 'ADD':
+        print_stderr('Executing Calico pod-creation plugin')
+        create(
+            container_id=env['CNI_CONTAINERID']
```
TrimBiggs: This could just be one line as `create(...)`

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+        sys.exit(1)
 
+    # Allocate and Assign ip address through IPAM Client
+    ip = _allocate_IP()
 
-def isolate(args):
+    # Create Endpoint object
+    try:
+        ep = client.create_endpoint(HOSTNAME, ORCHESTRATOR_ID,
+                                      container_id, [IPAddress(ip)])
```
TrimBiggs: Misalignment

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+    print_stderr('Configuring Calico networking.', file=sys.stderr)
+
+    interface = env['CNI_IFNAME']
+
+    endpoint, ip = _container_add(hostname=HOSTNAME,
+                                orchestrator_id=ORCHESTRATOR_ID,
+                                container_id=container_id,
+                                netns_path=netns_path,
+                                interface=interface,
+                                client=client)
```
TrimBiggs: These params aren't quite aligned with the `hostname` param.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+    Add a container to Calico networking with the given IP
+    Return Endpoint object
+    """
     try:
-        args = stdin_json['args']
+        _ = client.get_endpoint(
+            hostname=hostname,
+            orchestrator_id=orchestrator_id,
+            workload_id=container_id
+        )
```
TrimBiggs: This alignment isn't quite right, but I don't think the alignment should be done like this anyway.

I think a more *pythonic* way of writing this is to have the first param on the same line as the call (like line 98 with `_container_add`, then aligning the other params with the first:
```
func_call(param_1,
          param_2,
          param_3)
```

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
-        "container-id": "ba11f1de-fc4d-46fd-9f15-424f4ef05a3a", # Required
-        "pid": 3789 # Required
-    }
+    Remove the indicated container on this host from Calico networking
     """
-    pass
+    # Find the endpoint ID. We need this to find any ACL rules
+    try:
+        endpoint = client.get_endpoint(
+            hostname=hostname,
```
TrimBiggs: See my previous comment about alignment

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
 
+    else:
+        client.create_profile(profile_name)
+        # _apply_rules(profile_name, client)
+
+    # Also set the profile for the workload.
+    print_stderr('Setting profile %s on endpoint %s' %
+                 (profile_name, endpoint.endpoint_id))
+    client.set_profiles_on_endpoint(
+        [profile_name], endpoint_id=endpoint.endpoint_id
```
TrimBiggs: I'd style this like I mentioned above.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+env = os.environ.copy()
+ETCD_AUTHORITY_ENV = 'ETCD_AUTHORITY'
+env[ETCD_AUTHORITY_ENV] = 'localhost:2379'
+
+# PROFILE_LABEL = 'CALICO_PROFILE'
+# ETCD_PROFILE_PATH = '/calico/'
+
+ORCHESTRATOR_ID = "rkt"
+HOSTNAME = socket.gethostname()
+NETNS_ROOT= '/var/lib/rkt/pods/run'
```
TrimBiggs: `NETNS_ROOT =`

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+        print_stderr('No initialization work to perform')
+    elif mode == 'ADD':
+        print_stderr('Executing Calico pod-creation plugin')
+        create(
+            container_id=env['CNI_CONTAINERID']
+            )
+    elif mode == 'DEL':
+        print_stderr('Executing Calico pod-deletion plugin')
+        delete(
+            container_id=env['CNI_CONTAINERID']
```
TrimBiggs: See above

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+            )
+
+def create(container_id):
+    """"Handle rkt pod-create event."""
+    print_stderr('Configuring pod %s' % container_id, file=sys.stderr)
```
TrimBiggs: I don't think the `file` param is necessary here since the `print_stderr` func has it baked in.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
-    # Convert input data to JSON object
     try:
-        stdin_json = json.loads(stdin_raw_data)
-    except ValueError as e:
-        quit_with_error(str(e))
-
-    # Extract command
+        endpoint, ip = _create_calico_endpoint(container_id=container_id,
+                                            netns_path=netns_path,
+                                            client=_datastore_client)
```
TrimBiggs: Misaligned params

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+
+        _create_profile(endpoint=endpoint, 
+                        profile_name=input_json['name'], 
+                        ip=ip,
+                        client=_datastore_client)
+    except CalledProcessError as e:
+        print_stderr('Error code %d creating pod networking: %s\n%s' % (
```
TrimBiggs: Just stick this paren on the next line with the rest of it and align it to the string.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
     """
-    pass
+    print_stderr('Configuring Pod Profile: %s' % profile_name)
 
+    if client.profile_exists(profile_name):
+        print_stderr("Profile with name %s already exists, applying to endpoint." % (profile_name))
 
+    else:
+        client.create_profile(profile_name)
+        # _apply_rules(profile_name, client)
```
TrimBiggs: Can be removed?

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+    Currently assumes one pod with each name.
     """
-    pass
+    print_stderr('Configuring Pod Profile: %s' % profile_name)
 
+    if client.profile_exists(profile_name):
+        print_stderr("Profile with name %s already exists, applying to endpoint." % (profile_name))
 
+    else:
+        client.create_profile(profile_name)
```
TrimBiggs: With the printing you have, it would make sense to have a line saying the profile is being created here.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+                "action": "allow",
+            },
+        ],
+    }
+    rules_json = json.dumps(rules_dict, indent=2)
+    rules = Rules.from_json(rules_json)
+    return rules
 
-def quit_with_error(msg=None):
+def _apply_rules(profile_name, client):
```
TrimBiggs: This isn't used anymore

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+                                netns_path=netns_path,
+                                interface=interface,
+                                client=client)
+    print_stderr('Finished configuring network interface', file=sys.stderr)
+    return endpoint, ip
+
+def _container_add(hostname, orchestrator_id, container_id, netns_path, interface, client):
+    """
+    Add a container to Calico networking with the given IP
+    Return Endpoint object
```
TrimBiggs: and the newly allocated ip

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+                                container_id=container_id,
+                                netns_path=netns_path,
+                                interface=interface,
+                                client=client)
+    print_stderr('Finished configuring network interface', file=sys.stderr)
+    return endpoint, ip
+
+def _container_add(hostname, orchestrator_id, container_id, netns_path, interface, client):
+    """
+    Add a container to Calico networking with the given IP
```
TrimBiggs: Looks like we're generating and allocating the IP

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+    """Cleanup after a pod."""
+    print_stderr('Deleting pod %s' % container_id, file=sys.stderr)
+
+    _datastore_client = IPAMClient()
+
+    # Remove the profile for the workload.
+    _container_remove(hostname=HOSTNAME,
+                        orchestrator_id=ORCHESTRATOR_ID,
+                        container_id=container_id,
+                        client=_datastore_client)
```
TrimBiggs: Alignment

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
+    except CalledProcessError as e:
+        print_stderr('Error code %d creating pod networking: %s\n%s' % (
+            e.returncode, e.output, e))
+        sys.exit(1)
+    print_stderr('Finished Creating pod %s' % container_id)    
+
+def delete(container_id):
+    """Cleanup after a pod."""
+    print_stderr('Deleting pod %s' % container_id, file=sys.stderr)
```
TrimBiggs: This should go at the end, if it's being deleted then the user probably already knows and just wants to know if it was deleted successfully. Make sure to add the `sys.exit` call in the except case when you do this.

calico_rkt/calico_rkt.py
```diff
@@ -1,91 +1,249 @@
...
     try:
-        command = stdin_json['command']
-    except KeyError:
-        quit_with_error("Missing command")
-
-    # Extract args
+        _datastore_client.remove_profile(input_json['name'])
+    except:
+        print_stderr("Cannot remove profile %s; Profile cannot be found." % container_id)
+
```
TrimBiggs: You should have a print at the bottom here for then the pod has been removed.  Make sure you add a `sys.exit` call above too when you do this!

calico_rkt/calico_rkt.py
alexhersh: I figured I would keep it as a placeholder if we decide to do anything with rules

TrimBiggs:
There's a decent number of prints to the screen in here, which is nice for debugging but probably not all that useful for the user.  Look through the functions and see if you can get each function down to one log (or two max), where the important information is printed, like "Created pod X", "Deleted pod X", "Configured pod X with Calico", etc.

calico_rkt/calico_rkt.py
TrimBiggs: Ok, with your comment above.

calico_rkt/calico_rkt.py
TrimBiggs: Ok, might want to put a TODO in here then.

PR #2, opened 2015-08-21
Update #2: UTs, formatting/cleanup, requirements, logger and README

alexhersh:
Working PR to fix UTs and add features

Also added in a few changes as per Tim's [comments](https://github.com/projectcalico/calico-rkt/pull/1)


caseydavenport:
Could you remove your .idea folder?  Shouldn't be checked in.

Makefile
```diff
@@ -1,16 +1,16 @@
...
 default: all
-all: binary test
+all: test
 binary: dist/calico_rkt
 test: ut
 
 # Build a new docker image to be used by binary or tests
 rktbuild.created: $(BUILD_FILES)
-	docker build -t calico/rkt-build .
+	cd build_calico_rkt; docker build -t calico/rkt-build .
```
caseydavenport: Maybe use $BUILD_DIR here?

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
 from pycalico import netns
 from pycalico.ipam import IPAMClient, SequentialAssignment
 from pycalico.netns import Namespace
-from pycalico.datastore_datatypes import Rules
+from pycalico.datastore_datatypes import Rules, IPPool
 from pycalico.datastore import IF_PREFIX, DatastoreClient
 from pycalico.datastore_errors import PoolNotFound
 
-print_stderr = functools.partial(print, file=sys.stderr)
+logger = logging.getLogger(__name__)
```
caseydavenport: We usually use _log as the name for our loggers - I know the Kube plugin does this, but we should change that one up too.

README.md
```diff
@@ -3,3 +3,27 @@
 ## Building the plugin
 To build the calico-kubernetes plugin, clone this repository and run `make`.  This will build the binary, as well as run the unit tests.  To just build the binary, with no tests, run `make binary`.  To only run the unit tests, simply run `make ut`.
 
+## Using rkt Plugins
+The [documentation](https://github.com/coreos/rkt/blob/master/Documentation/networking.md) for Network Plugins in rkt will walk you through the basics of setting up networking in rkt.
+
+## Requirements
+* A working [ETCD](https://github.com/coreos/etcd) service
```
djosborne: `etcd` should be all lowercase

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
-
-# Append to existing env, to avoid losing PATH etc.
-# TODO-PAT: This shouldn't be hardcoded
-env = os.environ.copy()
 ETCD_AUTHORITY_ENV = 'ETCD_AUTHORITY'
-env[ETCD_AUTHORITY_ENV] = 'localhost:2379'
 
-# PROFILE_LABEL = 'CALICO_PROFILE'
-# ETCD_PROFILE_PATH = '/calico/'
+env, INPUT_JSON = {}, {}
```
caseydavenport: Could be missing something, but I don't think you need to initialize these here - they get set in __main__

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
 
 ORCHESTRATOR_ID = "rkt"
 HOSTNAME = socket.gethostname()
-NETNS_ROOT= '/var/lib/rkt/pods/run'
+NETNS_ROOT = '/var/lib/rkt/pods/run'
 
 def main():
     mode = env['CNI_COMMAND']
 
     if mode == 'init':
```
caseydavenport: I thought there was no "init" mode for CNI - I think this can go.

README.md
```diff
@@ -3,3 +3,27 @@
...
+* Currently, Calico is capable of networking rkt containers, but our `calico/node` image does not yet work in rkt. For now, you will need to run our `calico/node` [docker image](https://github.com/projectcalico/calico-docker/blob/master/docs/getting-started/default-networking/Demonstration.md) in parallel to your rkt images.
```
djosborne: We might want to provide a little more detail on  how to run the calico/node docker container here. The page you link to is just a bland tutorial for starting a cluster. It might be simpler to just explain it here with something like:
> To start calico/node, download the calicoctl binary(link) and run the following command, using the IP on your host: `calicoctl node --ip=$IP`

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
 
-        _create_profile(endpoint=endpoint, 
-                        profile_name=input_json['name'], 
+        _create_profile(endpoint=endpoint,
+                        profile_name=INPUT_JSON['name'],
                         ip=ip,
                         client=_datastore_client)
     except CalledProcessError as e:
-        print_stderr('Error code %d creating pod networking: %s\n%s' % (
+        logger.info('Error code %d creating pod networking: %s\n%s' % (
```
caseydavenport: logs in exception blocks should use logger.exception(..) rather than logger.info(..).

In general, make sure to be thinking about logging levels for individual messages.  Some are info level, but some are only useful for debug level and might become repetitive / cluttered if at info level, etc.

README.md
```diff
@@ -3,3 +3,27 @@
...
+
+## Installing
+* Move the binary for this plugin to `/usr/lib/rkt/plugins/net/calico`.
+* Configure your network with a `*.conf` file in `/etc/rkt/net.d/`. This `conf` file should specify `calico` as `type`. For example,
```
djosborne: I'm not familiar with how rkt executes, but can we give the `.conf` file a static name, like `calico.conf`?

calico_rkt/calico_rkt.py
```diff
@@ -229,21 +238,52 @@ def _apply_rules(profile_name, client):
...
+
+    pool = IPPool(subnet)
+    version = IPNetwork(subnet).version
+
+    client.add_ip_pool(version, pool)
+    logger.info("Using Pool %s" % pool)
+
+    return pool
+
+def _allocate_IP(pool):
```
djosborne: function names should always be lower case: `_allocate_ip(pool):`

calico_rkt/calico_rkt.py
```diff
@@ -229,21 +238,52 @@ def _apply_rules(profile_name, client):
...
+    :return: The next avail address from the pool
+    :rtype IPAddress object
+    """
+    candidate = SequentialAssignment().allocate(pool)
+    logger.info("Using IP %s" % candidate)
+    return IPAddress(candidate)
 
 if __name__ == '__main__':
+    env = os.environ.copy()
+    env[ETCD_AUTHORITY_ENV] = 'localhost:2379'  if ETCD_AUTHORITY_ENV not in env.keys() else env[ETCD_AUTHORITY_ENV]
```
djosborne: There's an extra space before your if statement here. but otherwise, nice line

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
-        _datastore_client.remove_profile(input_json['name'])
-    except:
-        print_stderr("Cannot remove profile %s; Profile cannot be found." % container_id)
+                      orchestrator_id=ORCHESTRATOR_ID,
+                      container_id=container_id,
+                      client=_datastore_client)
+
+    profile_name = INPUT_JSON['name']
+
+    # Delete profile if only member
```
caseydavenport: I think this opens a race-condition if running a Calico + rkt deployment across multiple nodes.

E.g. One node creates a container in this network, at the same time another deletes the last one.  The new container could come up, and then its profile could be deleted on the other node.

This is a known issue in profile management, and probably requires libcalico work to make it safe, but worth knowing about.

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
-        )
+        _ = client.get_endpoint(hostname=HOSTNAME,
+                                orchestrator_id=ORCHESTRATOR_ID,
+                                workload_id=container_id)
     except KeyError:
         # Calico doesn't know about this container.  Continue.
         pass
     else:
-        print_stderr("This container has already been configured with Calico Networking.")
+        logger.info("This container has already been configured with Calico Networking.")
```
caseydavenport: This should be an `error` level log.

calico_rkt/calico_rkt.py
```diff
@@ -229,21 +238,52 @@ def _apply_rules(profile_name, client):
...
+    Take Input subnet (global), create IP pool in datastore
+    Will complete silently if it exists
+    return IPPool  object of subnet pool
     """
-    pool = input_json['ipam']['subnet']
-    candidate = SequentialAssignment().allocate(IPNetwork(pool))
-    print_stderr("Using IP %s" % candidate)
-    return candidate
+    try:
+        subnet = INPUT_JSON['ipam']['subnet']
```
djosborne: I'm not a big fan of pulling the global variable into this function. It'd be much cleaner and more easily testable if you pass it in as a parameter. You can still pass in the global variable - just do it where this is called. @caseydavenport agreed?

README.md
djosborne: I think saying `our calico/node image does not yet work in rkt` is unclear to someone not familiar with the calico packages. How about: 

>Though Calico is capable of networking rkt containers, our core software is distributed and deployed in a docker container. For now, you will need to launch calico by starting the [docker image](https://github.com/projectcalico/calico-docker/blob/master/docs/getting-started/default-networking/Demonstration.md) in parallel to your rkt images.

README.md
```diff
@@ -3,3 +3,27 @@
...
 
+## Using rkt Plugins
+The [documentation](https://github.com/coreos/rkt/blob/master/Documentation/networking.md) for Network Plugins in rkt will walk you through the basics of setting up networking in rkt.
+
+## Requirements
+* A working [ETCD](https://github.com/coreos/etcd) service
+* Currently, Calico is capable of networking rkt containers, but our `calico/node` image does not yet work in rkt. For now, you will need to run our `calico/node` [docker image](https://github.com/projectcalico/calico-docker/blob/master/docs/getting-started/default-networking/Demonstration.md) in parallel to your rkt images.
+
+## Installing
+* Move the binary for this plugin to `/usr/lib/rkt/plugins/net/calico`.
```
djosborne: Can we change this to:

* Install the plugin binary
```
sudo mv ./binary-name /usr/lib/rkt/plugins/net/calico
```

README.md
```diff
@@ -3,3 +3,27 @@
...
+{
+    "name": "network-name",
+    "type": "calico",
+    "ipam": {
+        "type": "host-local",
+        "subnet": "10.1.0.0/16",
+    }
+}
+```
+* When you spin up a node with `rkt run`, specify the `--private-net=network-name` flag to enable Calico Networking
```
djosborne: I think the correct rkt terminology is a "container", not a "node"

README.md
```diff
@@ -3,3 +3,27 @@
...
+
+## Networking Behavior
+In rkt deployments, Calico will allocate an available IP within the specified subnet pool and enforce the default Calico networking rules on pods. The default behavior is to allow traffic only from other containers in the network. For each network with a unique `"name"` parameter (as shown above), Calico will create a single profile that will be applied to each pod added to that network.
```
djosborne: "Container", not "pod"?

djosborne: Instead of `The default behavior is...`, how about `By default, Calico will...`

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
 
     if mode == 'init':
-        print_stderr('No initialization work to perform')
+        logger.info('No initialization work to perform')
     elif mode == 'ADD':
-        print_stderr('Executing Calico pod-creation plugin')
-        create(
-            container_id=env['CNI_CONTAINERID']
-            )
+        logger.info('Executing Calico pod-creation plugin')
```
djosborne: You've got a bit of a redundant logging message here, since you log a second message immediately in the `create` and `delete` functions. Though a lot of things can go wrong in python, we can assume that we'll make it to that next logger, so remove either this logging message or the one down there

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
+    profile_name = INPUT_JSON['name']
+
+    # Delete profile if only member
+    if _datastore_client.profile_exists(profile_name) and \
+       len(_datastore_client.get_profile_members(profile_name)) < 1:
+        try:
+            logger.info("Profile %s has no members, removing from datastore" % profile_name)
+            _datastore_client.remove_profile(profile_name)
+        except:
+            logger.info("ERROR: Cannot remove profile %s; Profile cannot be found." % container_id)
```
djosborne: We typically use colon's here, not semicolons

djosborne: Also, if this is an error, log it as one with `logger.error("Cannot remove profile %s.....`

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,149 @@
...
+    interface = env['CNI_IFNAME']
+
+    endpoint, ip = _container_add(hostname=HOSTNAME,
+                                  orchestrator_id=ORCHESTRATOR_ID,
+                                  container_id=container_id,
+                                  netns_path=netns_path,
+                                  interface=interface,
+                                  client=client)
+
+    logger.info('Finished configuring network interface')
```
djosborne: might be useful to print the IP and maybe some other information from the endpoint in this logging message

calico_rkt/calico_rkt.py
```diff
@@ -167,32 +175,30 @@ def _container_remove(hostname, orchestrator_id, container_id, client):
...
     # Remove the container from the datastore.
-    client.remove_workload(hostname, orchestrator_id, container_id)
+    client.remove_workload(hostname=hostname, 
+                           orchestrator_id=orchestrator_id, 
+                           workload_id=container_id)
 
-    print_stderr("Removed Calico interface from %s" % container_id)
+    logger.info("Removed Calico interface from %s" % container_id)
 
 def _create_profile(endpoint, profile_name, ip, client):
```
djosborne: This function primarily sets the profile of an endpoint, and optionally creates it as well. The function name should probably be `set_endpoint_profile`

calico_rkt/calico_rkt.py
```diff
@@ -201,9 +207,12 @@ def _create_profile(endpoint, profile_name, ip, client):
         })
     print(dump)
 
-def _create_rules(id_):
+def _create_rules(profile):
```
djosborne: Since these are the default rules for rkt (and they may change), lets rename this `_create_default_rules`

calico_rkt/calico_rkt.py
```diff
@@ -221,7 +230,7 @@ def _create_rules(id_):
 
 def _apply_rules(profile_name, client):
```
djosborne: How about `_apply_default_rules`

djosborne:
Nice work pulling a lot of moving parts together to get this working. Besides some minor adjustments to logging, and a few other naming nitpicks, I think the big thing you need to resolve is how you pass around your input variables. All of these calico plugins basic job is to convert raw input passed in from another program into calico function calls. I think the program flow should represent that - get rid of the global JSON, parse out all the info you need in the beginning, and then pass it in to the functions which require it as you go down. It will make for much maintainable code down the road.

calico_rkt/calico_rkt.py
djosborne: Is this behavior really necessary? I don't see why we can't leave the profile if every container is removed.

alexhersh:
I think I've addressed all of your comments. :D

README.md
```diff
@@ -3,3 +3,28 @@
 ## Building the plugin
 To build the calico-kubernetes plugin, clone this repository and run `make`.  This will build the binary, as well as run the unit tests.  To just build the binary, with no tests, run `make binary`.  To only run the unit tests, simply run `make ut`.
 
+## Using rkt Plugins
+The [documentation](https://github.com/coreos/rkt/blob/master/Documentation/networking.md) for Network Plugins in rkt will walk you through the basics of setting up networking in rkt.
+
+## Requirements
+* A working [etcd](https://github.com/coreos/etcd) service
+* Currently, Calico is capable of networking rkt containers, but our `calico/node` image does not yet work in rkt. For now, you will need to run our `calico/node` [docker image](https://github.com/projectcalico/calico-docker/blob/master/docs/getting-started/default-networking/Demonstration.md) in parallel to your rkt images.
```
djosborne: I think saying `our calico/node image does not yet work in rkt` is unclear to someone not familiar with the calico packages. How about: 

>Though Calico is capable of networking rkt containers, our core software is distributed and deployed in a docker container. For now, you will need to run calico in docker before starting your rkt contianers. This can be easily done by downloading the calicoctl binary(link) and running the following command: `calicoctl node --ip=$IP`

README.md
```diff
@@ -3,3 +3,28 @@
...
+	- To start calico/node, download the [calicoctl binary](https://github.com/projectcalico/calico-docker/releases) and run the following command, using the IP on your host: `calicoctl node --ip=$IP`
+
+## Installing
+* Move the binary for this plugin to `/usr/lib/rkt/plugins/net/calico`.
```
djosborne: Any thoughts on changing this to a move command?

README.md
```diff
@@ -3,3 +3,28 @@
...
+* Configure your network with a `*.conf` file in `/etc/rkt/net.d/`. This `conf` file should specify `calico` as `type`. For example,
```
djosborne: Any thoughts on the static conf file?

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,150 @@
...
 
-def create(container_id):
+def create(env, conf_in):
     """"Handle rkt pod-create event."""
-    print_stderr('Configuring pod %s' % container_id, file=sys.stderr)
+    container_id = env['CNI_CONTAINERID']
+
+    _log.info('Configuring pod %s' % container_id)
     netns_path='%s/%s/%s' % (NETNS_ROOT, container_id, env['CNI_NETNS'])
     _datastore_client = IPAMClient()
```
djosborne: Since you only every need a single copy of `datastore_client`, and since its used at least once every time this plugin is called, it makes sense to make it a global variable. We do that a lot to the client across pycalico/calico-docker.

To remedy this, move it up to the top of this page just below your imports, keep it lower case, remove the preceeding underscore, and pull it out of your function calls.

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,150 @@
...
-        print_stderr('Error code %d creating pod networking: %s\n%s' % (
+        _log.error('ERROR: error code %d creating pod networking: %s\n%s' % (
             e.returncode, e.output, e))
         sys.exit(1)
-    print_stderr('Finished Creating pod %s' % container_id)    
 
-def delete(container_id):
+    _log.info('Finished Creating pod %s' % container_id)
+
+def delete(env, conf_in):
```
djosborne: pep8 nitpick - functions should be seperated by 2 newlines

Makefile
```diff
@@ -1,16 +1,16 @@
 .PHONY: all binary ut clean
 
 BUILD_DIR=build_calico_rkt
-BUILD_FILES=Dockerfile requirements.txt
+BUILD_FILES=$(BUILD_DIR)/Dockerfile $(BUILD_DIR)/requirements.txt
```
Symmetric: Why have you moved things to build_calico_rkt? It's only necessary in the main repo because we have two Dockerfiles, and it makes things more complex.

Standard Python project structure has requirements.txt at the root, and it's normal to have Makefile and Dockerfile at the root too.

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,150 @@
...
+        create(env=env, conf_in=conf_in)
     elif mode == 'DEL':
-        print_stderr('Executing Calico pod-deletion plugin')
-        delete(
-            container_id=env['CNI_CONTAINERID']
-            )
+        delete(env=env, conf_in=conf_in)
 
-def create(container_id):
+def create(env, conf_in):
```
Symmetric: You don't really need to pass in `env`; since this file is being run once per action, keeping it as a global is fine. Minor nit though.

calico_rkt/logutils.py
```diff
@@ -0,0 +1,17 @@
...
+    if log_dir is None:
+        log_dir = LOG_DIR
+
+    if not os.path.exists(log_dir):
+        os.makedirs(log_dir)
+    hdlr = logging.FileHandler(filename=log_dir+'/calico.log')
+    formatter = logging.Formatter('%(asctime)s %(levelname)s %(message)s')
+    hdlr.setFormatter(formatter)
+    logger.addHandler(hdlr)
+    logger.setLevel(logging.INFO)
```
Symmetric: Missing \n

calico_rkt/logutils.py
```diff
@@ -0,0 +1,17 @@
...
+
+LOG_DIR = '/var/log/calico/rkt-plugin'
+
+def configure_logger(logger, log_dir=None):
+    if log_dir is None:
+        log_dir = LOG_DIR
+
+    if not os.path.exists(log_dir):
+        os.makedirs(log_dir)
+    hdlr = logging.FileHandler(filename=log_dir+'/calico.log')
```
Symmetric: Should probably be more specific than 'calico.log'. 'calico-rkt.log'?

Makefile
alexhersh: since the structure was copied from other repos, calico_rkt inherited two Dockerfile/req files. I deleted one of them and decided to keep them out of root. I suppose I got rid of the wrong one.

Symmetric: When I created the calico-rkt repo I fixed that; the build_calico_rkt directory has been added back in at some point since then. I think you should back it out (see Tom's libnetwork-plugin for supporting evidence that it's not necessary for plugins).

calico_rkt/logutils.py
```diff
@@ -0,0 +1,17 @@
+#!/bin/python
+import logging
+import os
+
+LOG_DIR = '/var/log/calico/rkt-plugin'
+
+def configure_logger(logger, log_dir=None):
```
Symmetric: Minor nit/code style point:

I'm not really convinced that this function justifies its own module; do we expect to call configure_logger from anywhere other than calico_rkt.py? And do we expect to add other logutils to this module in the future? If the answer to either is yes then this module is justified, but I think the answer to both questions is no.

It might be enough code to put in a function, but I'd just put that in calico_rkt.py. Or you could just put the 7 lines from this function that we actually use (11-17) at the top of calico_rkt.py. Again, minor nit; but worth trying to keep the project structure simple where possible (particularly in these greenfield situations where we're starting a new repo and have the chance to keep it minimally complex).

Symmetric: Alternatively since we have similar code elsewhere, there's an argument that it could go in libcalico and we import it from there.

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,150 @@
...
-# ETCD_PROFILE_PATH = '/calico/'
 
 ORCHESTRATOR_ID = "rkt"
 HOSTNAME = socket.gethostname()
-NETNS_ROOT= '/var/lib/rkt/pods/run'
+NETNS_ROOT = '/var/lib/rkt/pods/run'
 
-def main():
+def main(env, conf_in):
     mode = env['CNI_COMMAND']
```
djosborne: To simplify the naming convention, how about just naming this `cni_command = env['CNI_COMMAND']`

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,150 @@
...
-            container_id=env['CNI_CONTAINERID']
-            )
+    if mode == 'ADD':
+        create(env=env, conf_in=conf_in)
     elif mode == 'DEL':
-        print_stderr('Executing Calico pod-deletion plugin')
-        delete(
-            container_id=env['CNI_CONTAINERID']
-            )
+        delete(env=env, conf_in=conf_in)
```
djosborne: Might be a good idea to do:
```
else:
     log.error("Unrecognized command: %s" % cni_command)
     sys.exit(1)
```

calico_rkt/calico_rkt.py
```diff
@@ -229,21 +239,51 @@ def _apply_rules(profile_name, client):
...
+    candidate = SequentialAssignment().allocate(pool)
+    _log.info("Using IP %s" % candidate)
+    return IPAddress(candidate)
 
 if __name__ == '__main__':
-    main()
+    ENV = os.environ.copy()
+    ENV[ETCD_AUTHORITY_ENV] = 'localhost:2379' if ETCD_AUTHORITY_ENV not in ENV.keys() else ENV[ETCD_AUTHORITY_ENV]
+
+    input_ = ''.join(sys.stdin.readlines()).replace('\n', '')
```
djosborne: What's the trailing underscore indicate?

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,150 @@
...
 
-        _create_profile(endpoint=endpoint, 
-                        profile_name=input_json['name'], 
+        _set_profile_on_endpoint(endpoint=endpoint,
+                        profile_name=conf_in['name'],
                         ip=ip,
                         client=_datastore_client)
     except CalledProcessError as e:
-        print_stderr('Error code %d creating pod networking: %s\n%s' % (
+        _log.error('ERROR: error code %d creating pod networking: %s\n%s' % (
```
djosborne: Your logger format already has ` %(levelname)s` in the header, so it will automatically add the "ERROR" prefix to this message.

calico_rkt/calico_rkt.py
```diff
@@ -1,138 +1,150 @@
...
+                                               conf_in=conf_in,
+                                               interface = env['CNI_IFNAME'])
 
-        _create_profile(endpoint=endpoint, 
-                        profile_name=input_json['name'], 
+        _set_profile_on_endpoint(endpoint=endpoint,
+                        profile_name=conf_in['name'],
                         ip=ip,
                         client=_datastore_client)
     except CalledProcessError as e:
```
djosborne: You should split these two function calls each into their own try/except blocks. It ensures that your only catching the errors you expect each to call. 

Even if in this situation its possible that both functions throw the error, we'll still want them split so later on we can add teardown if the second fails but not the first. e.g. the except block for _set_profile_on_endpoint will eventually need to delete the endpoint (but thats something we'll implement later)

calico_rkt/calico_rkt.py
djosborne: I was strongly suggesting Evan weed out some of the global variables since his input sanitation is a bit scattered, and it makes testing the code more difficult since you have to override the global variables.

calico_rkt/calico_rkt.py
```diff
@@ -229,21 +239,51 @@ def _apply_rules(profile_name, client):
...
-    main()
+    ENV = os.environ.copy()
+    ENV[ETCD_AUTHORITY_ENV] = 'localhost:2379' if ETCD_AUTHORITY_ENV not in ENV.keys() else ENV[ETCD_AUTHORITY_ENV]
+
+    input_ = ''.join(sys.stdin.readlines()).replace('\n', '')
+    INPUT_JSON = json.loads(input_).copy()
+
+    configure_logger(_log)
+        
+    main(ENV, INPUT_JSON)
```
Symmetric: Missing \n

Symmetric:
If you're not going to use an IDE which has a built-in pep-8 linter, you should run pyflakes or similar before pushing code for review; pep-8 compliance is strongly preferred (though I don't think we've specified it as a hard requirement (yet)).

calico_rkt/calico_rkt.py
Symmetric: Fair enough, I think it's a minor point either way.

If you're not making env a global, I think it's clearer to go all the way
and explicitly pull the variables out of ENV before passing them to the
create() function; that way create() is more clearly spec'd, and testing
becomes even more neat (no need to construct an 'env' with who-knows-what
in it, just look at the function args).

alexhersh:
Now using Sublime's AutoPEP8. :+1:

calico_rkt/calico_rkt.py
```diff
@@ -1,198 +1,212 @@
...
-def create(container_id):
+
+def create(args):
     """"Handle rkt pod-create event."""
-    print_stderr('Configuring pod %s' % container_id, file=sys.stderr)
-    netns_path='%s/%s/%s' % (NETNS_ROOT, container_id, env['CNI_NETNS'])
-    _datastore_client = IPAMClient()
+    container_id = args['container_id']
+    netns = args['netns']
+    interface = args['interface']
```
djosborne: awww yussss

calico_rkt/calico_rkt.py
```diff
@@ -1,198 +1,212 @@
...
+LOG_DIR = '/var/log/calico/calico-rkt'
 
+ORCHESTRATOR_ID = "rkt"
+HOSTNAME = socket.gethostname()
+NETNS_ROOT = '/var/lib/rkt/pods/run'
 
-input_ = ''.join(sys.stdin.readlines()).replace('\n', '')
-input_json = json.loads(input_)
+_log = logging.getLogger(__name__)
+_datastore_client = IPAMClient()
```
djosborne: Can you replace this to just be `datastore_client`? Its not really a private value at this point...

calico_rkt/calico_rkt.py
```diff
@@ -1,198 +1,212 @@
...
+    # Delete profile if only member
+    if _datastore_client.profile_exists(net_name) and \
+       len(_datastore_client.get_profile_members(net_name)) < 1:
+        try:
+            _log.info(
+                "Profile %s has no members, removing from datastore" % net_name)
+            _datastore_client.remove_profile(net_name)
+        except:
+            _log.error(
+                "Cannot remove profile %s: Profile cannot be found." % container_id)
```
djosborne: Can you make these two log messages one line? I'd say this constitutes a time when its ok to bump over 80 characters

Symmetric: I think we have previously agreed that the pep8 limit should be 99 chars
for Calico projects. But if you hit that limit in an error message, just do
a line break somewhere in the string.

There are a couple ways to do that;

```
_log.error("abc"
              "def")
```
works, because of Python's string concatenation rules.

Or,

```
_log.error("abc %s" \
               % var)
```
Should also work.

calico_rkt/calico_rkt.py
```diff
@@ -219,31 +237,154 @@ def _create_rules(id_):
...
-    profile.rules = _create_rules(profile_name)
-    client.profile_update_rules(profile)
-    print_stderr("Finished applying rules.")
+    profile.rules = _create_default_rules(profile_name)
+    _datastore_client.profile_update_rules(profile)
+    _log.info("Finished applying rules.")
+
 
-def _allocate_IP():
+def _generate_pool(subnet):
```
djosborne: I took a closer look and I think this function should be removed. All it really does is call a datastore function after a very small type conversion. Can you move the functionality up into container_add? Same with  _allocate_ip_pool (But definitely argue for keeping it if you think It'll be useful later on!)

calico_rkt/calico_rkt.py
```diff
@@ -1,198 +1,212 @@
...
         sys.exit(1)
 
     # Create the veth, move into the container namespace, add the IP and
     # set up the default routes.
     ep.mac = ep.provision_veth(Namespace(netns_path), interface)
-    client.set_endpoint(ep)
+    _datastore_client.set_endpoint(ep)
 
-    # Let the caller know what endpoint was created.
     return ep, ip
```
djosborne: I'm wondering if this is necessary, since IP is stored in ep.ip_nets. But this is just a note-to-self, not sure if action is required

Symmetric: Think you're right, at the time I wrote this I didn't grok the ep.nets vs ep ip.

calico_rkt/calico_rkt.py
```diff
@@ -1,198 +1,212 @@
...
+    pool = _generate_pool(subnet)
+    ip = _allocate_ip(pool)
 
     # Create Endpoint object
     try:
-        ep = client.create_endpoint(HOSTNAME, ORCHESTRATOR_ID,
-                                      container_id, [IPAddress(ip)])
+        ep = _datastore_client.create_endpoint(HOSTNAME, ORCHESTRATOR_ID,
+                                               container_id, [ip])
     except AddrFormatError:
```
djosborne: Seems odd to me that we'd get an AddrFormatError here. Any clue why this is here?

calico_rkt/calico_rkt.py
```diff
@@ -1,198 +1,212 @@
...
+                                                  workload_id=container_id)
     except KeyError:
-        print_stderr("Container %s doesn't contain any endpoints" % container_id)
+        _log.error(
+            "Container %s doesn't contain any endpoints" % container_id)
         sys.exit(1)
 
     # Remove any IP address assignments that this endpoint has
     for net in endpoint.ipv4_nets | endpoint.ipv6_nets:
         assert(net.size == 1)
```
djosborne: Add a message here, otherwise if this assertion is ever hit the logs will just read `Assertion Error` and with no stack trace or message its pretty frustrating to figure out which one was hit: `assert net.size == 1, "assertion message yadda"`

calico_rkt/calico_rkt.py
```diff
@@ -1,198 +1,212 @@
...
-    print_stderr('Configuring Pod Profile: %s' % profile_name)
+    Configure the calico profile to the endpoint
+    """
+    _log.info('Configuring Pod Profile: %s' % profile_name)
 
-    if client.profile_exists(profile_name):
-        print_stderr("Profile with name %s already exists, applying to endpoint." % (profile_name))
+    if _datastore_client.profile_exists(profile_name):
+        _log.info(
+            "Profile with name %s already exists, applying to endpoint." % (profile_name))
```
djosborne: This branch of the if statement doesn't seem extremely important, so lets invert the statement and just leave the True branch out
```
if not _datastore_client.profile_exists(profile_name):
    _log.info("Creating profile %s." % (profile_name))
    _datastore_client.create_profile(profile_name)
```

calico_rkt/calico_rkt.py
```diff
@@ -1,198 +1,212 @@
...
-    print_stderr('Setting profile %s on endpoint %s' %
-                 (profile_name, endpoint.endpoint_id))
-    client.set_profiles_on_endpoint(
-        [profile_name], endpoint_id=endpoint.endpoint_id
-    )
-    print_stderr("Finished configuring profile.")
+    _datastore_client.set_profiles_on_endpoint(profile_names=[profile_name],
+                                               endpoint_id=endpoint.endpoint_id)
+
     dump = json.dumps(
```
djosborne: Passing in the IP just to print it out doesn't seem sensible since set_profile_on_endpoint doesn't use the IP for anything else - try moving this up into the caller (create)

calico_rkt/calico_rkt.py
```diff
@@ -219,31 +237,154 @@ def _create_rules(id_):
     rules = Rules.from_json(rules_json)
     return rules
 
-def _apply_rules(profile_name, client):
+
+def _apply_default_rules(profile_name):
```
djosborne: I think apply_default_rules can be merged with create_default_rules, they're both pretty small and will probably always be called one after another. Rename the new one something like `assign_default_rules_to_profile`

djosborne:
Very nice work, pulling out those global arguments makes this look a thousand times better. 

One small change - it looks like your sublime linter went off the chain a bit - a lot of those log messages that are just over 80 characters have been split up to 2-3 lines instead of just 1. That's one of the typical scenarios that we ignore pep8, as 1 line is much more readable. squish those guys back down to 1 line and we're almost ready to merge

calico_rkt/calico_rkt.py
Symmetric: I think we have previously agreed that the pep8 limit should be 99 chars
for Calico projects. But if you hit that limit in an error message, just do
a line break somewhere in the string.

There are a couple ways to do that;
```
_log.error("abc"
              "def")
```
works, because of Python's string concatenation rules.

Or,

```
_log.error("abc %s" \
               % var)
```
Should also work. It's OK to break pep8 in test code where it aids clarity, but we're generally not OK with breaking it in normal code.

calico_rkt/calico_rkt.py
alexhersh: create_endpoint can raise it on a IP version mismatch. if that happens, we need to roll back our IP allocation

calico_rkt/calico_rkt.py
```diff
@@ -1,209 +1,227 @@
...
+    return endpoint
+
+
+def _container_add(hostname, orchestrator_id, container_id, netns_path, interface, subnet):
+    """
+    Add a container to Calico networking
+    Return Endpoint object and newly allocated IP
+    """
+    # Allocate and Assign ip address through datastore_client
+    pool, ip = _assign_to_pool(subnet)
```
djosborne: I'm still not convinced this needs its own function, I think you can perform it inline here. but I'm 50/50

Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM debian:wheezy
+FROM ubuntu:14.04
```
caseydavenport: Why have you reverted this to ubuntu:14.04?

We build on debian for various dependency reasons (libc version, etc).

calico_rkt/calico_rkt.py
```diff
@@ -1,209 +1,256 @@
...
     except KeyError:
-        print_stderr("Container %s doesn't contain any endpoints" % container_id)
+        _log.error("Container %s doesn't contain any endpoints" % container_id)
         sys.exit(1)
 
     # Remove any IP address assignments that this endpoint has
     for net in endpoint.ipv4_nets | endpoint.ipv6_nets:
-        assert(net.size == 1)
-        client.unassign_address(None, net.ip)
+        assert net.size == 1, "Net Size of endpoint's ip' is not 1. %s" % net
```
caseydavenport: This is a pretty awkward error message.

net.size returns the number of IP addresses in the IPNetwork, so this should be something like "Only 1 address is supported per endpoint, found %s"

caseydavenport:
LGTM.

PR #4, opened 2015-09-10
Fix bugs in README.

fasaxc:
* rkt requires plugins to be executable.
* Trailing comma in network config file (rejected by rkt's parser).


alexhersh:
LGTM

PR #7, opened 2015-12-09
Modernized CNI plugin with basic functionality

L-MA:


calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+        Uses the given CNI_COMMAND to determine which action to take.
+
+        :return: None.
+        """
+        if self.command == CNI_CMD_ADD:
+            # If an add fails, we need to clean up any changes we may
+            # have made.
+            try:
+                self.add()
+            except BaseException:
```
caseydavenport: This was just a carry over from the previous plugin layout.  We need to do some more thinking on the correct approach to cleanup, so I say we remove this whole try/catch thing for now.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+        """
+        try:
+            _log.info("Retrieving endpoint that matches container ID %s",
+                      self.container_id)
+            ep = self._client.get_endpoint(hostname=HOSTNAME,
+                                           orchestrator_id=ORCHESTRATOR_ID,
+                                           workload_id=self.container_id)
+        except KeyError:
+            _log.warning("No endpoint found matching ID %s", self.container_id)
+            sys.exit(0)
```
caseydavenport: I prefer we follow the format of the exec plugin - if the endpoint isn't found, log a debug log and return None.  The calling code might not want to exit if the endpoint doesn't exist, so it can make that decision on its own.

e.g
```
# Check if already exists.
endpoint = self._get_endpoint()
if endpoint:
  _log.error("Endpoint already exists!")
  return

_log.info("Endpoint does not exist, creating")
```

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+        plugin_path = ""
+        for path in self.cni_path.split(":"):
+            if os.path.isfile(os.path.abspath(os.path.join(path, plugin_type))):
+                plugin_path = path
+                break
 
-def _set_profile_on_endpoint(endpoint, profile_name):
-    """
-    Configure the calico profile to the endpoint
+        if plugin_path == "":
```
caseydavenport: This can just be `if not plugin_path`

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+        :return: None.
+        """
+        _log.info('Deleting pod %s' % self.container_id)
+
+        # Step 1: Get the Calico endpoint for this workload. If it does not
+        # exist, log a warning and exit successfully.
+        endpoint = self._get_endpoint()
+
+        # Step 2: The endpoint exists - remove any IP assignments.
+        self._release_ips()
```
caseydavenport: We actually don't need the endpoint when we're using CNI endpoint, so this should get executed even if the endpoint doesn't exist.

IPAM uses the `container_id` as a handle for the IP allocation, and can remove it even if the endpoint is missing.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+
+        # Step 3: Delete the veth interface for this endpoint.
+        _log.info("Removing veth for endpoint %s", endpoint.name)
+        netns.remove_veth(endpoint.name)
```
caseydavenport: We'll want error handling for this (as well as a bunch of other stuff) - happy to merge this to master with the promise of a dedicated error handling PR later (as discussed in person).

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
 
-    # Remove the profile for the workload.
-    _container_remove(hostname=HOSTNAME,
-                      orchestrator_id=ORCHESTRATOR_ID,
-                      container_id=container_id)
+        # The request was successful.  Get the IP.
+        _log.info("IPAM result: %s", self.ipam_result)
+        return IPAddress(self.ipam_result["ip4"]["ip"])
+
+    def _release_ips(self, ip_list=[]):
```
caseydavenport: You shouldn't ever use a list object as the default value for a parameter - it will get re-used across multiple calls to this method.

Instead, you should do this:
```
def _release_ips(self, ip_list=None):
  if not ip_list:
    ip_list = []
```

As it stands, I don't think the `ip_list` argument is ever used, so it should probably just be removed.

Also, this should probably just be called `_release_ip` for symmetry with `_assign_ip`.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+        """
+        # TODO - add ips in ip_list to network_config attribute
+        # (network_config['ipam']['ips']
+        # TODO - Handle errors thrown by IPAM plugin
+        _log.debug("Releasing IP address with IPAM plugin")
+        try:
+            result = self._call_ipam_plugin()
+            _log.debug("IPAM plugin result: %s", result)
+        except CalledProcessError:
+            _log.exception("IPAM plugin failed to un-assign IP address.")
```
caseydavenport: s/un-assign/release.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+        return IPAddress(self.ipam_result["ip4"]["ip"])
+
+    def _release_ips(self, ip_list=[]):
+        """Releases the IP address(es) for this container using the IPAM plugin
+        specified in the network config file.
+
+        :param ip_list (optional) - List of IPs to release
+        :return: None.
+        """
+        # TODO - add ips in ip_list to network_config attribute
```
caseydavenport: Don't think we need to do this top TODO.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+            _log.warning("No endpoint found matching ID %s", self.container_id)
+            sys.exit(0)
+        except MultipleEndpointsMatch:
+            _log.error("Multiple endpoints found matching ID %s", self.container_id)
+            sys.exit(1)
 
-    _log.info("Removed Calico interface from %s" % container_id)
+        return ep
 
+    def _find_ipam_plugin(self):
```
caseydavenport: This could be more generically useful if it was changed to `_find_plugin(self, plugin_name)`.

 It could then search the CNI_PATH for a plugin with the given name, and could be used for more than just IPAM plugins sometime in the future.  For now, the only type of plugin that exists is IPAM, so this is fine, but thought I'd mention it anyway.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+        """
+        try:
+            _log.info("Removing endpoint from the Calico datastore.")
+            self._client.remove_workload(hostname=HOSTNAME,
+                                         orchestrator_id=ORCHESTRATOR_ID,
+                                         workload_id=self.container_id)
+        except KeyError:
+            _log.error("Unable to remove workload with ID %s from datastore.",
+                       self.container_id)
+            sys.exit(1)
```
caseydavenport: We shouldn't sys.exit() in any of the cleanup methods.  We want to continue trying to clean up state even if we fail to clean up one piece. 

e.g We fail to delete the endpoint, we should still release IPs, delete profiles, if we can.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
+    def _provision_veth(self, endpoint):
+        """Provisions veth for given endpoint.
+
+        Uses the netns relative path passed in through CNI_NETNS_ENV and
+        interface passed in through CNI_IFNAME_ENV.
+
+        :param endpoint
+        :return Calico endpoint object
+        """
+        netns_path = os.path.abspath(os.path.join(os.getcwd(), self.netns))
```
caseydavenport: Let's rename `self.netns` to `self.cni_netns` to better match the CNI_NETNS environment variable naming.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
...
-                                     orchestrator_id=orchestrator_id,
-                                     workload_id=container_id)
+    def _remove_profile(self, profile_name):
+        """Remove the profile from datastore if it has no endpoints attached.
+
+        :param profile_name: Name of profile to remove.
+        :return: None.
+        """
+        profile_members = self._client.get_profile_members(profile_name)
+        if self._client.profile_exists(profile_name) and len(profile_members) < 1:
```
caseydavenport: As discussed, this won't work because of the race condition when creating / deleting a pod in the same profile at the same time.

We need to solve this, but not in this PR - for now, we just won't delete profiles (there should be very few of them).

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,474 @@
 import os
 import sys
 
-from subprocess import check_output, CalledProcessError, check_call
+from subprocess import check_output, CalledProcessError, check_call, Popen, PIPE
```
caseydavenport: This should be subprocess32 in order to match libcalico.

caseydavenport:
Good start.  I've made some comments.  

Once you address them, could you re-open this PR against the branch `k8s-cni-integration` and let me know it is ready for another look?

PR #8, opened 2015-12-09
Create policy plugin for CNI

L-MA:


calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,199 @@
...
+            _log.info("Creating new profile '%s'", profile_name)
+            self._client.create_profile(profile_name)
+
+        # Set the default profile on this pod's Calico endpoint.
+        _log.info("Setting profile %s on endpoint.", profile_name)
+        self._client.set_profiles_on_endpoint(profile_names=[profile_name],
+                                              endpoint_id=endpoint.endpoint_id)
+
+    def remove_profile(self):
+        """Remove the profile from datastore.
```
caseydavenport: Docstring on this guy is wrong - it should explain why we aren't doing anything here.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,199 @@
...
+    def remove_profile(self):
+        """
+        Removes a profile for the networked container.
+
+        :return:
+        """
+        raise NotImplementedError("Must implement method in subclass")
+
+
+class DefaultPolicyPlugin(BasePolicyPlugin):
```
caseydavenport: Class needs a docstring describing its behavior and when it is used.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,199 @@
...
+from pycalico.datastore import DatastoreClient
+from pycalico.datastore_datatypes import Rule, Rules
+
+KUBERNETES_DEFAULT_PROFILE = "default_profile"
+
+_log = logging.getLogger(__name__)
+configure_logging(_log)
+
+class BasePolicyPlugin(object):
+    def __init__(self, network_config, env):
```
caseydavenport: There is a whole bunch of stuff here that isn't used, and so doesn't need to be stored.

I don't think we should be copying the __init__ procedure from the CniPlugin class, since they're pretty different.  It is OK for the subclasses of BasePolicyPlugin to require different arguments in __init__, and it is up to them to decide what the required arguments are.

For example:
```
# DefaultPolicyDriver only needs the network name.
driver = DefaultPolicyDriver(network_name)
driver.apply_policy(endpoint)

# KubernetesDefaultDriver doesn't need any extra arguments
driver = KubernetesDefaultDriver()
driver.apply_policy()

# KubernetesAdvPolicyDriver needs more information!
driver = KubernetesAvcPolicyDriver(env, calico_config)
driver.apply_policy()
```

caseydavenport: In other words, these aren't CNI plugins (which is why I'd call them driver's instead of plugins), and so they shouldn't need to be aware of CNI specific concepts, like CNI_ARGS, etc.

calico_rkt/calico_rkt.py
```diff
@@ -20,381 +20,448 @@
...
+                self.env
+            )
+        else:
+            policy_plugin = calico_policy.KubernetesDefaultPolicyPlugin(
+                self.network_config,
+                self.env
+            )
+
+        policy_plugin.execute(endpoint)
+
```
caseydavenport: Extra newline here.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        Path in which to search for CNI plugins.
+        """
+
+        self.network_name = network_config["name"]
+        """
+        Name of the network from the provided network config file.
+        """
+
+    def parse_cni_args(self, cni_args):
```
caseydavenport: Don't need this guy.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,199 @@
...
+
+        # For each arg, split at the equal sign and add to the dictionary.
+        _log.debug("Parsing CNI_ARGS: %s", cni_args)
+        for arg in cni_args.split(";"):
+            _log.debug("\tParsing CNI_ARG: %s", arg)
+            k, v = arg.split("=")
+            args_to_return[k] = v
+        return args_to_return
+
+    def execute(self, endpoint):
```
caseydavenport: Don't need this guy.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,126 @@
...
+    def set_profile(self, endpoint):
+        """Sets a profile for the networked container on the given endpoint.
+
+        Create a profile if it is not yet created.
+
+        :param endpoint:
+        :return:
+        """
+        if not self.profile_name:
+            raise NotImplementedError("A profile name has not been set.")
```
caseydavenport: I don't think NotImplementedError is the right exception here.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,126 @@
...
+        """Right now this function is no-op.
+
+        Need to think more about how to handle the race condition that exists
+        between removing a profile and creating a new one of the same name.
+
+        For now we'll leak profiles in Calico datastore.
+
+        :return:
+        """
+        _log.info("Called remove profile. Doing nothing for now.")
```
caseydavenport: This should still raise a NotImplementedError - sub-classes need to explicitly override this.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def generate_rules(self):
```
caseydavenport: You should populate the return: value here - indicate that returning None uses default rules.

calico_rkt/calico_rkt.py
```diff
@@ -13,388 +13,445 @@
...
+        assigned_ip = self._assign_ip()
+
+        # Step 2: Create the Calico endpoint object.
+        endpoint = self._create_endpoint(assigned_ip)
+
+        # Step 3: Provision the veth for this endpoint.
+        endpoint = self._provision_veth(endpoint)
+        
+        # Step 4: Provision / set profile on the created endpoint.
+        self.policy_driver.set_profile(endpoint)
```
caseydavenport: What error cases do we need to handle for this?

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,126 @@
...
+            self._client.create_profile(self.profile_name, rules)
+
+        # Set the default profile on this pod's Calico endpoint.
+        _log.info("Setting profile '%s' on endpoint %s",
+                  self.profile_name, endpoint.endpoint_id)
+        self._client.set_profiles_on_endpoint(profile_names=[self.profile_name],
+                                              endpoint_id=endpoint.endpoint_id)
+
+    def remove_profile(self):
+        """Right now this function is no-op.
```
caseydavenport: typo - "function is a no-op"

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,126 @@
...
+
+        Need to think more about how to handle the race condition that exists
```
caseydavenport: This comment belongs in the sub-classes - this isn't a general case problem (for example, this doesn't apply when using the (not implemented yet) annotation-based policy driver.

calico_rkt/calico_rkt.py
```diff
@@ -368,34 +362,18 @@ def _provision_veth(self, endpoint):
...
-        :return: None.
+        :return: a policy driver of type BasePolicyDriver
         """
-        pass
+        try:
+            self.cni_args["K8S_POD_NAME"]
+        except KeyError:
+            _log.debug("Using Default Policy Driver")
+            return calico_policy.DefaultPolicyDriver(self.network_name)
+        else:
```
caseydavenport: Need a log here to indicate which policy driver is being used.

calico_rkt/calico_rkt.py
```diff
@@ -121,6 +110,12 @@ def __init__(self, network_config, env):
         to stdout at the end of execution.
         """
 
+        self.policy_driver = self._get_policy_driver()
+        """
+        Stores the policy driver based on the container runtime manager.
```
caseydavenport: I'd reword this, since eventually the policy driver might be chosen based on a number of things.  I'd say "Chooses the correct policy driver based on the given configuration"

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,149 @@
...
+        raise NotImplementedError("Must implement method in subclass")
+
+
+class DefaultPolicyDriver(BasePolicyDriver):
+    """
+    Implements default network policy for a generic CNI plugin.
+    """
+    def __init__(self, network_name):
+        BasePolicyDriver.__init__(self)
+        self.profile_name = network_name
```
caseydavenport: I think we're going to need to validate and sanitize this network name.  Calico profiles have a limited character set, so we can't just allow any old string here.

Might need to check for illegal characters and raise an Exception if we find them.

caseydavenport:
I don't want to forget this, so I'm commenting here.

I don't think the json we print to stdout contains the CNI version - `"cniVersion": "0.1.0"`

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,149 @@
...
+    endpoints.
+    """
+    def __init__(self):
+
+        self._client = DatastoreClient()
+        """
+        DatastoreClient for access to the Calico datastore.
+        """
+
+        self.profile_name = None
```
caseydavenport: I think this would be better as a @property.

e.g
```
@property
def profile_name(self):
  raise NotImplementedError("Must specify a profile name")
```

Then we don't need to override __init__ in the subclasses, and allows the subclasses to be a bit more creative with how they generate a profile name.

caseydavenport: We could also get rid of that assert then.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,149 @@
...
+        """Right now this function is a no-op.
+
+        Need to think more about how to handle the race condition that exists
+        between removing a profile and creating a new one of the same name.
+
+        For now we'll leak profiles in Calico datastore.
+
+        :return: None
+        """
+        _log.info("Called remove profile. Doing nothing for now.")
```
caseydavenport: I'd reword this to be more clear.

"Not removing profile %s, clean up manually if desired"

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,156 @@
...
+                                              endpoint_id=endpoint.endpoint_id)
+
+    def remove_profile(self):
+        """Remove the profile if there are no endpoints attached.
+
+        :return: None
+        """
+        raise NotImplementedError("Must implement method in subclass")
+
+    def generate_rules(self):
```
caseydavenport: We'll eventually want to add support for labels as well as rules, but not needed for this iteration.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class DefaultPolicyDriver(BasePolicyDriver):
+    """
+    Implements default network policy for a generic CNI plugin.
+    """
+    def __init__(self, network_name):
+        BasePolicyDriver.__init__(self)
+        if not validate_characters(network_name):
+            raise ValueError("Invalid characters detected in the given network "
```
caseydavenport: Can we print a list of valid characters so the user knows what is wrong? Is that defined in libcalico somewhere?

caseydavenport:
One log to consider, then LGTM.

calico_rkt/calico_policy.py
```diff
@@ -0,0 +1,156 @@
...
+from calico_rkt import configure_logging
+from pycalico.datastore import DatastoreClient
+from pycalico.datastore_datatypes import Rule, Rules
+from pycalico.utils import validate_characters
+
+KUBERNETES_DEFAULT_PROFILE = "default_profile"
+
+_log = logging.getLogger(__name__)
+configure_logging(_log)
+
```
caseydavenport: Two newlines before a class

caseydavenport:
Let's rename this to "policy_drivers.py" - it is more specific than "calico_policy.py" (we might have more than one policy related file)

PR #9, opened 2015-12-09
Modernized CNI plugin and provide basic functionality

L-MA:


calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
+
+        # Step 3: Provision the veth for this endpoint.
+        endpoint = self._provision_veth(endpoint)
+        
+        # Step 4: Provision / set profile on the created endpoint.
+        self._set_profile_on_endpoint(endpoint, self.network_name)
+
+        # Step 5: If all successful, print the IPAM plugin's output to stdout.
+        dump = json.dumps(self.ipam_result)
+        _log.info("Dumping info to container manager: %s", dump)
```
caseydavenport: "Printing CNI result to stdout" is probably a more accurate log.

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
-                                     orchestrator_id=orchestrator_id,
-                                     workload_id=container_id)
+    def _remove_profile(self, profile_name):
+        """Remove the profile from datastore if it has no endpoints attached.
+
+        :param profile_name: Name of profile to remove.
+        :return: None.
+        """
+        profile_members = self._client.get_profile_members(profile_name)
+        if self._client.profile_exists(profile_name) and len(profile_members) < 1:
```
caseydavenport: This is a race condition - we shouldn't be doing anything here.

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
+        """
+        try:
+            _log.info("Retrieving endpoint that matches container ID %s",
+                      self.container_id)
+            ep = self._client.get_endpoint(hostname=HOSTNAME,
+                                           orchestrator_id=ORCHESTRATOR_ID,
+                                           workload_id=self.container_id)
+        except KeyError:
+            _log.warning("No endpoint found matching ID %s", self.container_id)
+            return
```
caseydavenport: Be explicit about returning None here - it makes the code easier to read.

```
return None
```

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
+        Return None if no endpoint is found.
+        Exits with an error if multiple endpoints are found.
+
+        :param container_id:
+        :return: Calico endpoint object if found, None if not found
+        """
+        try:
+            _log.info("Retrieving endpoint that matches container ID %s",
+                      self.container_id)
+            ep = self._client.get_endpoint(hostname=HOSTNAME,
```
caseydavenport: Avoid abbreviating unless needed - you don't need to call this `ep` - it can be called `endpoint`

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
 
-    _log.info("Removed Calico interface from %s" % container_id)
+        return ep
 
+    def _find_ipam_plugin(self):
+        plugin_type = self.network_config['ipam']['type']
+        plugin_path = ""
+        for path in self.cni_path.split(":"):
+            if os.path.isfile(os.path.abspath(os.path.join(path, plugin_type))):
+                plugin_path = path
```
caseydavenport: Instead of breaking like this, you could just return plugin_path here.

Also, this method needs some debug logging.

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
+                break
 
-def _set_profile_on_endpoint(endpoint, profile_name):
-    """
-    Configure the calico profile to the endpoint
+        if not plugin_path:
```
caseydavenport: I'd prefer if this was done when we call `_find_ipam_plugin` rather than in the method itself.  Easier to follow what is going on.

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
+        _log.info('Configuring pod %s' % self.container_id)
+
+        # Step 1: Assign an IP address using the given IPAM plugin.
+        assigned_ip = self._assign_ip()
+
+        # Step 2: Create the Calico endpoint object.
+        endpoint = self._create_endpoint(assigned_ip)
+
+        # Step 3: Provision the veth for this endpoint.
+        endpoint = self._provision_veth(endpoint)
```
caseydavenport: I think we're missing a step here - see https://github.com/projectcalico/calico-kubernetes/blob/master/calico_kubernetes/calico_kubernetes.py#L342

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
+            args_to_return[k] = v
+        return args_to_return
+
+    def execute(self):
+        """Executes this plugin.
+        Handles unexpected Exceptions in plugin execution.
+
+        :return The plugin return code.
+        """
+        rc = 0
```
caseydavenport: Stick a debug log here indicating that we've started plugin execution.

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
-from pycalico.datastore_errors import PoolNotFound
+from pycalico.netns import Namespace, remove_veth
+from pycalico.datastore import DatastoreClient
+from pycalico.datastore_errors import MultipleEndpointsMatch
 
+# Calico Configuration Constants
 ETCD_AUTHORITY_ENV = 'ETCD_AUTHORITY'
-LOG_DIR = '/var/log/calico/calico-rkt'
 
 ORCHESTRATOR_ID = "rkt"
```
caseydavenport: I need to look into what libcalico uses this for, but I suspect we're going to want this to be dynamic (this plugin could be rkt, k8s, or something else).

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
+        :param assigned_ip - IPAddress that has been already allocated
+        :return Calico endpoint object
+        """
+        try:
+            endpoint = self._client.create_endpoint(HOSTNAME,
+                                                    ORCHESTRATOR_ID,
+                                                    self.container_id,
+                                                    [assigned_ip])
+        except AddrFormatError:
+            _log.error("Unable to create endpoint. An invalid IP address was "
```
caseydavenport: I think this log is a little bit wrong.  `create_endpoint` also throws an `AddrFormatError` when an IPv6 address is provided, but IPv6 is not configured on the host (and vice-versa) even if the IPv6 address is valid.

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
-                      orchestrator_id=ORCHESTRATOR_ID,
-                      container_id=container_id)
+    def _release_ip(self):
+        """Releases the IP address(es) for this container using the IPAM plugin
+        specified in the network config file.
 
-    # Delete profile if only member
-    if datastore_client.profile_exists(net_name) and \
-       len(datastore_client.get_profile_members(net_name)) < 1:
+        :param ip_list (optional) - List of IPs to release
```
caseydavenport: This parameter doesn't exit any more so can be removed from the docstring.

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,473 @@
...
-    :param args: dict of values to pass to other functions (see: validate_args)
-    """
-    container_id = args['container_id']
-    net_name = args['name']
+        try:
+            # Load the response and get the assigned IP address.
+            self.ipam_result = json.loads(self.ipam_result)
+        except ValueError:
+            _log.exception("Failed to parse IPAM response, exiting")
+            sys.exit(1)
```
caseydavenport: We should be checking to make sure the `self.ipam_result["ip4"]["ip"] exists before we try to access it.  If it doesn't, we need to log an appropriate error and exit.

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,485 @@
...
 
-def delete(args):
-    """
-    Cleanup after a pod.
+        try:
+            assigned_ip = IPAddress(self.ipam_result["ip4"]["ip"])
+        except KeyError:
+            _log.error("IPAM plugin did not return an IPv4 address.")
+            sys.exit(1)
+        except (AddrFormatError, ValueError):
```
caseydavenport: Will it throw both of these? In which cases does it throw an `AddrFormatError`, and which cases does it throw a `ValueError`?

calico_rkt/calico_rkt.py
```diff
@@ -14,387 +14,485 @@
...
-    # Remove the profile for the workload.
-    _container_remove(hostname=HOSTNAME,
-                      orchestrator_id=ORCHESTRATOR_ID,
-                      container_id=container_id)
+        :return: None.
+        """
+        # TODO - Handle errors thrown by IPAM plugin
+        _log.debug("Releasing IP address with IPAM plugin")
+        try:
+            result = self._call_ipam_plugin
```
caseydavenport: You're not actually calling the the ipam plugin here - need some `( )` up in here.

calico_rkt/calico_rkt.py
L-MA: Yes it throws both - throws a ValueError if there is a '/' in the ip addr string; throws AddrFormatError for all other format errors.

calico_rkt/calico_rkt.py
L-MA: nice catch

PR #11, opened 2015-12-11
Move IPAM plugin to CNI repo

L-MA:
Commonize the configure_logging function

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ BUILD_FILES=Dockerfile requirements.txt
 
 default: all
 all: test
-binary: dist/calico_rkt
+binary: dist/calico dist/calico-ipam
 test: ut
 
 # Build a new docker image to be used by binary or tests
@@ -12,7 +12,7 @@ rktbuild.created: $(BUILD_FILES)
 	docker build -t calico/rkt-build .
 	touch rktbuild.created
 
-dist/calico_rkt: rktbuild.created
+dist/calico: rktbuild.created
 	mkdir -p dist
 	chmod 777 `pwd`/dist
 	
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -4,7 +4,7 @@ BUILD_FILES=Dockerfile requirements.txt
 
 default: all
 all: test
-binary: dist/calico_rkt
+binary: dist/calico_rkt dist/ipam
```
caseydavenport: This should be `dist/calico` and `dist/calico-ipam`

calico_rkt/constants.py
```diff
@@ -33,4 +33,8 @@
 
 # CNI Constants
 CNI_CMD_ADD = "ADD"
-CNI_CMD_DELETE = "DEL"
\ No newline at end of file
+CNI_CMD_DELETE = "DEL"
+
+# Logging Configuration
+LOG_DIR = "/var/log/calico/cni"
+LOG_FILENAME = "cni.log"
```
caseydavenport: I don't think the filename should be common.

The CNI plugin should have a separate log from the IPAM plugin.  The IPAM plugin can emit logs to its own logfile, as well as to `stderr`, and then the CNI plugin can capture the `stderr` output from the IPAM plugin and log it out into its own log file.

calico_rkt/util.py
```diff
@@ -0,0 +1,39 @@
...
+    if not os.path.exists(log_dir):
+        os.makedirs(log_dir)
+
+    log_path = os.path.join(log_dir, log_filename)
+
+    # Create a log handler and formtter and apply to _log.
+    hdlr = logging.FileHandler(filename=log_path)
+    formatter = logging.Formatter('%(asctime)s %(levelname)s %(message)s')
+    hdlr.setFormatter(formatter)
+    logger.addHandler(hdlr)
```
caseydavenport: Add a `stderr` handler as well.

calico_rkt/util.py
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import logging
+from constants import LOG_DIR, LOG_FILENAME
+
+
+def configure_logging(logger, log_filename=LOG_FILENAME, log_dir=LOG_DIR,
```
caseydavenport: log_filename should be a required argument.

calico_rkt/ipam.py
```diff
@@ -0,0 +1,173 @@
...
+from pycalico.ipam import IPAMClient
+from util import configure_logging
+
+LOG_FILENAME = "ipam.log"
+
+_log = logging.getLogger(__name__)
+datastore_client = IPAMClient()
+
+# Error codes
+ERR_CODE_UNHANDLED = 100
```
caseydavenport: These should probably be common so we don't duplicate error codes (though I admit I don't think we need error codes just yet, can probably use the same code for all errors)

caseydavenport:
LGTM.

PR #15, opened 2015-12-14
Introduce error handling

L-MA:


calico_cni/calico_cni.py
```diff
@@ -152,12 +152,13 @@ def execute(self):
...
             # SystemExit indicates an error that was handled earlier
             # in the stack.  Just set the return code.
-            rc = e.code 
+            rc = e.code
         except BaseException:
             # An unexpected Exception has bubbled up - catch it and
             # log it out.
             _log.exception("Unhandled Exception killed plugin")
-            rc = 1
+            rc = ERR_CODE_GENERIC
```
caseydavenport: This guy can still be ERR_CODE_UNHANDLED.  This is the only place that should be used, because it is hit when an unhandled exception is thrown :)

calico_cni/calico_cni.py
```diff
@@ -245,16 +246,17 @@ def delete(self):
 
         _log.info("Finished removing container: %s", self.container_id)
 
-    def _assign_ip(self):
+    def _assign_ip(self, env):
```
caseydavenport: Do we ever pass anything besides `self.env` to this method?

I'm not sure that an `env` argument is really required for `_assign_ip`, in which case we should just remove it.

caseydavenport: Actually, I'm not too bothered by this.  Fine to leave it as is.

calico_cni/calico_cni.py
```diff
@@ -245,16 +246,17 @@ def delete(self):
...
         """Assigns and returns an IPv4 address using the IPAM plugin
         specified in the network config file.
 
         :return: IPAddress - The assigned IP address.
         """
+        assert env[CNI_COMMAND_ENV] == CNI_CMD_ADD
```
caseydavenport: Put this assert after the log statement - it will be more clear what is going on if we ever hit this.

calico_cni/calico_cni.py
```diff
@@ -265,39 +267,57 @@ def _assign_ip(self):
 
         try:
             # Load the response and get the assigned IP address.
+            _log.debug("IPAM plugin result: %s", self.ipam_result)
```
caseydavenport: _call_ipam_plugin() already logs this out, so I don't think we need to do this here.

Also, `self.ipam_result` will always be `None` at this point, since it hasn't been set yet.

calico_cni/calico_cni.py
```diff
@@ -245,16 +246,17 @@ def delete(self):
...
         """
+        assert env[CNI_COMMAND_ENV] == CNI_CMD_ADD
         # Call the IPAM plugin.  Returns the plugin returncode,
         # as well as the CNI result from stdout.
         _log.info("Assigning IP address")
-        rc, result = self._call_ipam_plugin()
+        rc, result = self._call_ipam_plugin(env)
 
         if rc:
             # The IPAM plugin failed to assign an IP address. At this point in
```
caseydavenport: I think you need to move this `if rc:` block down after the `json.loads(result)` section, and do the error handling there.  We should be using the return code to determine if an error has occured - something like this.

```
if rc:
  code = self.ipam_response.get("code", ERROR_CODE_GENERIC)
  msg = self.ipam_response.get("msg", "Unknown IPAM error")
  details = self.ipam_response.get("details")
  self._print_error_response(code, msg, details)
  sys.exit(int(code))
```

calico_cni/calico_cni.py
```diff
@@ -265,39 +267,57 @@ def _assign_ip(self):
...
 
         _log.info("IPAM plugin assigned IP address: %s", assigned_ip)
         return assigned_ip
 
-    def _release_ip(self):
+    def _release_ip(self, env):
         """Releases the IP address(es) for this container using the IPAM plugin
         specified in the network config file.
 
         :return: None.
```
caseydavenport: Ditto - need to describe the purpose of the `env` variable.

calico_cni/calico_cni.py
```diff
@@ -336,14 +357,18 @@ def _create_endpoint(self, assigned_ip):
                                                     self.container_id,
                                                     [assigned_ip])
         except AddrFormatError:
-            _log.error("This node is not configured for IPv%s", assigned_ip.version)
-            # TODO - call release_ip / cleanup
-            sys.exit(1)
+            self.env[CNI_COMMAND_ENV] = CNI_CMD_DELETE
```
caseydavenport: I don't think we should be modifying `self.env` here - we should make a copy first.

e.g
```
env = self.env.copy()
env[CNI_COMMAND_ENV = CNI_CMD_DELETE
self._release_ip(env)
...

File: calico_cni/tests/unit/test_cni_plugin.py [hunk 8]
```diff
@@ -174,11 +174,15 @@ def test_add_mainline(self, m_json):
         self.plugin._provision_veth = MagicMock(spec=self.plugin._provision_veth)
         self.plugin._provision_veth.return_value = endpoint
 
+        # Mock out environment
```
caseydavenport: I don't think you need to mock out the environment here - you can just use the actual `self.plugin.env` value.

calico_cni/tests/unit/test_cni_plugin.py
```diff
@@ -174,11 +174,15 @@ def test_add_mainline(self, m_json):
...
+        m_env = Mock(spec=dict)
+        self.plugin.env = m_env
+
         # Call method.
         self.plugin.add()
 
         # Assert.
-        self.plugin._assign_ip.assert_called_once_with()
+        self.plugin._assign_ip.assert_called_once_with(m_env)
```
caseydavenport: As per above..
```
assert_called_once_with(self.plugin.env)
```

caseydavenport:
Done with the first round.

calico_cni/calico_cni.py
```diff
@@ -223,7 +224,7 @@ def delete(self):
         _log.info("Remove networking from container: %s", self.container_id)
 
         # Step 1: Remove any IP assignments.
-        self._release_ip()
+        self._release_ip(self.env)
 
         # Step 2: Get the Calico endpoint for this workload. If it does not
         # exist, log a warning and exit successfully.
```
caseydavenport: We need to handle errors thrown by `netns.remove_veth` a few lines down from here.

caseydavenport: We also need to handle policy driver errors when creating / removing profiles.

calico_cni/calico_cni.py
```diff
@@ -336,14 +357,18 @@ def _create_endpoint(self, assigned_ip):
                                                     self.container_id,
                                                     [assigned_ip])
         except AddrFormatError:
```
caseydavenport: Could you stick a comment in each branch explaining each case?

caseydavenport: We can probably combine these two branches, since they are performing the same thing.

For example:
```
try:
  ...
except (AddrFormatError, KeyError), e:
  ... do cleanup ...
  self._print_error_response(ERR_CODE_GENERIC, e.message)
```

We can split again if we want to use different error codes, but I don't think that we really care about differentiating these cases on the CNI - they're both still "failed to create endpoint".

caseydavenport: We probably still want to log an exception here, since it will help debugging.
```
_log.exception("Failed to create Calico endpoint")
```

I'm not sure how I feel about logging an error in the _print_error_message(...) method, since it isn't a flexible as handling the logging in each individual failure branch.

calico_cni/calico_cni.py
```diff
@@ -400,14 +424,17 @@ def _get_policy_driver(self):
...
             try:
                 driver = policy_drivers.DefaultPolicyDriver(self.network_name)
             except ValueError:
-                _log.error("Invalid characters detected in the network name "
-                           "'%s'. Only letters a-z, numbers 0-9, and _.- "
-                           " are supported", self.network_name)
-                sys.exit(1)
+                self.env["command"] = CNI_CMD_DELETE
+                self._release_ip(self.env)
+                message = "Invalid characters detected in the network name, " \
```
caseydavenport: I think this message should live in the DefaultPolicyDriver, and this code should just access the `message` attribute on the ValueError

calico_cni/calico_cni.py
```diff
@@ -400,14 +424,17 @@ def _get_policy_driver(self):
...
+            _log.debug("Using default policy driver")
             try:
                 driver = policy_drivers.DefaultPolicyDriver(self.network_name)
             except ValueError:
-                _log.error("Invalid characters detected in the network name "
-                           "'%s'. Only letters a-z, numbers 0-9, and _.- "
-                           " are supported", self.network_name)
-                sys.exit(1)
+                self.env["command"] = CNI_CMD_DELETE
+                self._release_ip(self.env)
```
caseydavenport: We don't actually need to release anything or clean up here, since this gets called in the __init__ method before we've done anything.

calico_cni/calico_cni.py
```diff
@@ -435,8 +462,8 @@ def _get_endpoint(self):
             _log.warning("No endpoint found matching ID %s", self.container_id)
             endpoint = None
         except MultipleEndpointsMatch:
-            _log.error("Multiple endpoints found matching ID %s", self.container_id)
-            sys.exit(1)
+            _log.warning("Multiple endpoints found matching ID %s", self.container_id)
```
caseydavenport: I'm not sure if this is the right behavior here.  It is definitely an error if multiple endpoints are returned for a single container_id.

calico_cni/calico_cni.py
```diff
@@ -195,7 +196,7 @@ def add(self):
                   self.container_id)
 
         # Step 1: Assign an IP address using the given IPAM plugin.
```
caseydavenport: I think we need to do a little more re-structure for the ADD case.  Didn't spot this before.

For CNI, it is possible that the plugin will be called multiple times for multiple different CNI networks (all of which use Calico for networking).  In that case, we don't want to assign multiple IP addresses, we don't want to create another endpoint, and we don't want to set up another veth.  We just want to append a profile for the new network.

Let's talk about this before making the change.

calico_cni/tests/unit/test_cni_plugin.py
```diff
@@ -132,7 +132,7 @@ def test_execute_unhandled_exception(self):
         rc = self.plugin.execute()
 
         # Assert success.
```
caseydavenport: I think this comment is wrong.

calico_cni/calico_cni.py
```diff
@@ -202,7 +202,16 @@ def add(self):
         endpoint = self._provision_veth(endpoint)
         
         # Step 4: Provision / set profile on the created endpoint.
-        self.policy_driver.set_profile(endpoint)
+        try:
+            self.policy_driver.set_profile(endpoint)
+        except policy_drivers.ApplyProfileError, e:
```
caseydavenport: We need to do more than just release the IP here.  

We've done a few things at this point:
1) Assign IP address
2) Create Endpoint object
3) Provision veth interface

Each of these needs to be un-done (probably in reverse order) if we hit an error.

calico_cni/calico_cni.py
```diff
@@ -202,7 +202,16 @@ def add(self):
         endpoint = self._provision_veth(endpoint)
```
caseydavenport: What errors can this throw?

We need to handle anything that goes wrong here and clean up the IP assignment and Endpoint from above.

calico_cni/policy_drivers.py
```diff
@@ -155,3 +164,15 @@ def generate_rules(self):
...
                       outbound_rules=[allow])
         return rules
+
+class ApplyProfileError(Exception):
+    """
+    Attempting to apply a profile to an endpoint that does not exist.
+    """
+    pass
+
+class RemoveProfileError(Exception):
```
caseydavenport: This isn't used anywhere, so there is no need to include it.

calico_cni/tests/fv/test_calico_cni.py
```diff
@@ -323,13 +323,13 @@ def test_add_error_profile_create(self):
         p = self.create_plugin()
 
         # Configure EtcdException when setting profile.
-        p.policy_driver._client.set_profiles_on_endpoint.side_effect = MagicMock(side_effect=EtcdException)
```
caseydavenport: Remove the `EtcdException` import at the top if it isn't used anywhere else.

calico_cni/calico_cni.py
```diff
@@ -233,7 +242,11 @@ def delete(self):
 
         # Step 3: Delete the veth interface for this endpoint.
         _log.info("Removing veth for endpoint: %s", endpoint.name)
-        netns.remove_veth(endpoint.name)
+        try:
+            if not netns.remove_veth(endpoint.name):
+                _log.warning("Veth for host %s does not exist.", endpoint.name)
```
caseydavenport: Change the wording here - host implies the VM or metal that the container is running on, when the veth itself is actually a per-container / pod construct.

```
"Veth %s does not exist", endpoint.name
```
and 
```
"Unable to remove veth %s", endpoint.name
```

calico_cni/calico_cni.py
```diff
@@ -252,50 +265,59 @@ def _assign_ip(self):
...
         """Releases the IP address(es) for this container using the IPAM plugin
         specified in the network config file.
 
+        :param env  - A dictionary of environment variables to pass to the
+        IPAM plugin
         :return: None.
         """
         _log.info("Releasing IP address")
         rc, result = self._call_ipam_plugin()
+        assert env[CNI_COMMAND_ENV] == CNI_CMD_DELETE
```
caseydavenport: You should assert this before calling the IPAM plugin.

caseydavenport:
LGTM

PR #24, opened 2016-01-05
Make CNI_ARGS optional

tomdee:
The scripts in the appc/cni repo don't set it.


caseydavenport:
Maybe a quick UT for this case?

caseydavenport:
Up to you if you think a UT is worth doing, otherwise LGTM.

PR #30, opened 2016-01-07
Annotation based policy.

caseydavenport:
Port annotation-based policy to CNI plugin as a policy driver.

Includes an FV test for this case and 100% UT coverage of new code.


calico_cni/calico_cni.py
```diff
@@ -512,11 +512,21 @@ def _get_policy_driver(self):
...
         else:
-            _log.debug("Using Default Kubernetes Policy Driver")
-            driver = policy_drivers.KubernetesDefaultPolicyDriver(
-                    self.network_name
-            )
-
+            # Decide which Kubernetes driver to use.
+            policy_block = self.network_config.get(POLICY_KEY, {})
+            policy_type = policy_block.get("type")
+            if policy_type == POLICY_MODE_ANNOTATIONS:
```
tomdee: Will there be any log at info level to indicate which policy engine was used? Should there be?

tomdee:
LGTM - I'll take a look at the wider CNI driver code on Monday

PR #51, opened 2016-02-11
Use ipv6 address if IPAM provides it.

tomdee:


calico.py
```diff
@@ -311,30 +313,22 @@ def _assign_ips(self, env):
...
-            message = "IPAM plugin did not return an IPv4 address."
-            print_cni_error(ERR_CODE_GENERIC, message)
-            sys.exit(ERR_CODE_GENERIC)
-        except (AddrFormatError, ValueError):
-            message = "Invalid or Empty IPv4 address: %s" % \
-                      (ipam_result["ip4"]["ip"])
-            print_cni_error(ERR_CODE_GENERIC, message)
-            sys.exit(ERR_CODE_GENERIC)
+            _log.info("IPAM plugin assigned IPv4 address: %s", ipv4)
+        except (KeyError, AddrFormatError, ValueError):
```
caseydavenport: I think AddrFormatError should still have its own except branch here, at least from a diagnostics point of view.  An invalid IPv4 address is different from no IPv4 address at all.

caseydavenport:
Looks nice to me.

Have you run FVs as well as the UTs?

PR #53, opened 2016-02-12
Don't assign IPv6 address when using calico-ipam

tomdee:
IPv6 isn't enabled by default on some hosts.

We should make it a user configurable option to turn it on.

### DIFF
--- a/ipam.py
+++ b/ipam.py
@@ -62,11 +62,10 @@ class IpamPlugin(object):
         if self.command == "ADD":
             # Assign an IP address for this container.
             _log.info("Assigning address to container %s", self.container_id)
-            ipv4, ipv6 = self._assign_address(handle_id=self.container_id)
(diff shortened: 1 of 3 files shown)


caseydavenport:
> We should make it a user configurable option to turn it on.

Why don't we just do this rather than rip out all the ipv6?

tomdee:
I'm not ripping it all out. I'm just telling our IPAM not to allocate IPv6 since we are not in a position to use the result. I don't have time for a more complete solution so I'd like to get this in as is.

caseydavenport:
Sure, if you think a more complete solution would take too much time that's fine.

My first impression was that making this configurable would have been two, maybe three lines of code, but perhaps there is more to it.

LGTM.

tomdee:
Easy to do as another PR then :smile:

PR #54, opened 2016-02-12
Changes for v1alpha1 policy API

caseydavenport:
This is semi backwards compatible with existing deployments.  Upgrading to this version will still be able to handle pods / endpoints created with previous versions of the plugin.  However, downgrade from this commit to an earlier commit will leak endpoints. 

- Uses the pod namespace and name for the workload_id in Kubernetes.
- Use 'k8s' as the orchestrator when running under Kubernetes rather than 'cni'
- Also introduces a new policy driver which denies all incoming traffic.

TODO:
- [x] Try to make this backwards compatible
- [x] Use the CNI network name for the calico profile (rather than `deny-inbound`)?
- [x] Re-name policy type "none" to something that makes sense.

### DIFF
--- a/calico.py
+++ b/calico.py
@@ -125,6 +125,20 @@ class CniPlugin(object):
         Path in which to search for CNI plugins.
         """
 
+        self.running_under_k8s = self.k8s_namespace and self.k8s_pod_name
+        if self.running_under_k8s:
+            self.workload_id = "%s.%s" % (self.k8s_namespace, self.k8s_pod_name)
+            self.orchestrator_id = "k8s"
+        else:
+            self.workload_id = self.container_id
+            self.orchestrator_id = "cni"
+        """
+        Configure orchestrator specific settings.
+        workload_id: In Kubernetes, this is the pod's namespace and name.  
+                     Otherwise, this is the container ID.
+        orchestrator_id: Either "k8s" or "cni".
+        """
+
     def execute(self):
         """
         Execute the CNI plugin - uses the given CNI_COMMAND to determine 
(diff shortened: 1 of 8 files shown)


tomdee:
Unassigning myself as the title says WIP are there are TODOs in the comment

caseydavenport:
After discussions with @tomdee, decided that to make this backwards compatible we should attempt to get / remove endpoints using the container_id should a lookup with the workload_id fail.

Note that this means once new pods have been deployed using the workload_id, downgrade to a previous CNI plugin version will be incompatible.

calico.py
```diff
@@ -125,6 +125,19 @@ def __init__(self, network_config, env):
         Path in which to search for CNI plugins.
         """
 
+        if self.k8s_namespace and self.k8s_pod_name:
+            self.workload_id = "%s.%s" % (self.k8s_namespace, self.k8s_pod_name)
```
tomdee: What about adding a `self.running_under_k8s = True` line. Then elsewhere, you get more readable code by just checking that(e.g. line 178)

calico.py
```diff
@@ -162,13 +175,29 @@ def add(self):
 
         _log.debug("Checking for existing Calico endpoint")
         endpoint = self._get_endpoint()
-        if endpoint:
+        if endpoint and not self.k8s_namespace:
```
tomdee: Why is the behavior different if running under k8s? The comment should explain

calico.py
```diff
@@ -162,13 +175,29 @@ def add(self):
...
+            # Not running under Kubernetes.
             # This endpoint already exists, add it to another network.
             _log.info("Endpoint for container exists - add to new network")
             output = self._add_existing_endpoint(endpoint)
+        elif endpoint and self.k8s_namespace:
+            # Running under Kubernetes and we've received a create for
```
tomdee: Is this code part of the new policy support or a bug fix?

caseydavenport: It is sort of both.

With the switch to using pod name and namespace for the workload ID, we need to handle this case.  If a host gets restarted, for example, we'll try to set up networking for a workload that already exists.

This make sure we don't have duplicate endpoints for the same workload.

tomdee: OK

tomdee:
Couple of small comments - otherwise LGTM

PR #55, opened 2016-02-12
Configuration for IPv4 vs IPv6 address assignment

caseydavenport:
@tomdee Turned out to be a bit more than 3 lines of code..

Allows the following:
```
{
  "type": "calico-ipam",
  "assign_ipv4": "true",
  "assign_ipv6": "true"
}

```

### DIFF
--- a/ipam.py
+++ b/ipam.py
@@ -62,10 +62,11 @@ class IpamPlugin(object):
         if self.command == "ADD":
             # Assign an IP address for this container.
             _log.info("Assigning address to container %s", self.container_id)
-            ipv4, _ = self._assign_address(handle_id=self.container_id)
+            ipv4, ipv6 = self._assign_address(handle_id=self.container_id)
     
             # Output the response and exit successfully.
-            return json.dumps({"ip4": {"ip": str(ipv4.cidr)}})
+            return json.dumps({"ip4": {"ip": str(ipv4.cidr)},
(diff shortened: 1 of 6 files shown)


ipam.py
```diff
@@ -95,15 +119,25 @@ def _assign_address(self, handle_id):
...
         else:
-            try:
-                ipv4 = ipv4_addrs[0]
-            except IndexError:
-                _log.error("No IPv4 address returned, exiting")
-                raise CniError(ERR_CODE_GENERIC,
-                               msg="No IPv4 addresses available in pool")
+            if num_v4:
+                try:
+                    ipv4 = ipv4_addrs[0]
```
tomdee: What about `ipv4 = IPNetwork(ipv4_addrs[0])` then you can simplify your return statement below.

ipam.py
```diff
@@ -95,15 +119,25 @@ def _assign_address(self, handle_id):
...
+                except IndexError:
+                    _log.error("No IPv4 address returned, exiting")
+                    raise CniError(ERR_CODE_GENERIC,
+                                   msg="No IPv4 addresses available in pool")
+
+            if num_v6:
+                try:
+                    ipv6 = ipv6_addrs[0]
```
tomdee: As above

tomdee:
One comment - otherwise LGTM

PR #57, opened 2016-02-26
Include calico.py, ipam.py in coverage reports

caseydavenport:


TrimBiggs:
LGTM

tests/unit/test_cni_plugin.py
```diff
@@ -355,6 +355,34 @@ def test_assign_ip_no_ip(self):
         e = err.exception
         assert_equal(e.code, ERR_CODE_GENERIC)
 
+    def test_assign_ip_invalid_ip4(self):
```
TrimBiggs: Would be nice to have docstrings for these tests, but looks like most tests here don't have them either :\

TrimBiggs:
LGTM with one minor nitpicky comment

tests/unit/test_cni_plugin.py
caseydavenport: Yeah you're right.

PR #65, opened 2016-03-09
Allow user to choose IP

tomdee:
e.g. using sudo rkt run --net="net1:IP=192.168.2.5"
--insecure-options=image docker://busybox --exec=/bin/ifconfig

Tested manually and 100% code coverage in UTs


Makefile
```diff
@@ -35,6 +35,10 @@ update-version:
 	echo "__commit__ = '$(shell git rev-parse HEAD)'" >> calico_cni/__init__.py 
 	echo "__branch__ = '$(shell git rev-parse --abbrev-ref HEAD)'" >> calico_cni/__init__.py
 
+# Copy the plugin into place
+deploy:
```
caseydavenport: Should this be `deploy-rkt` so it isn't misleading?

Should this depend on `dist/calico`?

ipam.py
```diff
@@ -98,9 +104,18 @@ def execute(self):
...
+            if self.ip:
+                # The user has specifically requested an IP (v4) address.
+                _log.info("User assigned address: %s for workload: %s",
+                          self.ip,
+                          self.workload_id)
+
+                ipv4 = self._assign_existing_address()
+                ipv6 = None
+            else:
+                # Assign an IP address for this workload.
```
caseydavenport: "Auto-assign an IP address for this workload"

Should clean up the terminology in this module now to be clear when we're auto-assigning and when we're not.

caseydavenport:
Couple of minor comments, a rebase, and then LGTM.

PR #73, opened 2016-03-19
Add a way to easily created a k8s "cluster"

tomdee:


caseydavenport:
I _kind of_ would rather this be separate from this repo.  For example, we already have a `docker-compose-kubernetes` repo that brings up a local cluster and I'd rather we collectively use and support that repo since it is one of our recommended "try out Calico on k8s" options.  I'm not a huge fan of sticking a bunch of scripting inside of the makefile to spin up Kubernetes, especially when we have other code which already does it.

That said, it is a pretty isolated change so I can't really complain too much...

tomdee:
Discussed offline and agreed it's useful so merging.

PR #74, opened 2016-03-22
Build and store the binary on CireclCI

tomdee:


caseydavenport:
CircleCI will automatically store the binary produced this way?

caseydavenport:
LGTM

PR #75, opened 2016-03-22
Use latest build container

### LINKED ISSUE #72: build are slow because binaries are rebuilt even if no changes
Maybe related to the versioning stuff?

caseydavenport:
Fixed #72


tomdee:
LGTM

PR #77, opened 2016-03-23
Turn on logging for libcalico

tomdee:


caseydavenport:
Failing tests.

caseydavenport:
LGTM once tests pass.

caseydavenport:
We should do this in the IPAM module as well I think.

PR #78, opened 2016-03-23
Use yaml to read network config

tomdee:
Avoids converting strings to/from unicode


caseydavenport:
Test are failing.

caseydavenport:
Looks fine to me once tests pass.  Any reason we can't handle unicode values in libcalico?

PR #81, opened 2016-03-28
Fix typo in substitution variable

EmmEff:
Fix typo which would cause those using search-and-replace to have a malformed cloud-config template


caseydavenport:
@EmmEff Thanks for spotting and fixing this!

PR #82, opened 2016-03-28
Allow hostname to be configured

### LINKED ISSUE #79: CNI plugin doesn't support having the hostname configured
It will only use a hostname of `socket.gethostname()` but hosts might be created with different names.

tomdee:
Fixes #79

### DIFF
--- a/calico.py
+++ b/calico.py
@@ -92,6 +92,11 @@ class CniPlugin(object):
         Type of IPAM to use, e.g calico-ipam.
         """
 
+        self.hostname = network_config.get("hostname", socket.gethostname())
+        """
+        The hostname to register endpoints under.
+        """
+
         self.policy_driver = get_policy_driver(self.k8s_pod_name,
                                                self.k8s_namespace,
                                                self.network_config)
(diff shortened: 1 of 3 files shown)


calico.py
```diff
@@ -38,6 +38,9 @@
 from calico_cni.policy_drivers import ApplyProfileError, get_policy_driver
 from ipam import IpamPlugin
 
+# Hostname which can be overridden with config
+HOSTNAME = socket.gethostname()
```
caseydavenport: This value is only used within the `CniPlugin` class.  Why don't we define this within in the class (like the rest of the config options) so we don't need to do this global stuff.

This should reduce the number of lines, and make it easier to manipulate in UTs.

```
self.hostname = config.get("hostname", socket.gethostname())
```

Or something like that.

calico.py
```diff
@@ -686,6 +689,11 @@ def main():
         os.environ[ETCD_ENDPOINTS_ENV] = etcd_endpoints
         _log.debug("Using %s=%s", ETCD_ENDPOINTS_ENV, etcd_endpoints)
 
+    hostname_config = network_config.get("hostname")
```
caseydavenport: All the other config options use constants for this.

tomdee:
OK, markups applied.

caseydavenport:
LGTM.

PR #84, opened 2016-04-05
Force called IPAM binaries to support unknown args

tomdee:


caseydavenport:
LGTM - do we need a UT to keep 100% coverage?

tomdee:
Coverage is still 100%.

calico.py
```diff
@@ -153,6 +153,14 @@ def __init__(self, network_config, env):
         orchestrator_id: Either "k8s" or "cni".
         """
 
+        # Ensure that the ipam_env CNI_ARGS contains the IgnoreUnknown=1 option
+        # See https://github.com/appc/cni/pull/158
+        # And https://github.com/appc/cni/pull/127
+        self.ipam_env[CNI_ARGS_ENV] = 'IgnoreUnknown=1'
```
caseydavenport: Actually, is this right?

We're blatting over all of the CNI_ARGS passed to the with "IgnoreUnknown=1.

PR #86, opened 2016-04-07
Correctly set labels and profile when using DefaultDeny driver

tomdee:


calico_cni/policy_drivers.py
```diff
@@ -320,6 +307,21 @@ def _escape_chars(self, unescaped_string):
         return re.sub('[^a-zA-Z0-9\.\_\-]', swap_char, unescaped_string)
 
 
+class DefaultDenyInboundDriver(KubernetesAnnotationDriver):
```
caseydavenport: This shouldn't be called `DefaultDenyInboundDriver` anymore.

calico_cni/policy_drivers.py
```diff
@@ -320,6 +307,21 @@ def _escape_chars(self, unescaped_string):
...
+    """
+    This driver sets the labels and correct profile on the endpoint.
+    """
+    def apply_profile(self, endpoint):
+        profile_id = "k8s_ns.%s" % self.namespace
```
caseydavenport: Could you please organize this code a little more - perhaps split into steps? It's a bit ugly as a single blob of code with no comments.

calico_cni/policy_drivers.py
```diff
@@ -320,6 +307,21 @@ def _escape_chars(self, unescaped_string):
...
+        _log.debug("Constructed profile ID - %s" % profile_id)
+        pod = self._get_api_pod()
+        labels = pod["metadata"].get("labels", {})
```
caseydavenport: The labels that need to be applied are a combination of:
1) The labels from the Kubernetes pod (which you're doing here)
2) A special namespace label (which you're not).  See the policy agent code here:  https://github.com/projectcalico/k8s-policy/blob/master/handlers/pod.py#L26

calico_cni/policy_drivers.py
caseydavenport: You'll need to override remove_profile as well I think. 

The annotation based driver tries to delete the profile on pod deletion which:
1) we don't want to do for this driver 
2) will output a warning whenever a pod is deleted.

calico_cni/policy_drivers.py
```diff
@@ -320,6 +307,25 @@ def _escape_chars(self, unescaped_string):
...
+        labels["calico/k8s_ns"] = self.namespace
+        _log.debug("Got labels - %s" % labels)
+        endpoint.labels = labels
+        endpoint.profile_ids = [profile_id]
+        self._client.update_endpoint(endpoint)
+
+    def remove_profile(self):
+        _log.debug("Nothing to do")
```
caseydavenport: More friendly/useful log plz (I understand that this is a "get it to casey fast" iteration for now)

calico_cni/constants.py
```diff
@@ -60,8 +60,8 @@
 ERR_CODE_GENERIC = 100   # Use this for all errors.
 
 # Policy modes.
-POLICY_MODE_ANNOTATIONS = "k8s-annotations"
-POLICY_MODE_DENY_INBOUND = "default-deny-inbound"
+POLICY_MODE_KUBERNETES_ANNOTATIONS = "k8s-annotations"
+POLICY_MODE_KUBERNETES = "k8s"
```
caseydavenport: Just a thought... do we want this to be `kubernetes` spelled out? Is that nicer or less nice?

calico_cni/policy_drivers.py
```diff
@@ -338,28 +356,40 @@ def get_policy_driver(k8s_pod_name, k8s_namespace, net_config):
...
+            elif policy_type == POLICY_MODE_KUBERNETES_ANNOTATIONS:
+                _log.debug("Using Kubernetes Annotation Policy Driver")
+                driver_cls = KubernetesAnnotationDriver
+
+            driver_args = [k8s_pod_name,
+                           k8s_namespace,
                            k8s_auth_token,
                            k8s_api_root]
         else:
             _log.debug("Using Default Kubernetes Policy Driver")
```
caseydavenport: I might re-word this log now to say "Using Kubernetes Driver - no policy" or something like that to be more clear.

calico_cni/policy_drivers.py
```diff
@@ -338,28 +356,40 @@ def get_policy_driver(k8s_pod_name, k8s_namespace, net_config):
     policy_config = net_config.get(POLICY_KEY, {})
     network_name = net_config["name"]
     policy_type = policy_config.get("type")
+    supported_policy_types = [None,
+                              POLICY_MODE_KUBERNETES,
+                              POLICY_MODE_KUBERNETES_ANNOTATIONS]
+    if policy_type not in supported_policy_types:
+        print_cni_error(ERR_CODE_GENERIC,
```
caseydavenport: So, this is going to print out "supported values: [None, k8s-annotations, ...]" and the value `None` is not actually supported.  I think it shouldn't include `None` when printing out the list.

caseydavenport:
A couple of comments to address, but otherwise looking good I think.

PR #96, opened 2016-05-04
"no handlers found for calico_cni", when conf file contains tab characters 

euprogramador:
correction of an error happenig when config file /etc/cna/net.d/10-calico.conf contains tab characters, and printing full stacktrace in log file in order to make easier to figure out where errors happen


caseydavenport:
LGTM - Thanks @euprogramador for digging into this and providing a fix!

PR #99, opened 2016-05-26
Added labels

djosborne:


caseydavenport:
A few concerns with this:
- The other side of this interface isn't written yet in Mesos (or any other orchestrator).
- how do we keep this consistent with the way Kubernetes is doing labels - see https://github.com/projectcalico/calico-cni/blob/master/calico_cni/policy_drivers.py#L368.  At the moment, any labels passed in via the network config file in this PR will be blatted over by labels from Kubernetes.  Not sure what behavior we want there (combine the labels maybe?)

All in all, we should spend some more time thinking about long term use-cases, goals, strategy before we merge this.  Once it's merged then it's there for a while.

calico_cni/policy_drivers.py
```diff
@@ -131,6 +131,14 @@ def generate_tags(self):
         """
         return []
 
+class MesosPolicyDriver(DefaultPolicyDriver):
+    def generate_rules(self):
+        allow = Rule(action="allow")
+        rules = Rules(id=self.profile_name,
+                      inbound_rules=[allow],
```
caseydavenport: I don't think this is correct - Mesos supports multiple CNI networks that need to be isolated from each other by default, so they shouldn't allow all inbound by default.

djosborne: There is no policy-on / policy-off switch for Mesos. I went with allow-all for a better demo story. 

I think the user would prefer standard mesos networking rules (everything talks to everything) and then lock down their app from there.

caseydavenport: I'm hesitant because it goes against the expectation of multiple CNI networks.  The only reason we do this in Kubernetes is because the Kubernetes networking model explicitly states it.  Is this the case in Mesos?

I think users will want to put a container in one network, another container in another network, and expect them not to be able to talk to each other.

calico_cni/constants.py
```diff
@@ -70,3 +71,9 @@
 LOG_DIR = "/var/log/calico/cni"
 LOG_FORMAT = '%(asctime)s %(process)d [%(identity)s] %(levelname)s %(message)s'
 
+# Mesos namespaced data.
+MESOS_NS_KEY = "org.apache.mesos"
```
caseydavenport: Doesn't matter for this PR, but I should note it anyway.  

The `calico-cni` code made a mistake of using way too many constants when we should have just used inline string values.  Ultimately, making a new constant for something that's only used once or twice in the code just makes it harder to understand what the code is doing.

These are fine though because it's in line with what the rest of the code does.

calico_cni/policy_drivers.py
```diff
@@ -131,13 +131,55 @@ def generate_tags(self):
...
+        the key and values and populate the labels dictionary.
+        """
+        self.labels = {}
+
+        labels_list = network_args[MESOS_NS_KEY]. \
+            get(MESOS_NETWORK_INFO_KEY, {}). \
+            get(MESOS_LABELS_OUTER_KEY, {}). \
+            get(MESOS_LABELS_KEY, [])
+
+        for label in labels_list:
```
caseydavenport: There's no way that Mesos might pass us multiple keys with the same value, right?

I'd assume not, but if it were theoretically possible then we'd at least want a warning in this code.  I say leave it as is, but maybe have a quick think to see if it might be possible.

calico_cni/policy_drivers.py
```diff
@@ -131,13 +131,55 @@ def generate_tags(self):
...
+    """
+    def __init__(self, network_name, network_args):
+        """
+        Extract any labels that have been specified in the mesos-namespaced labels field
+        of the network_args. Mesos CNI inserts the following additional data into
+        the args field:
+        "args" : {
+          "org.apache.mesos" : {
+            "network_info" : {
+              "name" : "mynet",
```
caseydavenport: For my own understanding, the value in "network_name" will be the same as in network_info["name"], correct?

tests/unit/test_policy_drivers.py
```diff
@@ -517,10 +565,24 @@ def test_get_policy_driver_value_error(self, m_driver):
...
 
         # Call
         with assert_raises(SystemExit) as err:
             get_policy_driver(cni_plugin)
         e = err.exception
         assert_equal(e.code, ERR_CODE_GENERIC)
 
+    def test_get_policy_driver_mesos(self):
+        cni_plugin = Mock(spec=CniPlugin)
+        cni_plugin.network_config = {"name": "testnetwork",
```
caseydavenport: I think ideally we'd want this to be an error case (since two policy types are specified).

I don't think we should write that code right now though...

calico_cni/constants.py
```diff
@@ -70,3 +71,9 @@
...
+MESOS_NETWORK_INFO_KEY = "network_info"
+MESOS_LABELS_OUTER_KEY = "labels"
+MESOS_LABELS_KEY = "labels"
+
```
caseydavenport: nit: Extra newline at the end of the file

calico.py
```diff
@@ -140,6 +140,17 @@ def __init__(self, network_config, env):
         Path in which to search for CNI plugins.
         """
 
+        self.running_under_mesos = bool(self.network_config.get("args", {}). \
```
caseydavenport: Any risk of both `running_under_mesos` and `running_under_k8s` being set at the same time?  I think it can only happen in a really unlikely user error case, so should be fine, but ideally I'd put some sort of checking mechanism in place to make sure we're not hitting weird code paths we didn't intend to ever be hit.

caseydavenport:
Some comments but I don't think you need to fix any of them.

LGTM.  Depending on your urgency, we still probably want to wait on the Go version before we cut the next release.  @tomdee is planning to merge the core Go code to master soon-ish, so you can add this function on top of that once it is in.

djosborne:
Closing in favor of [golang implementation](https://github.com/tomdee/calico-cni/pull/1)

Issue #107, opened 2016-06-10
Support empty CNI_IFNAME when/if it gets added to the CNI spec


caseydavenport:
This will be merged to upstream CNI soon: https://github.com/containernetworking/cni/pull/145

dcbw:
After #145 gets merged we'd need to relax pkg/skel's CNI_IFNAME check and add some testcases.  I think that's it...

caseydavenport:
@dcbw yep, when we switch this to Go this issue might just fall out, not sure.  Unfortunately at the moment this plugin is written in python and doesn't use pkg/skel so from the looks of it we'd need add support for empty CNI_IFNAME, a conflicting CNI_IFNAME, passing back interface information to the runtime, etc.

caseydavenport:
Plugin is now written in Go, so we shouldn't have any work to do on this end.

Issue #106, opened 2016-06-10
Don't require CNI_NETNS for delete calls


caseydavenport:
As per: https://github.com/containernetworking/cni/pull/230

Plugin currently expects this in all cases: https://github.com/projectcalico/calico-cni/blob/master/calico.py#L128

tomdee:
Fixed in Go version

Issue #94, opened 2016-04-20
Plugin doesn't handle multiple networks for single container correctly


tomdee:
When a second network is added, provided it's a calico network, we should just be updating the profiles and returning the existing IP address. However, we're returning the wrong IP. I think it's because of these lines
```
return {"ip4": {"ip": str(ip4.cidr)},
                "ip6": {"ip": str(ip6.cidr)}}
```
They shouldn't be return the cidr...

caseydavenport:
I thought the plugin already returned the IP addresses that were already assigned.  Is the problem here that we return a cidr, or that the IP address itself bad?

This should be something like `192.168.0.1/32`.

Is this only a problem in conjunction with #92?

tomdee:
Fixed in master

Issue #92, opened 2016-04-17
CNI plugin configures incorrect endpoint data in felix when using non Calico IPAM


tomdee:
When using the `host-local` IPAM plugin with the calico CNI plugin, the wrong endpoint information is written. e.g. 
```
{"name": "cali22d2946604b", "labels": {}, "state": "active", "ipv6_nets": [], "mac": "de:88:76:04:7b:df", "ipv4_nets": ["10.0.0.15/8"], "profile_ids": ["net1"]}
```

The subnet of the ipv4_nets should be /32. The correct netmask is used on the interface in the container, it's just the etcd data that's wrong. I'm not sure what the impact of thss is without more testing and looking through the felix code.

caseydavenport:
Presumably this is because the `host-local` IPAM is returning that prefix length, rather than the Calico CNI plugin generating that?

https://github.com/projectcalico/calico-cni/blob/master/calico.py#L379

tomdee:
Yes

tomdee:
Fixed in master

PR #108, opened 2016-06-10
Rewrite CNI plugin in Go

tomdee:
This PR replaces the Python Calico CNI plugin with a Go version.

It fixes the following bugs:
#68 #70 #86 #92 #94 and maybe #106 and #107

It removes the following features:
- K8s:
 - Annotation based policy
- logging 
 - file based logging (though we expect to add this back)

It adds the following features:
- Support for k8s username/password explicit config

It has the following isuues for the release note
    - include note about need to upgrade calico and calico-ipam at the same time
    - Users shouldn't need to change network config and should be able to upgrade a running system - 
    - Version output change
    - New default for k8s_api_root


calico.go
```diff
@@ -0,0 +1,366 @@
...
+
+func cmdAddK8s(args *skel.CmdArgs, k8sArgs K8sArgs, conf NetConf, theEndpoint *libcalico.Endpoint) (*types.Result, *libcalico.Endpoint, error) {
+	var err error
+	var result *types.Result
+
+	profileID := fmt.Sprintf("k8s_ns.%s", k8sArgs.K8S_POD_NAMESPACE)
+
+	if theEndpoint != nil {
+		// This happens when Docker or the node restarts. K8s calls CNI with the same details as before.
+		// Do the networking but no etcd changes should be required.
```
caseydavenport: I don't think this is true, is it?  We're going to set up a new veth with a new name and we need to tell Felix about that so it knows where to hang iptables rules.  Right?

calico.go
```diff
@@ -0,0 +1,366 @@
...
+			return nil, nil, err
+		}
+
+		// We're assuming that the endpoint is fine and doesn't need to be changed.
+		// However, the veth does need to be recreated since the namespace has been lost.
+		_, err := DoNetworking(args, conf, result)
+		if err != nil {
+			return nil, nil, err
+		}
+		// TODO - what if labels changed during the restart?
```
caseydavenport: Policy controller should still pick up label changes during the restart and update the Endpoint accordingly, so I don't think there's anything we need to do here?

calico.go
```diff
@@ -0,0 +1,366 @@
...
+
+	} else {
+		// There's no existing endpoint, so we need to do the following:
+		// 1) Call the configured IPAM plugin to get IP address(es)
+		// 2) Create the veth, configuring it on both the host and container namespace.
+		// 3) Configure the calico endpoint
+
+		// 1) run the IPAM plugin and make sure there's an IPv4 address
+		result, err = ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
```
caseydavenport: We should talk about whether we're going to continue supporting the "all-in-one" plugin/ipam binary or if we're going to start making people download two binaries.

calico.go
```diff
@@ -0,0 +1,366 @@
...
+
+		// 3) Update the endpoint
+		labels, err := GetK8sLabels(conf, k8sArgs)
+		if err != nil {
+			return nil, nil, err
+		}
+		theEndpoint = &libcalico.Endpoint{}
+		theEndpoint.Name = hostVethName
+		theEndpoint.Labels = labels
+		theEndpoint.ProfileID = []string{profileID}
```
caseydavenport: Minor note: ProfileID should probably have an S on the end since it is a list.

ipam/calico-ipam.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return err
+	}
+
+	ipamClient, err := ipam.NewIPAMClient()
+	if err != nil {
+		return err
+	}
+
+	//TODO - does this code need to fetch the pools or should ipamClient do it automatically.
```
caseydavenport: Plan is for the IPAMClient to do this automatically, though we can discuss if we think this is right.

ipam/calico-ipam.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/projectcalico/libcalico/lib/ipam"
+)
+
+func main() {
+	skel.PluginMain(cmdAdd, cmdDel)
+}
+
+// IPAMConfig represents the IP related network configuration.
+type IPAMConfig struct {
```
caseydavenport: Making a note: will need to support https://github.com/projectcalico/calico-cni/blob/master/ipam.py#L62

samples/kubernetes/cloud-config/master-config-template.yaml
```diff
@@ -1,179 +0,0 @@
-#cloud-config
```
caseydavenport: I think I agree with removing all the samples from this repo, but we should think about where the belong if we want to keep maintaining them (they are used by some of our guides)

utils/network.go
```diff
@@ -0,0 +1,96 @@
...
+		if err = netlink.RouteAdd(&netlink.Route{
+			LinkIndex: contVeth.Attrs().Index,
+			Scope:     netlink.SCOPE_LINK,
+			Dst:       ipn}); err != nil {
+			return fmt.Errorf("failed to add route %v", err)
+		}
+
+		_, defNet, _ := net.ParseCIDR("0.0.0.0/0")
+
+		if err = ip.AddRoute(defNet, gw, contVeth); err != nil {
```
caseydavenport: Doesn't Felix do this?

caseydavenport:
Not super happy with checking in the `vendor/` directory - is the reasoning just to cut down initial download time?  Can't remember if we reached a decision on this already...

calico.go
tomdee: discussed and agreed we won't

calico.go
tomdee: yep

calico.go
tomdee: I put up a libcalico PR for this

Makefile
```diff
@@ -1,92 +1,70 @@
...
+	-v ${PWD}/dist:/go/src/github.com/projectcalico/calico-cni/dist \
+	cni_container bash -c '\
+		make binary; \
+		chown -R $(shell id -u):$(shell id -u) dist'
 
-# Copy the plugin into place
-deploy-rkt: dist/calicoctl
-	cp dist/calico /etc/rkt/net.d
+cni_container.created: Dockerfile
+	docker build -t cni_container .
```
caseydavenport: I'd like to see this tag namespaced.

Something like `calico/cni_container`

Makefile
```diff
@@ -128,12 +106,55 @@ run-kube-proxy:
...
-
 dist/calicoctl:
-	mkdir -p dist
-	sudo chmod a+w dist
-	curl -o dist/calicoctl -L https://github.com/projectcalico/calico-containers/releases/download/v0.17.0/calicoctl
+	curl -o dist/calicoctl -L https://github.com/projectcalico/calico-containers/releases/download/v$(CALICO_NODE_VERSION)/calicoctl
 	chmod +x dist/calicoctl
+
+.PHONY: vendor-up
+vendor-up:
```
caseydavenport: Could you please add some comments describing what this and the other new makefile targets do? Why / when do I want to run this instead of just running "glide up"

Makefile
```diff
@@ -128,12 +106,55 @@ run-kube-proxy:
...
+	errcheck . ./ipam/... ./utils/...
+
+	# Check code style
+	-golint calico.go
+	-golint utils
+	-golint ipam
+
+dist/calico: $(SRCFILES)
+	CGO_ENABLED=0 go build -v --tags "$(BUILD_TAGS)" -o dist/calico \
+	-ldflags "-X main.VERSION=$(shell git describe --tags --dirty)" calico.go;
```
caseydavenport: ```-X main.VERSION=$(shell git describe --tags --dirty)```

I believe we also include the commit / branch in the version output, so we should try to do that in the Go version as well.  At least the commit ID (I know getting the correct branch for a release was awkward before).

Makefile
```diff
@@ -128,12 +106,55 @@ run-kube-proxy:
...
+
+dist/calico-ipam: ipam/calico-ipam.go
+	CGO_ENABLED=0 go build -o dist/calico-ipam  \
+	-ldflags "-X main.VERSION=$(shell git describe --tags --dirty)" ipam/calico-ipam.go;
+
+dist/host-local:
+	mkdir -p dist
+	curl -L https://github.com/containernetworking/cni/releases/download/v0.2.2/cni-v0.2.2.tgz | tar -zxv -C dist
+
+dist/calipo:
```
caseydavenport: This is just to get the Python version?

Maybe call this `calico-v1.3.1-python` or something instead so it's less cryptic, and add a comment.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+
+var hostname string
+
+func init() {
+	// this ensures that main runs only on main thread (thread group leader).
+	// since namespace ops (unshare, setns) are done for a single thread, we
+	// must ensure that the goroutine does not jump from OS thread to thread
+	runtime.LockOSThread()
+
+	hostname, _ = os.Hostname()
```
caseydavenport: I forget - did we decide the Python version respects the HOSTNAME environment variable?  We should check the behavior of this in Go and make sure it's documented in the config guide.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
+	. "github.com/projectcalico/calico-cni/utils"
+	"github.com/projectcalico/libcalico/lib"
+)
+
+var hostname string
+
+func init() {
+	// this ensures that main runs only on main thread (thread group leader).
```
caseydavenport: Capitalization.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+
+func createResultFromIP(ip string) (*types.Result, error) {
+	existingIPv4 := types.IPConfig{}
+	theIP := fmt.Sprintf(`{"ip": "%s"}`, ip)
+	err := existingIPv4.UnmarshalJSON([]byte(theIP))
+	if err != nil {
+		return nil, err
+	}
+	return &types.Result{IP4: &existingIPv4}, nil
+
```
caseydavenport: Extra newline.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+func init() {
+	// this ensures that main runs only on main thread (thread group leader).
+	// since namespace ops (unshare, setns) are done for a single thread, we
+	// must ensure that the goroutine does not jump from OS thread to thread
+	runtime.LockOSThread()
+
+	hostname, _ = os.Hostname()
+}
+
+func createResultFromIP(ip string) (*types.Result, error) {
```
caseydavenport: We need to support IPv6 here.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	existingIPv4 := types.IPConfig{}
+	theIP := fmt.Sprintf(`{"ip": "%s"}`, ip)
+	err := existingIPv4.UnmarshalJSON([]byte(theIP))
```
caseydavenport: Is `UnmarshalJSON` the best way to create an `IPConfig`? Why not this:

```go
existingIPv4 := types.IPConfig{IP: ip}
```

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	conf := NetConf{}
+	if err := json.Unmarshal(args.StdinData, &conf); err != nil {
+		return fmt.Errorf("failed to load netconf: %v", err)
+	}
+
+	if err := ValidateNetworkName(conf.Name); err != nil {
+		return err
+	}
+
+	etcd, err := libcalico.GetKeysAPI(conf.EtcdAuthority, conf.EtcdEndpoints)
```
caseydavenport: Is there a plan to stop exposing etcd through the libcalico API?

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return err
+	}
+
+	runningUnderK8s, err := RunningUnderK8s(args)
+	if err != nil {
+		return err
+	}
+
+	//var orchestratorID, workloadID string
```
caseydavenport: Commented out code - should be commented out comments instead :)

calico.go
```diff
@@ -0,0 +1,302 @@
...
+		if err != nil {
+			return err
+		}
+	} else {
+		// Not running in an environment that needs special casing.
+		workloadID := args.ContainerID
+		orchestratorID := "cni"
+		profileID := conf.Name
+
+		theEndpoint, err := libcalico.GetEndpoint(
```
caseydavenport: Why `theEndpoint` and not just `endpoint`, or `endpt`, or `ep`, or `e`?

calico.go
```diff
@@ -0,0 +1,302 @@
...
+		//TODO - this is the wrong test. It should be on the policy type
+		// Start by checking if the profile already exists. If it already exists then there is no work to do (the CNI plugin never updates a profile).
+		exists, err := libcalico.ProfileExists(conf.Name, etcd)
+		if err != nil {
+			return err
+		}
+
+		if !exists {
+			// The profile doesn't exist so needs to be created. The rules vary depending on whether k8s is being used.
+			// Under k8s (without full policy support) the rule is permissive and allows all traffic.
```
caseydavenport: I'd be interested in getting your opinion on this - I wonder if the CNI plugin should actually be making this judgement call.  Instead, users could just configure the profile as part of the install step (this makes it more clear what is going on and gives some more flexibility in terms of what the default policy is, it also means the profile processing for Kubernetes vs non-Kubernetes is the same).

I don't think this would affect any existing deployments.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	return result.Print()
+}
+
+func cmdDel(args *skel.CmdArgs) error {
+	n := NetConf{}
+	if err := json.Unmarshal(args.StdinData, &n); err != nil {
+		return fmt.Errorf("failed to load netconf: %v", err)
+	}
+
+	if err := AddIgnoreUnknownArgs(); err != nil {
```
caseydavenport: nit: Make this the first thing that occurs so that it matches the ADD code.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+				return err
+			}
+		}
+	}
+
+	return result.Print()
+}
+
+func cmdDel(args *skel.CmdArgs) error {
+	n := NetConf{}
```
caseydavenport: nit: Either rename this to "conf" or rename the ADD version to "n" so that they are consistent.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	if err := json.Unmarshal(args.StdinData, &n); err != nil {
+		return fmt.Errorf("failed to load netconf: %v", err)
+	}
+
+	if err := AddIgnoreUnknownArgs(); err != nil {
+		return err
+	}
+
+	// Always try to release the address
+	if err := ipam.ExecDel(n.IPAM.Type, args.StdinData); err != nil {
```
caseydavenport: We're short-circuiting if we encounter an error releasing the IP address.  Is this OK?

I _think_ we want to hold on to this error, and attempt to cleanup the endpoint as well.  Is there a way to return/combine multiple errors somehow?

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	k8sArgs := K8sArgs{}
+	if args.Args != "" {
+		err := LoadArgs(args.Args, &k8sArgs)
+		if err != nil {
+			return err
+		}
+	}
+	var orchestratorID, workloadID string
+
+	runningUnderK8s := string(k8sArgs.K8S_POD_NAMESPACE) != "" && string(k8sArgs.K8S_POD_NAME) != ""
```
caseydavenport: Let's make this function common.  It's used twice, and we'll soon have this for Mesos as well, which will want at least a new orchestrator ID, making this code fairly bulky.

e.g:
```
workloadID, orchestratorID := getIdentifiers()
```

caseydavenport: Do we actually need the `runningUnderK8s` variable?  Might it be better just to do `orchestratorID == "k8s"` (to prevent the inevitable `runningUnderK8s`, `runningUnderMesos`, `runningUnderWhatHaveYou` mess)  

What do you think?

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	// Always try to clean up the workload/endpoint.
+	// First determine if running under k8s to get the right workload and orchestrator IDs
+	k8sArgs := K8sArgs{}
+	if args.Args != "" {
+		err := LoadArgs(args.Args, &k8sArgs)
+		if err != nil {
+			return err
+		}
+	}
+	var orchestratorID, workloadID string
```
caseydavenport: It would be nice to define these closer to where they're assigned.  However, my comment below would make this moot.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// Always try to release the address
+	if err := ipam.ExecDel(n.IPAM.Type, args.StdinData); err != nil {
+		return err
+	}
+
+	// Always try to clean up the workload/endpoint.
+	// First determine if running under k8s to get the right workload and orchestrator IDs
+	k8sArgs := K8sArgs{}
+	if args.Args != "" {
```
caseydavenport: I don't think you need the `!= ""` check here (it's done in LoadArgs)

calico.go
```diff
@@ -0,0 +1,302 @@
...
+	flagSet := flag.NewFlagSet("Calico", flag.ExitOnError)
+
+	version := flagSet.Bool("v", false, "Display version")
+	err := flagSet.Parse(os.Args[1:])
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+	if *version {
+		fmt.Println(VERSION)
```
caseydavenport: This doesn't match the output format of the Python code (a dictionary).  I remember at least one user was checking the output of `-v` programmatically.  I think this is an incompatible change.

calico.go
```diff
@@ -0,0 +1,302 @@
...
+		OrchestratorID: orchestratorID,
+		WorkloadID:     workloadID}
+	if err := workload.Delete(etcd); err != nil {
+		return err
+	}
+
+	// Only try to delete the device if a namespace was passed in
+	if args.Netns != "" {
+		err := ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+			var err error
```
caseydavenport: Don't think you need to declare the variable `err` right here, unless I'm missing something.

caseydavenport: Do you actually need to declare `err` here?  It's in another scope, yeah?

ipam/calico-ipam.go
```diff
@@ -0,0 +1,125 @@
...
+	num6 := 0
+	if conf.AssignIpv6 != nil && *conf.AssignIpv6 == true {
+		num6 = 1
+	}
+
+	// TODO - Read the IP from CNI_ARGS and use it.
+	// TODO - Use the workloadID as the handle (i.e. need to know about k8s)
+	// TODO - plumb through hostname
+	// TODO - plumb through etcd auth
+	// TODO - confirm with Casey if HandleID really needs to be a pointer to a string.
```
caseydavenport: The Python code certainly differentiated between "None" and empty string, so I make the Go code understand that distinction as well.  

I think this will be needed for the upgrade case, but I haven't tested it yet.  Let's still chat about this when I'm in the office.

test_utils/utils.go
```diff
@@ -0,0 +1,134 @@
...
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/dchest/uniuri"
+	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega/gexec"
+)
+
+func RunIpam(netconf string) string {
```
caseydavenport: This guy is never used.

utils/k8s.go
```diff
@@ -0,0 +1,221 @@
...
+	if args.Args != "" {
+		err := LoadArgs(args.Args, &k8sArgs)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return string(k8sArgs.K8S_POD_NAMESPACE) != "" && string(k8sArgs.K8S_POD_NAME) != "", nil
+}
+
+func CmdAddK8s(args *skel.CmdArgs, k8sArgs K8sArgs, conf NetConf) (*types.Result, error) {
```
caseydavenport: I'd like to think and then chat a bit more about how we're structuring the different CmdAdd calls for different orchestrators.

Mostly:
- Ensuring we're not duplicating excessive amounts of code as we go on.
- Ensuring that the differences between orchestrators don't get lost in the abundance of common code between them.
- As an example / case study, how easy will it be to fit Mesos in to this model?

ipam/calico-ipam.go
tomdee: Done

Makefile
tomdee: Done

Makefile
tomdee: Done. It should now be possible to checkout the repo on a clean system (with docker installed) and just run `make`. That will populate the vendor directory, run a build and run the tests.

Makefile
tomdee: The current command produces output like `v1.3.1-5-g9197375-dirty` I think that's enough - it shows the most recent tag and the number of commits since (and the id of the last commit)

Makefile
tomdee: done

calico.go
tomdee: Python takes the hostname parameter in the netconf. I've now done this in Go too.

calico.go
tomdee: Done

calico.go
tomdee: Yes I think so, since `ip` is a string and `IP` is a net.IPNet

calico.go
tomdee: Not yet, but we should make one

calico.go
tomdee: Good question. Let's chat through this offline.

calico.go
tomdee: I've made the change to continue the delete processing and only return the IPAM if there were no other errors. let's discuss.

calico.go
tomdee: OK, I'll be sure to put it in the release note.

ipam/calico-ipam.go
tomdee: OK, let's chat.

test_utils/utils.go
tomdee: Sooon.

tomdee:
OK, I've applied markups. When you're free let's talk about
https://github.com/projectcalico/calico-cni/pull/108#discussion_r68864445
https://github.com/projectcalico/calico-cni/pull/108#discussion_r68863606
https://github.com/projectcalico/calico-cni/pull/108#discussion_r68862134
https://github.com/projectcalico/calico-cni/pull/108#discussion_r68866840
https://github.com/projectcalico/calico-cni/pull/108#discussion_r68866263

caseydavenport:
Looking generally pretty good to me.  Let's chat about those things when I'm in the office.

calico.go
tomdee: I think we agreed to leave it as it is

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.6-alpine
```
caseydavenport: This is going to conflict with master, where we've added a new `calico/cni` docker image for shipping CNI binaries.

Makefile
```diff
@@ -1,92 +1,141 @@
-.PHONY: all binary test plugin ipam ut clean update-version
+SRCFILES=$(wildcard *.go) $(wildcard utils/*.go) $(wildcard test_utils/*.go) ipam/calico-ipam.go
+LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | cut -d' ' -f8)
 
-# Version of calico/build to use.
-BUILD_VERSION=latest
+K8S_VERSION=1.2.4
```
caseydavenport: There are newer versions to use :)

Makefile
```diff
@@ -1,92 +1,141 @@
...
+CALICO_NODE_VERSION=0.19.0
 
-SRCFILES=$(shell find calico_cni -type f ! -path calico_cni/version.py) calico.py ipam.py
-LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | cut -d' ' -f8)
+# Ensure that the dist directory is always created
+MAKE_SURE_DIST_EXIST := $(shell mkdir -p dist)
 
-K8S_VERSION=1.2.0
+BUILD_CONTAINER_NAME=calico/cni_container
```
caseydavenport: `calico/cni_build_container` to distinguish from the `calico/cni` container?

Makefile
```diff
@@ -1,92 +1,141 @@
...
-binary: update-version dist/calico dist/calico-ipam
-test: ut fv
+all: vendor build-containerized test-containerized
+binary:  plugin ipam
 plugin: dist/calico
 ipam: dist/calico-ipam
 
+.PHONY: clean
+clean:
+	-sudo rm -rf dist vendor
```
caseydavenport: Does this need to have sudo?

Makefile
```diff
@@ -1,92 +1,141 @@
...
 
-# Builds the Calico CNI plugin binary.
-dist/calico: $(SRCFILES) 
-	docker run  --rm \
-	-v `pwd`:/code \
-	calico/build:$(BUILD_VERSION) \
-	/bin/sh -c "pip install pykube && pyinstaller calico.py -ayF"
+# Updates all the vendored libraries, i.e. changes the contents of the glide.lock. Running this should NOT be a common operation.
```
caseydavenport: Since this isn't a common operation, it should probably include a comment explaining when it should be run.

Makefile
```diff
@@ -1,92 +1,141 @@
...
+.PHONY: update-tools
+update-tools:
+	go get -u github.com/Masterminds/glide
+	go get -u github.com/kisielk/errcheck
+	go get -u golang.org/x/tools/cmd/goimports
+	go get -u github.com/golang/lint/golint
+
+# Perform static checks on the code. The golint checks are allowed to fail, the others must pass.
+.PHONY: static-checks
+static-checks: vendor
```
caseydavenport: Seems like most repos have a `make release` type option to create a new release that builds the binaries and runs any required tests / checks.  I think we should have that to make releases easier and more repeatable.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+			return err
+		}
+	} else {
+		// Not running in an environment that needs special casing.
+		profileID := conf.Name
+
+		if endpoint != nil {
+			// Don't create the veth or do any networking. Just update the profile on the endpoint. The profile will
+			// be created if needed during the profile processing step.
+
```
caseydavenport: nit: I think the first bit of comment in this branch should be "There is an existing endpoint - no need to create another.  This occurs when adding an existing container to a new CNI network" or something similar so that it is abundantly obvious what case this branch is handling.  The rest of these comments which explain what we actually do in this branch.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+	var result *types.Result
+
+	// If running under Kubernetes then branch off into the kubernetes code, otherwise handle everything in this
+	// function.
+	if orchestratorID == "k8s" {
+		if result, err = k8s.CmdAddK8s(args, conf, hostname, etcd, endpoint); err != nil {
+			return err
+		}
+	} else {
+		// Not running in an environment that needs special casing.
```
caseydavenport: nit: The comment here doesn't describe the code that follows (setting the profileID).  Could be fixed by rewording: "Default CNI behavior - use the CNI network name as the Calico profile."

calico.go
```diff
@@ -0,0 +1,265 @@
...
+		profileID := conf.Name
+
+		if endpoint != nil {
+			// Don't create the veth or do any networking. Just update the profile on the endpoint. The profile will
+			// be created if needed during the profile processing step.
+
+			// There's an existing endpoint - no need to create another. Find the IP address from the endpoint
+			// and use that in the response.
+			fmt.Fprintf(os.Stderr, "Calico CNI appending networking profile=%s\n", profileID)
```
caseydavenport: s/networking profile=/profile:

calico.go
```diff
@@ -0,0 +1,265 @@
...
+		etcd, libcalico.Workload{
+			Hostname:       hostname,
+			OrchestratorID: orchestratorID,
+			WorkloadID:     workloadID})
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(os.Stderr, "Calico CNI checking for existing endpoint. endpoint=%v\n", endpoint)
```
caseydavenport: Is this user facing message necessary?

I don't think we need to expose this.   Instead we could print out "creating new endpoint" vs "updating existing endpoint" based on the action we chose below.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+			fmt.Fprintf(os.Stderr, "Calico CNI appending networking profile=%s\n", profileID)
+			endpoint.ProfileIDs = append(endpoint.ProfileIDs, profileID)
+			result, err = CreateResultFromEndpoint(endpoint)
+			if err != nil {
+				return err
+			}
+		} else {
+			// There's no existing endpoint, so we need to do the following:
+			// 1) Call the configured IPAM plugin to get IP address(es)
+			// 2) Configure the calico endpoint
```
caseydavenport: nit: s/calico/Calico/g

calico.go
```diff
@@ -0,0 +1,265 @@
...
+			// 3) Create the veth, configuring it on both the host and container namespace.
+
+			// 1) run the IPAM plugin and make sure there's an IP address returned.
```
caseydavenport: nit: Capitalize beginning of sentences.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+			result, err = ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
+			if err != nil {
+				return err
+			}
+
+			// 2) Create the endpoint object
+			endpoint = &libcalico.Endpoint{ProfileIDs: []string{profileID}, OrchestratorID: orchestratorID,
+				WorkloadID: workloadID, Hostname: hostname, State: "active"}
+			if err = PopulateEndpointNets(endpoint, result); err != nil {
```
caseydavenport: I think we should be releasing the IP address if we encounter errors post-ip-assignment.  It prevents us from leaking and eventually running out of IP addresses if, for example, a single node is failing somehow and containers are being cyclically killed / re-scheduled.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+			endpoint.Name = hostVethName
+		}
+
+		// Write the endpoint object (either the newly created one, or the updated one with a new ProfileIDs).
+		if err := endpoint.Write(etcd); err != nil {
+			return err
+		}
+	}
+
+	// Handle profile creation - this is only done if there isn't a specific policy controller (like is sometimes
```
caseydavenport: nit: policy controller has another meaning in Kubernetes land.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+		exists, err := libcalico.ProfileExists(conf.Name, etcd)
+		if err != nil {
+			return err
+		}
+
+		if !exists {
+			// The profile doesn't exist so needs to be created. The rules vary depending on whether k8s is being used.
+			// Under k8s (without full policy support) the rule is permissive and allows all traffic.
+			// Otherwise, incoming traffic is only allowed from profiles with the same tag.
+			k8sInboundRule := []libcalico.Rule{libcalico.Rule{Action: "allow"}}
```
caseydavenport: I don't think there's a need to create `k8sInboundRule` and `tagInboundRule` in advance - just do it on lines 172 and 174.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+
+	// Allow the hostname to be overridden by the network config
+	if conf.Hostname != "" {
+		hostname = conf.Hostname
+	}
+
+	// Always try to release the address. Don't deal with any errors till the endpoints are cleaned up.
+	ipamErr := ipam.ExecDel(conf.IPAM.Type, args.StdinData)
+
+	// Always try to clean up the workload/endpoint.
```
caseydavenport: nit: comment doesn't match the following code.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+			// The profile doesn't exist so needs to be created. The rules vary depending on whether k8s is being used.
+			// Under k8s (without full policy support) the rule is permissive and allows all traffic.
+			// Otherwise, incoming traffic is only allowed from profiles with the same tag.
+			k8sInboundRule := []libcalico.Rule{libcalico.Rule{Action: "allow"}}
+			tagInboundRule := []libcalico.Rule{libcalico.Rule{Action: "allow", SrcTag: conf.Name}}
+			fmt.Fprintf(os.Stderr, "Calico CNI creating profile. profile=%s\n", conf.Name)
+
+			var inboundRule []libcalico.Rule
+			if orchestratorID == "k8s" {
+				inboundRule = k8sInboundRule
```
caseydavenport: nit: Should be named `inboundRules` since it is a slice.

calico.go
```diff
@@ -0,0 +1,265 @@
...
+			_, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+			return err
+		})
+
+		if err != nil {
+			return err
+		}
+	}
+
+	// Return the ipam_error if there was one. The ipam error will be lost if there was also an error in cleaning up
```
caseydavenport: ipam_error -> ipamErr (or IPAM error)

calico.go
```diff
@@ -0,0 +1,265 @@
...
+	// Always try to release the address. Don't deal with any errors till the endpoints are cleaned up.
+	ipamErr := ipam.ExecDel(conf.IPAM.Type, args.StdinData)
+
+	// Always try to clean up the workload/endpoint.
+	workloadID, orchestratorID, err := GetIdentifiers(args)
+	if err != nil {
+		return err
+	}
+
+	// Actually remove the workload
```
caseydavenport: Inconsistent punctuation in the comments.

ipam/calico-ipam.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/containernetworking/cni/pkg/types"
+	"github.com/projectcalico/calico-cni/utils"
+	"github.com/projectcalico/libcalico/lib/ipam"
+)
+
+func main() {
+	skel.PluginMain(cmdAdd, cmdDel)
+}
+
+// IPAMConfig represents the IP related network configuration.
```
caseydavenport: s/IPAMConfig/ipamConfig

ipam/calico-ipam.go
```diff
@@ -0,0 +1,146 @@
...
+
+	workloadID, _, err := utils.GetIdentifiers(args)
+	if err != nil {
+		return err
+	}
+
+	r := &types.Result{}
+	if conf.Args != nil && conf.Args.IP != nil {
+		fmt.Fprintf(os.Stderr, "Calico CNI IPAM request IP=%v\n", conf.Args.IP)
+		// hostname will be defaulted to the actual hostname if cong.Hostname is empty
```
caseydavenport: nits: punctuation, capitalization, and a newline before the comment.

ipam/calico-ipam.go
```diff
@@ -0,0 +1,146 @@
...
+		assignArgs := ipam.AssignIPArgs{IP: conf.Args.IP, HandleID: &workloadID, Hostname: conf.Hostname}
+		err := ipamClient.AssignIP(assignArgs)
+
```
caseydavenport: extra newline

ipam/calico-ipam.go
```diff
@@ -0,0 +1,146 @@
...
+		if err != nil {
+			return err
+		}
+
+		ipV4Network := net.IPNet{IP: conf.Args.IP, Mask: net.CIDRMask(32, 32)}
+		r.IP4 = &types.IPConfig{IP: ipV4Network}
+
```
caseydavenport: extra newline

ipam/calico-ipam.go
```diff
@@ -0,0 +1,146 @@
...
+	} else {
+
```
caseydavenport: extra newline

ipam/calico-ipam.go
```diff
@@ -0,0 +1,146 @@
...
+			ipV6Network := net.IPNet{IP: assignedV6[0], Mask: net.CIDRMask(128, 128)}
+			r.IP6 = &types.IPConfig{IP: ipV6Network}
+		}
+	}
+
+	return r.Print()
+}
+
+func cmdDel(args *skel.CmdArgs) error {
+	// Release by handle - which is workloadID.
```
caseydavenport: Comment doesn't match the following code block.

k8s/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/projectcalico/libcalico/lib"
+
+	"encoding/json"
+
+	"k8s.io/kubernetes/pkg/api"
+	k8sclient "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/kubernetes/pkg/client/unversioned/clientcmd"
+)
+
+// CmdAddK8s performs the "ADD" operation on a kubernetes pod
```
caseydavenport: Worth some more detail here - i.e why we need a separate add function for k8s.

k8s/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, etcd client.KeysAPI, endpoint *libcalico.Endpoint) (*types.Result, error) {
+	var err error
+	var result *types.Result
+
+	k8sArgs := utils.K8sArgs{}
+	if args.Args != "" {
```
caseydavenport: Isn't it an error if `args.Args == ""` and we're running in Kubernetes?

k8s/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		if conf.IPAM.Type == "host-local" && conf.IPAM.Subnet == "podCidr" {
+			// We've been told to use the "host-local" IPAM plugin with the Kubernetes podCidr for this node.
+			// Replace the actual value in the args.StdinData as that's what's passed to the IPAM plugin.
+			// args.StdinData
+
+			// the args.StdData should be changed
```
caseydavenport: Capitalization, punctuation.

k8s/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+	} else {
+		client, err := newK8sClient(conf)
+		if err != nil {
+			return nil, err
+		}
+
+		if conf.IPAM.Type == "host-local" && conf.IPAM.Subnet == "podCidr" {
+			// We've been told to use the "host-local" IPAM plugin with the Kubernetes podCidr for this node.
+			// Replace the actual value in the args.StdinData as that's what's passed to the IPAM plugin.
+			// args.StdinData
```
caseydavenport: Stray `// args.StdinData`

k8s/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+		labels, err := getK8sLabels(client, k8sArgs)
+		if err != nil {
+			return nil, err
+		}
+		endpoint.Labels = labels
+
+		fmt.Fprintf(os.Stderr, "Calico CNI using IPv4=%s IPv6=%s\n", endpoint.IPv4Nets, endpoint.IPv6Nets)
+	}
+
+	// Whether the endpoint existed or not, the veth needs recreating.
```
caseydavenport: s/recreating/creating

k8s/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+		if override.value != "" {
+			*override.variable = override.value
+		}
+	}
+
+	// Use the kubernetes client code to load the kubeconfig file and combine it with the overrides.
+	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeconfig},
+		configOverrides).ClientConfig()
+
```
caseydavenport: nit: extra newline.

k8s/k8s.go
```diff
@@ -0,0 +1,182 @@
...
+	return labels, nil
+}
+
+func getPodCidr(client *k8sclient.Client, k8sargs utils.K8sArgs, conf utils.NetConf, hostname string) (podCidr string, err error) {
+	nodeName := hostname
+
+	if conf.Kubernetes.NodeName != "" {
+		nodeName = conf.Kubernetes.NodeName
+	}
+	var node *api.Node
```
caseydavenport: I don't think you need to pre-declare the node var - it's only used within the `if` block below.

utils/utils.go
```diff
@@ -0,0 +1,106 @@
...
+	return workloadID, orchestratorID, nil
+}
+
+func PopulateEndpointNets(endpoint *libcalico.Endpoint, result *types.Result) error {
+	if result.IP4 == nil && result.IP6 == nil {
+		return errors.New("IPAM plugin did not return any IP addresses")
+	}
+
+	if result.IP4 != nil {
+		result.IP4.IP.Mask = net.CIDRMask(32, 32)
```
caseydavenport: Should these be using the `slash32` from `network.go`?

utils/network.go
```diff
@@ -0,0 +1,119 @@
...
+	err := ns.WithNetNSPath(netns, func(hostNS ns.NetNS) error {
+		hostVeth, contVeth, err := ip.SetupVeth(ifName, mtu, hostNS)
+
+		if err != nil {
+			return err
+		}
+
+		// Handle IPv4
+		if res.IP4 != nil {
+			// Add a connected route to a dummy next hop so that a default route can be set
```
caseydavenport: I think this code used to live in libcalico, yeah?  Can we add a TODO to your list to migrate this to libcalico so that all integrations can configure container routes in the same way?

utils/network.go
```diff
@@ -0,0 +1,119 @@
...
+		contVethMAC = contVeth.Attrs().HardwareAddr.String()
+
+		return nil
+	})
+
+	return hostVethName, contVethMAC, err
+}
+
+func setupHostVeth(vethName, newVethName string) error {
+	// hostVeth moved namespaces and may have a new ifindex
```
caseydavenport: nit: Comment is not very helpful.  Should go into a bit more detail - "The host side veth was originally created within the container netns and was subsequently moved to the host netns.  The name will be the same, but the ifindex may have changed so we must re-query it."

utils/network.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Select the first 11 characters of the containerID for the host veth
+	newHostVethName := "cali" + args.ContainerID[:min(11, len(args.ContainerID))]
+	if err = setupHostVeth(hostVethName, newHostVethName); err != nil {
+		return "", "", err
+	}
+	return newHostVethName, contVethMac, nil
+}
+
+func setupContainerVeth(netns, ifName string, mtu int, res *types.Result) (string, string, error) {
```
caseydavenport: This does more than setup the container veth - it also configures default routes within the container.

I think this, as well as setupHostVeth need more documentation in terms of what their respective responsibilities are.

Makefile
tomdee: I disagree. Let's talk.

calico.go
tomdee: That means repeating the log in both the kubernetes and "normal" paths. Do you still think I should do it?

calico.go
tomdee: Sounds like a very good idea, but none of the bundled CNI plugins have that behaviour. If we follow the principle of least surprise we should do the same as them. WDYT?

caseydavenport: Maybe I could be convinced... 

My initial reaction is that the other CNI plugins are doing the wrong thing and that we should do better.

Our current CNI plugin handles this, so in a sense it would be a regression.

Makefile
tomdee: Let's get this merged first then worry about the release process.

caseydavenport:
Can't remember what we decided about this, but leaving a not for now since I'm seeing it on the latest plugin:

> https://172.18.18.101:443/api/v1/api/v1/namespaces/kube-system/pods/kube-dns-v9-mj0rf

Looks like the plugin is appending `/api/v1` to the configured API root (which used to include the `/api/v1` bit.  This makes it incompatible with previous CNI configs.

ipam/calico-ipam.go
```diff
@@ -0,0 +1,127 @@
...
+		return err
+	}
+
+	// Release the IP address by using the handle - which is workloadID.
+	workloadID, _, err := utils.GetIdentifiers(args)
+	if err != nil {
+		return err
+	}
+
+	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
```
caseydavenport: We use the containerID as the handle above, and release using the workloadID here.  I think this works for rkt, but is broken for k8s!

caseydavenport:
One more comment in the IPAM code then LGTM.

PR #110, opened 2016-06-20
Mesos label support

robbrockbank:


djosborne:
I squashed these commits into one, then rebased over master (which after resolving a conflict, fixed the  `from calico_cni import __version__, __commit__, __branch__` imports), then added 1 commit which adds a default allow-all rule for the profile in Mesos (which was included in the mesoscon demo but left out of the pushed branch :disappointed: ) . Pushed those changes at #99 , so closing this.

PR #114, opened 2016-07-06
Add dockerfile for a "calicokube"

bogdando:
Signed-off-by: Bogdan Dobrelya <bdobrelia@mirantis.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM debian:jessie
+MAINTAINER projectcalico
+
+RUN mkdir -p /opt/cni/bin
+ADD ./dist/calico /opt/cni/bin
+ADD ./dist/calico-ipam /opt/cni/bin
+VOLUME /opt/cni
+ENV PATH=$PATH:/opt/cni/bin
+WORKDIR /opt/cni/bin
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 
 # Version of calico/build to use.
 BUILD_VERSION=latest
+CALICO_VERSION=$(shell git describe --tags)
 
 SRCFILES=$(shell find calico_cni -type f ! -path calico_cni/version.py) calico.py ipam.py
 LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | cut -d' ' -f8)
@@ -9,8 +10,9 @@ LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | cut -d' ' -f8)
 K8S_VERSION=1.2.0
 
 default: all
-all: binary test
+all: docker test
 binary: update-version dist/calico dist/calico-ipam
+docker: binary dist/docker
 test: ut fv
 plugin: dist/calico
 ipam: dist/calico-ipam
@@ -30,11 +32,16 @@ dist/calico-ipam: $(SRCFILES)
 	calico/build:$(BUILD_VERSION) \
 	pyinstaller ipam.py -ayF -n calico-ipam
 
+# Makes a docker image
+dist/docker:
+	docker build -t calico/cni .
+	docker tag calico/cni calico/cni:$(CALICO_VERSION)
+
 # Updates the version information in version.py
 update-version:
 	echo "# Auto-generated contents.  Do not manually edit" > calico_cni/version.py
 	echo "# or check in this file." >> calico_cni/version.py
-	echo "__version__ = '$(shell git describe --tags)'" >> calico_cni/version.py
+	echo "__version__ = '${CALICO_VERSION}'" >> calico_cni/version.py
 	echo "__commit__ = '$(shell git rev-parse HEAD)'" >> calico_cni/version.py
 	echo "__branch__ = '$(shell git rev-parse --abbrev-ref HEAD)'" >> calico_cni/version.py
 
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ To build the Calico Networking Plugin for CNI locally, clone this repository and
 - To just build the binaries, with no tests, run `make binary`. This will produce `dist/calico` and `dist/calico-ipam`.
 - To only run the unit tests, simply run `make ut`.
 - To only run the fv tests, simply run `make fv`.
+- To build the docker image, run `make docker`.
 
 [cni]: https://github.com/appc/cni
 [config]: configuration.md


caseydavenport:
@bogdando Awesome, thanks for this - I'll add it to my list to look at today.

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM calico/build:latest
```
caseydavenport: We can probably reduce the size of this container by using a different base image (e.g scratch).  I don't _think_ we need anything present in the `calico/build` container (though happy to be corrected)

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+MAINTAINER projectcalico
+
+ADD . /code
+WORKDIR /code
+
+ARG K8S_VERSION=local
+ENV K8S_VERSION ${K8S_VERSION}
+
+RUN pip install pykube && pyinstaller calico.py -ayF && \
```
caseydavenport: I think that rather than building the binaries within this container, I'd prefer using the binaries already built and placed in the `dist` directory.

So, something like this instead:
```
ADD ./dist/calico /opt/cni/bin
ADD ./dist/calico-ipam /opt/cni/bin
```

docker-build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh -ex
+# Build a calicokube image containing cni and ipam
+# for a given K8S VERSION env or $1 arg
+
+K8S="${1:-1.3.1}"
+K8S=${K8S:-$K8S_VERSION}
+docker build --build-arg K8S_VERSION=$K8S -t calicokube .
```
caseydavenport: Let's make the tag `calico/cni` to be more consistent with how we generally tag our images.

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM calico/build:latest
+MAINTAINER projectcalico
+
+ADD . /code
+WORKDIR /code
+
+ARG K8S_VERSION=local
```
caseydavenport: Could you explain what the K8S_ARG might be used for? In general, we expect CNI releases to apply across Kubernetes versions.

Dockerfile
caseydavenport: I suppose this does mean we can't do an automated DockerHub build, but I'm worried that the binaries built here could be different from the binaries actually included in a release.

docker-build.sh
caseydavenport: I think this logic could be moved into a `make` target, so you can simply "make docker-image" or something similar to trigger the build.

Dockerfile
bogdando: I can remove it if K8S_VERSION is not used for the build scripts

Dockerfile
bogdando: Will do. Not sure I got it to adding the binaries. Nothing to build when? What about libs shown by ldd for the binaries? Shall I put them in the container?

bogdando:
@caseydavenport maybe let's put it with the builder base for now, then update to something more fancy?

bogdando:
Ok, done!
Note, I put debian jessie to the base as the alpine base is missing glibc and other things ldd shows for calico binaries. It would be too much challenge for me to install those on alpine :/ see https://github.com/gliderlabs/docker-alpine/issues/78 -> https://github.com/stackhub/service-prometheus/blob/master/Dockerfile and that's only for libc, there are more!..

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM debian:jessie
+MAINTAINER projectcalico
+
+ADD ./dist /dist
```
caseydavenport: Sorry, very possible I'm just being daft but I want to understand a few things here:

- Why does this place the binaries in `/usr/sbin` rather than `/opt/cni/bin`, which is the default location for CNI binaries?
- Why does this install the binaries in two steps, rather than one? e.g `ADD ./dist/calico /opt/cni/bin`
- What is the significance of the `sync` - is that necessary?

caseydavenport:
Hmm. So I checked this code out to give it a try and I'm doing the following:

```
macncheese:calico-cni casey$ docker create -v /usr/sbin --name calcni calico/cni /bin/true
1f3f369e3382b36dfd79f443d706577058c6271b4edc7358346ccc4fe9218b30
macncheese:calico-cni casey$ docker run -ti --volumes-from calcni busybox "/bin/sh"
/ # /usr/sbin/calico -v
/bin/sh: /usr/sbin/calico: not found
/ #
```

Am I doing this right?  It seems like I can't execute the binary from within the container I've run even though it appears to have grabbed the volume properly.

Dockerfile
bogdando: good points, updated!

bogdando:
this how it works for me:
http://pastebin.com/eyjfYD8L

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM debian:jessie
+MAINTAINER projectcalico
+
+RUN mkdir -p /opt/cni/bin
+ADD ./dist/calico /opt/cni/bin
+ADD ./dist/calico-ipam /opt/cni/bin
+ENV PATH=$PATH:/opt/cni/bin
+WORKDIR /opt/cni/bin
```
mattymo: It would be nice if you added VOLUME /opt/cni so that a hyperkube container could just include --volumes-from calico-cni in its runtime params

caseydavenport:
Hmm.  I still can't seem to get the `--volumes-from` option to work.

```
macncheese:calico-cni casey$ docker run -tid --name calico-cni calico/cni:v1.3.1-7-g60626e3 /bin/sh
a032316127e55d1877f2b34f7c79461b5ce91eee166eb07796cc1afd37ddad43
macncheese:calico-cni casey$ docker run -ti --volumes-from calico-cni busybox "/bin/sh"
/ # /opt/cni/bin/calico
/bin/sh: /opt/cni/bin/calico: not found
/ #
```

It looks like the binaries are there:
```
/ # ls /opt/cni/bin/
calico       calico-ipam
/ #
```

@mattymo @bogdando - any idea what I'm doing wrong?

bogdando:
@caseydavenport there is no required libs in the image, try busybox:glibc maybe. Also note, that it works for me with debian image and volumes from. Other images may fail unless they contain all of the standard libs required for calico.

bogdando:
Any chances to get this accepted with the debian:jessie base?

caseydavenport:
@bogdando cool, that works.  And, more importantly, it works with the latest hyperkube image.

Thanks a lot for this - very exciting :)

PR #104, opened 2016-06-06
Switch Kubernetes version from 1.1.3 to 1.2.4

MikeSpreitzer:
1.2.4 is the latest on the 1.2 line.

PR #115, opened 2016-07-07
Switch Kubernetes version from 1.1.3 to 1.3.0

MikeSpreitzer:
Now that 1.3.0 is released.

The changes to ` samples/kubernetes/cloud-config/master-config-template.yaml` include an assumption that this new content will be tagged `k8s-1.3-docs` (like the current content is tagged `k8s-1.1-docs`).

### DIFF
--- a/samples/kubernetes/cloud-config/master-config-template.yaml
+++ b/samples/kubernetes/cloud-config/master-config-template.yaml
@@ -20,8 +20,8 @@ write_files:
         echo "Kubenetes not installed - installing."
 
         # Extract the Kubernetes binaries.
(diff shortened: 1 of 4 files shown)


caseydavenport:
LGTM, I can merge once the `license/cla` goes green :)

MikeSpreitzer:
What about tagging?  There is a `k8s-1.1-docs` tag but no analogue for 1.3.  And I would be willing to update #104 to k8s 1.2.5 to support a `k8s-1.2-docs` tag.

caseydavenport:
Yep, I'm going to tag this once https://github.com/projectcalico/calico-cni/pull/116 goes in which includes Calico version updates.

Personally I'm not too fussed about a 1.2-docs tag (we'll be recommending people use 1.3 anyway), but happy to make the tag if you think it will be useful.

PR #130, opened 2016-08-09
Golang fixes

caseydavenport:
Contains a few distinct commits:

- Update to latest libcalico to pick up various fixes there.
- Use the container ifname as the endpoint ID rather than a randomly generated value.
- Compatibility with the old-style k8s_api_root variable.


tomdee:
LGTM

PR #133, opened 2016-08-09
Fix up build dockerfile to include all requried deps

tomdee:


tomdee:
Green ticks so merging

PR #137, opened 2016-08-10
Release IP addresses on failure.

### LINKED ISSUE #128: Cleanup IPAM allocations if error occurs in CNI plugin
Applies to the Go plugin.

The first thing the CNI plugin does is allocate an IP address for the endpoint.

If creation of the veth (or something else) fails afterwards, that IP address is never cleaned up.  This can lead to _a lot_ of leaked IP allocations.

caseydavenport:
Fixes #128


tomdee:
LGTM

PR #140, opened 2016-08-15
Make podCidr comparison case insensitive

tomdee:


k8s/k8s.go
```diff
@@ -58,7 +58,7 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
 			return nil, err
 		}
 
-		if conf.IPAM.Type == "host-local" && conf.IPAM.Subnet == "podCidr" {
+		if conf.IPAM.Type == "host-local" && strings.EqualFold(conf.IPAM.Subnet, "podCidr") {
```
caseydavenport: I think this is why it wasn't working for me:

https://github.com/kubernetes/kubernetes/blob/89651077b1caee586b9d77e38a80139b3e75371a/cluster/saltbase/salt/calico/10-calico.conf#L8

The Python plugin expects the value `usePodCidr` rather than just `podCidr`

caseydavenport:
One thing to fix, then LGTM.

PR #141, opened 2016-08-15
Add structured logging support

tomdee:


tomdee:
I need to add logrus to the glide.yaml

calico.go
```diff
@@ -79,9 +94,13 @@ func cmdAdd(args *skel.CmdArgs) error {
 	if err != nil {
 		return err
 	}
+
+	logger.Debug("Retrieved endpoints: %v", endpoints)
+
 	var endpoint *api.WorkloadEndpoint
 	if len(endpoints.Items) == 1 {
 		endpoint = &endpoints.Items[0]
+		endpoint.Metadata.
```
caseydavenport: What is this?

calico.go
```diff
@@ -140,11 +161,13 @@ func cmdAdd(args *skel.CmdArgs) error {
 			endpoint.Metadata.Labels = labels
 			endpoint.Spec.Profiles = []string{profileID}
 
+			logger.WithField("endpoint", endpoint).Info("Populated endpoint (without nets)")
```
caseydavenport: I would have thought this should be a Debug log.

utils/utils.go
```diff
@@ -136,24 +138,40 @@ func CreateClient(conf NetConf) (*client.Client, error) {
...
 
-	if err := ipam.ExecDel(ipamType, stdinData); err != nil {
-		// Failed to cleanup the IP allocation.
-		glog.Warningf("Failed to clean up IP allocations for failed ADD")
-	}
+// Create a logger which always includes common fields
+func CreateContextLogger(args *skel.CmdArgs) *log.Entry {
+	// A common pattern is to re-use fields between logging statements by re-using
+	// the logrus.Entry returned from WithFields()
+	contextLogger := log.WithFields(log.Fields{
```
caseydavenport: I'd prefer this to be the WorkloadID - container ID isn't really useful in Kubernetes land.

utils/utils.go
```diff
@@ -136,24 +138,40 @@ func CreateClient(conf NetConf) (*client.Client, error) {
...
+		"ContainerID": args.ContainerID,
+		"NetNS":       args.Netns,
+	})
```
caseydavenport: Why don't we set the log level here from the value specified in the NetConf?  We already support a "log_level" field.  We can get rid of the "debug: true|false" one.

tomdee:
OK, this is still with me to update glide, add logs to k8s stuff (with workloadID) and use the log_level

tomdee:
@caseydavenport Should be ready for another look

configuration.md
```diff
@@ -25,20 +25,23 @@ If both are set then `etcd_endpoints` is used.
...
+    "log_level": "DEBUG",
     "ipam": {
         "type": "calico-ipam"
     }
 }
 ```
 
 ### IPAM
-When using Calico IPAM, the following flags determine what IP addresses should be assigned.
+When using Calico IPAM, the following flags determine what IP addresses should be assigned. NOTE: These flags are strings and not boolean values.
```
caseydavenport: :(

configuration.md
```diff
@@ -25,20 +25,23 @@ If both are set then `etcd_endpoints` is used.
 
 ### Logging
 * Logging is always to `stderr`
-* Additional logging can be enabled by setting `"debug"` to `true` in the netconf, e.g.
+* Logging level can be controlled by setting `"log_level"` in the netconf. Allowed levels are
+  * `WARNING` - the default
```
caseydavenport: According to code below, `INFO` is also supported.

caseydavenport:
One comment on configuration.md, then LGTM.

PR #142, opened 2016-08-15
Improve the rkt support in the Makefile

tomdee:


Makefile
```diff
@@ -129,8 +129,16 @@ dist/calicoctl:
 	chmod +x dist/calicoctl
 
 # Copy the plugin into place
-deploy-rkt: dist/calico
+deploy-rkt: binary
```
caseydavenport: Why the switch from depending on the file, to the phony target?

tomdee: So it builds the IPAM plugin too

tomdee:
@caseydavenport This is still with you

caseydavenport:
LGTM.

PR #148, opened 2016-08-18
Use -p so that etcd is accessible with Docker for Mac

caseydavenport:
This also makes `make test-containerized` work on mac.


tomdee:
LGTM

PR #153, opened 2016-08-19
Pass through version into build container

### LINKED ISSUE #147: Version number is wrong in the calico/cni container
The Go version relies on getting the tag from git, but the .git directory isn't passed in during the build.

The fix _isn't_ to pass in the .git directory though, the version should be set outside the container and passed in as a variable.

tomdee:
Fixes #147


caseydavenport:
LGTM

PR #162, opened 2016-08-25
Allow to run under Kubernetes without API access

tomdee:


k8s/k8s.go
```diff
@@ -109,7 +100,13 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
 		endpoint.Metadata.Hostname = hostname
 		endpoint.Metadata.OrchestratorID = orchestratorID
 		endpoint.Metadata.WorkloadID = workloadID
-		endpoint.Spec.Profiles = []string{profileID}
+		// Set the profileID according to whether Kubernetes policy is required. If it's not, then just use the network
```
caseydavenport: There should be a newline before this comment.

k8s/k8s.go
```diff
@@ -109,7 +100,13 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		// name (which is the normal behavior) otherwise use one based on the Kubernetes pod namespace.
```
caseydavenport: Should be "Kubernetes pod's Namespace" or "Kubernetes Namespace"

k8s/k8s.go
```diff
@@ -109,7 +100,13 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		if conf.Policy.PolicyType == "" {
```
caseydavenport: I think this if statement should be the other way around (so that it matches the label check below).

We only want to set the per-namespace profile if we're using `PolicyType == "k8s"`.

caseydavenport:
A few minor comments.  Once those are fixed, LGTM.

PR #163, opened 2016-08-29
Fix IPv6 support

tomdee:
Fetch the correct IPv6 address to use as next hop

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,6 +1,7 @@
 package: github.com/projectcalico/calico-cni
 import:
 - package: github.com/containernetworking/cni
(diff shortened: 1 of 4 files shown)


glide.yaml
```diff
@@ -1,6 +1,9 @@
 package: github.com/projectcalico/calico-cni
 import:
 - package: github.com/containernetworking/cni
+  version: SetupVeth
```
caseydavenport: Is there a PR we're waiting for here?

tomdee: Yes - https://github.com/containernetworking/cni/pull/291/files

caseydavenport:
LGTM, but needs to wait on the upstream CNI issue to be merged.

PR #166, opened 2016-09-01
Add CNI install script to docker image

tomdee:
Helps with K8s self hosted

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,11 @@
-FROM scratch
+FROM busybox
 
 MAINTAINER Tom Denham <tom@tigera.io>
 
 ADD dist/calico /opt/cni/bin/calico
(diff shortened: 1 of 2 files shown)


caseydavenport:
LGTM.

Dockerfile
```diff
@@ -1,9 +1,11 @@
-FROM scratch
+FROM busybox
 
 MAINTAINER Tom Denham <tom@tigera.io>
 
 ADD dist/calico /opt/cni/bin/calico
 ADD dist/calico-ipam /opt/cni/bin/calico-ipam
+ADD k8s-install/scripts/install-cni.sh /install-cni.sh
```
ae6rt: Is the point of this to install CNI to the host filesystem via the action of a  short-lived container?

caseydavenport: Yep.

PR #170, opened 2016-09-08
Add retry for veth rename if the device is busy

tomdee:


utils/network.go
```diff
@@ -114,15 +118,26 @@ func setupHostNetworking(vethName, newVethName string) error {
...
-
 	if err := netlink.LinkSetDown(veth); err != nil {
 		return fmt.Errorf("failed to set %q DOWN: %v", vethName, err)
 	}
 
+	for i := 0; i < 10; i++ {
+		if err := netlink.LinkSetName(veth, newVethName); err != nil {
+			switch err {
+			case syscall.EBUSY:
+				time.Sleep(100 * time.Millisecond)
```
caseydavenport: I think this warrants a debug level log.

utils/network.go
```diff
@@ -114,15 +118,26 @@ func setupHostNetworking(vethName, newVethName string) error {
...
+				continue
+			default:
+				return err
+			}
+		}
+		return nil
+	}
+
 	if err := netlink.LinkSetName(veth, newVethName); err != nil {
```
caseydavenport: Why is this still here?

utils/network.go
```diff
@@ -114,15 +118,26 @@ func setupHostNetworking(vethName, newVethName string) error {
...
 		return fmt.Errorf("failed to rename veth: %v to %v (%v)", vethName, newVethName, err)
 	}
-
 	if err := netlink.LinkSetUp(veth); err != nil {
```
caseydavenport: We won't hit this anymore, right?  Since we're returning above?

utils/network.go
```diff
@@ -114,15 +118,26 @@ func setupHostNetworking(vethName, newVethName string) error {
 	if err != nil {
 		return fmt.Errorf("failed to lookup %q: %v", vethName, err)
 	}
-
 	if err := netlink.LinkSetDown(veth); err != nil {
 		return fmt.Errorf("failed to set %q DOWN: %v", vethName, err)
 	}
 
+	for i := 0; i < 10; i++ {
```
caseydavenport: And a comment explaining why we need this whole retry loop thing.

caseydavenport:
LGTM.

PR #172, opened 2016-09-08
Add API resource details

### LINKED ISSUE #168: Document needed permission for kubernetes
When running kubernetes in [ABAC authorization](http://kubernetes.io/docs/admin/authorization/) mode resource permissions need to be explicitly given.

After reading the source code of the new 1.4.0 rc I suspect that calico-cni needs readonly access to Pods and if configured to use the podCIDR of k8s it will also need readonly access to Nodes.  This information should be added to the configuration [readme](https://github.com/projectcalico/calico-cni/blob/master/configuration.md).

caseydavenport:
Fixes #168

### DIFF
--- a/configuration.md
+++ b/configuration.md
@@ -102,6 +102,8 @@ There is a single supported policy type, `k8s` which uses the Kubernetes Network
 }
 ```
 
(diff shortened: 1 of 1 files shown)


tomdee:
Is this ready for review? If so, please assign it to me.

PR #174, opened 2016-09-11
Strip for smaller binaries

tomdee:


caseydavenport:
LGTM - are we worried at all that we're stripping debug information - i.e you can't use gdb to debug anymore after this?  I suspect not.

tomdee:
Not sure about GDB. I do know that enough debug information is preserved for the stack traces produced by panics to still look normal (i.e. line numbers are still there)

PR #180, opened 2016-09-19
Fix nil labels field getting into etcd when not using k8s policy

dpindur:
If the policy type isn't "k8s" the endpoint labels metadata never gets initialised and ends up as null in etcd. It looks like this causes some issues when felix tries to read from it. Making sure it's initialised fixes this.

Spoken to @caseydavenport and he mentioned this used to be handle in https://github.com/projectcalico/calico-cni/blob/master/k8s/k8s.go#L222 but https://github.com/projectcalico/calico-cni/commit/34577483abe1d69c410657c9a44ae11dff46bbd9#diff-b396ca928d5bd0e161acd283b56ead9fL121 made this optional and inadvertently introduced the bug.


caseydavenport:
OK - Looks good to me.  I've checked out and ran the tests manually.

PR #182, opened 2016-09-30
Support templating the etcd TLS assets in the calico/cni container

caseydavenport:
For self-hosted access to TLS enabled etcd clusters.

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -8,8 +8,42 @@
 # Ensure all variables are defined.
 set -u
 
-# Clean up any existing binaries. 
+# The directory on the host where CNI networks are installed. Defaults to 
+# /etc/cni/net.d, but can be overridden by setting CNI_NET_DIR.  This is used
+# for populating absolute paths in the CNI network config to assets
+# which are installed in the CNI network config directory.
+HOST_CNI_NET_DIR=${CNI_NET_DIR:-/etc/cni/net.d}
+HOST_SECRETS_DIR=${HOST_CNI_NET_DIR}/calico-tls
+
+# Directory where we expect that TLS assets will be mounted into 
+# the calico/cni container.
+SECRETS_MOUNT_DIR=${TLS_ASSETS_DIR:-/calico-secrets}
+
+# Clean up any existing binaries / config / assets.
 rm -f /host/opt/cni/bin/calico /host/opt/cni/bin/calico-ipam
+rm -f /host/etc/cni/net.d/calico-tls/*
+
+# Copy over any TLS assets from the SECRETS_MOUNT_DIR to the host.
+echo "Installing any TLS assets from ${SECRETS_MOUNT_DIR}"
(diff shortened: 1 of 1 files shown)


k8s-install/scripts/install-cni.sh
```diff
@@ -51,17 +85,30 @@ sed -i s/__KUBERNETES_SERVICE_PORT__/${KUBERNETES_SERVICE_PORT:-}/g calico.conf.
...
 # Unless told otherwise, sleep forever.
 # This prevents Kubernetes from restarting the pod repeatedly.
 should_sleep=${SLEEP:-"true"}
 echo "Done configuring CNI.  Sleep=$should_sleep"
 while [ "$should_sleep" == "true"  ]; do
+	# Kubernetes Secrets can be updated.  If so, we need to install the updated
+	# version to the host. Just check the timestamp on the certificate to see if it 
+	# has been updated.  A bit hokey, but likely good enough.
+	stat_output=$(stat -c%y ${SECRETS_MOUNT_DIR}/etcd-cert 2>/dev/null)
 	sleep 1;
```
tomdee: Should you sleep for a little longer? 2 stats a second might be a bit much...

k8s-install/scripts/install-cni.sh
```diff
@@ -51,17 +85,30 @@ sed -i s/__KUBERNETES_SERVICE_PORT__/${KUBERNETES_SERVICE_PORT:-}/g calico.conf.
...
+sed -i s~__ETCD_CERT_FILE__~${CNI_CONF_ETCD_CERT:-}~g calico.conf.tmp
+sed -i s~__ETCD_KEY_FILE__~${CNI_CONF_ETCD_KEY:-}~g calico.conf.tmp
+sed -i s~__ETCD_CA_CERT_FILE__~${CNI_CONF_ETCD_CA:-}~g calico.conf.tmp
 sed -i s~__ETCD_ENDPOINTS__~${ETCD_ENDPOINTS:-}~g calico.conf.tmp
 
 # Move the temporary CNI config into place.
-mv calico.conf.tmp /host/etc/cni/net.d/${CNI_CONF_NAME:-10-calico.conf}
-echo "Wrote CNI config: $(cat /host/etc/cni/net.d/10-calico.conf)"
+FILENAME=${CNI_CONF_NAME:-10-calico.conf}
+mv calico.conf.tmp /host/etc/cni/net.d/${FILENAME}
```
tomdee: Moves are only atomic on the same filesystem. Maybe the temp file should be written to the same location as the final version.

caseydavenport: Hm.  I wanted to leave this here since I see a future where we might need to copy this to one of multiple potential mount locations.  I'm not really too worried about this.  Should I be?

tomdee: It's tricky to reason about. I suspect we're OK because the file is small, and a present-but-empty or half-written file will just get rejected and K8s will retry (and spot that the file is now complete. 

So - I think I'm OK leaving this as is.

tomdee:
LGTM

PR #184, opened 2016-10-06
Make fully containerized build

artem-panchenko:
Run 'vendor' part inside Docker container in
case of containerized build.

Change-Id: I6564ef4a7debb2735e8c1c847161442287a4b8d4


tomdee:
Thanks @artem-panchenko for the contribution - LGTM so merging.

I have a slight reservation about the choice of docker image for glide - I'd prefer an alpine based one for size and one that's more frequently updated. But we can always improve that later.

PR #185, opened 2016-10-09
Update libcalico references from tigera -> projectcalico

caseydavenport:


caseydavenport:
Closing as dup of #186

PR #187, opened 2016-10-17
Update 'libcalico-go' version

artem-panchenko:
The history of the reposirory 'libcalico-go' was
re-written, so the commit 0ddf9de2807feb4ef2953d78fab33e0bc78d2fdb
doesn't exist there anymore. Pinned 'libcalico-go' version
to the current master.

Change-Id: I1ab7daf307594acf19748359b17e376c8c611a18

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -111,7 +111,7 @@ imports:
 - name: github.com/spf13/pflag
   version: 6fd2ff4ff8dfcdf5556fbdc0ac0284408274b1a7
 - name: github.com/projectcalico/libcalico-go
-  version: 0ddf9de2807feb4ef2953d78fab33e0bc78d2fdb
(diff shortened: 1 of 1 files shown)


robbrockbank:
Hi @artem-panchenko, 

I'm actually just looking at updating calico-cni as well - the history re-write has additional knock on effects that need to be handled - namely the handling of endpoint removal.  I'll close this PR and raise one that fixes all of the related knock-on effects.

artem-panchenko:
Okay, thank you!

robbrockbank:
Hi @artem-panchenko :  Apologies for the delay, but I needed to check with our CNI expert before I could get the change merged.  The CNI code should now be updated to handle the latest revision of the libcalico-go library.

PR #189, opened 2016-10-19
Update glide lock file and handle workload endpoint deletion correctly

robbrockbank:
Updated glide to point to latest libcalico-go.

Also - need to update how workload endpoints are deleted.  Lib calico now deletes the workload when the last endpoint is deleted.


tomdee:
LGTM - merging

PR #191, opened 2016-10-20
Use local 'libcalico-go' for containerized build

artem-panchenko:
Add a possibility to use custom 'libcalico-go' repo
location while building binaries inside container:

 - bind additional directory with local copy of the
   repository to '/go/src/github.com/projectcalico/libcalico-go'
   folder inside Docker container;
 - modify 'glide.lock', so 'libcalico-go' module is
   taken from local path when binary is built within container
   'repo: file:///go/src/github.com/projectcalico/libcalico-go'.

With this change it's possible to build CNI binaries and
test them in order to verify changes to the 'libcalico-go'.

Change-Id: Id8466eed7f86c4870fbba35fe6d12872678dc45e


tomdee:
LGTM

PR #190, opened 2016-10-20
Check for number of returned IPs from IPAM

mihgen:
Current implementation of libcalico-go does not return any error if
no IPs were allocated. This patch fixes a crash listed in the bug
report.
There is a consideration to return an error from libcalico-go, if no
or not all requested IPs were allocated. Please see details here:
https://github.com/projectcalico/libcalico-go/issues/202

Closes #188

PR #192, opened 2016-10-20
Updated ref libcalico-go

mihgen:
libcalico-go git history seems to be rewritten, so tests do not pass
anymore. Updating ref to the latest patches, before serious changes
were introduced (e.g. remove calicoctl).
We need to get this build green first, and then make a further update
here to point to the latest libcalico-go version.

This patch should fix tests failing in #190.


robbrockbank:
HI @mihgen - we merged a fix for this last night - the cni should now be using the latest libcalico-go revision, there were some additional minor changes required alongside the ref update.

Closing this issue as it should no longer be required.

PR #207, opened 2016-11-14
Run all CI targets inside Docker containers

artem-panchenko:
Now it's possible to run static checks (Golang code)
inside container, so all make targets for CI can
be executed within Docker.

Change-Id: I9f8ac37909db712ee97c914b7766336889dfb875


caseydavenport:
Sorry for the delay on this, LGTM.

PR #211, opened 2016-11-18
Adding CNI binaries

heschlie:
Added host-local, loopback, and flannel CNI binaries, primarily to get Calico hosted deployment working with Hyperkube installs (e.g. CoreOS-k8s)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ BUILD_CONTAINER_NAME=calico/cni_build_container
 BUILD_CONTAINER_MARKER=cni_build_container.created
 DEPLOY_CONTAINER_NAME=calico/cni
 DEPLOY_CONTAINER_MARKER=cni_deploy_container.created
-TEMP_DIR:=$(shell mktemp -d)
 
 LIBCALICOGO_PATH?=none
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -89,10 +90,20 @@ test-watch: dist/calico dist/calico-ipam
 	sudo CGO_ENABLED=0 ETCD_IP=127.0.0.1 PLUGIN=calico GOPATH=$(GOPATH) $(shell which ginkgo) watch
 
 $(BUILD_CONTAINER_MARKER): Dockerfile.build
+	mkdir -p dist
```
tomdee: Do you really need these in the build container?

Makefile
```diff
@@ -16,6 +16,7 @@ BUILD_CONTAINER_NAME=calico/cni_build_container
 BUILD_CONTAINER_MARKER=cni_build_container.created
 DEPLOY_CONTAINER_NAME=calico/cni
 DEPLOY_CONTAINER_MARKER=cni_deploy_container.created
+TEMP_DIR:=$(shell mktemp -d)
```
tomdee: This means we'll leak temp dirs right?

Makefile
```diff
@@ -89,10 +90,20 @@ test-watch: dist/calico dist/calico-ipam
...
+	curl -sSL --retry 5 https://github.com/containernetworking/cni/releases/download/v0.3.0/cni-v0.3.0.tgz | tar -xz -C $(TEMP_DIR)
+	mv ${TEMP_DIR}/flannel dist/flannel
+	mv ${TEMP_DIR}/loopback dist/loopback
+	mv ${TEMP_DIR}/host-local dist/host-local
 	docker build -f Dockerfile.build -t $(BUILD_CONTAINER_NAME) .
 	touch $@
 
 $(DEPLOY_CONTAINER_MARKER): Dockerfile build-containerized
+	mkdir -p dist
+	curl -sSL --retry 5 https://github.com/containernetworking/cni/releases/download/v0.3.0/cni-v0.3.0.tgz | tar -xz -C $(TEMP_DIR)
```
tomdee: I'm not sure about the version you're choosing? It's a reasonable choice, but Kubernetes do use their own (later) build. Maybe you should check what's being used currently in the hyperkube and use that same version?

Makefile
tomdee: I'm slightly on the fence about mentioning this, but you could consider doing the binary inside the deploy container. Then  the build stays more containerized and there's work to do to clean things up or avoid leaks.

Makefile
heschlie: Whoops, no it does not

Makefile
heschlie: yes, but I figured being a temp dir it would not make much impact.  I'm open to just putting into a local dir and adding it to the make clean if that is preferable?

k8s-install/scripts/install-cni.sh
```diff
@@ -46,7 +46,7 @@ then
 fi
 
 # Place the new binaries.
-cp /opt/cni/bin/calico /host/opt/cni/bin/calico
+cp /opt/cni/bin/* /host/opt/cni/bin/
 cp /opt/cni/bin/calico-ipam /host/opt/cni/bin/calico-ipam
```
caseydavenport: This line is redundant now

Makefile
heschlie: Would that all be done in the Dockerfile?  Would RUN be the best way to get the tar pulled and extraced?

PR #217, opened 2016-11-23
add KUBERNETES_NODE_NAME param

liubog2008:
support kubelet arg `--hostname-override`


caseydavenport:
@liubog2008 looks good to me, thanks!

PR #221, opened 2016-11-25
Ensure directories exist before making any manipulations

SergK:
We need to create directories for several targets


fasaxc:
LGTM

PR #222, opened 2016-11-25
Build our own golang-build container

SergK:
Use our own golang-build container for building go code into a
binary. The original idea came from felix project [1]

[1]. https://github.com/projectcalico/felix/blob/master/Makefile


SergK:
I've updated the description with issues that can be resolved with this patch

 Use our own golang-build container for building go code into a

binary.

1. The original idea came from felix project [1]
2. Get rid of running container under root and as the result
   have correct permissions in ~/.glide cache directory on CI node
3. In isolated environment /vendor directory can be pupulated
   outside of `calico build` process (no access to 3rd party repos),
   then we can just put it and update target
   `touch vendor/.up-to-date` manually, e.g.
```
       cp -r /prepopulated_vendor calico-cni/vendor
       touch vendor/.up-to-date
       make build-containerized
```
[1]. https://github.com/projectcalico/felix/blob/master/Makefile

SergK:
I think we can close this PR in favor of https://github.com/projectcalico/calico-cni/pull/232

PR #223, opened 2016-11-28
Add support for the new IPv4Pools and IPv6Pools properties on AutoAssign

duglin:
Now that https://github.com/projectcalico/libcalico-go/pull/271 is merged.

Bump refs to libcalico-go too

Add pod annotations to enable them:
```
    spec:
      template:
        metadata:
          annotations:
            cni.calico/ipv4pools: [<ip-pool-1>, <ip-pool-2>]
            cni.calico/ipv6pools: [<ip-pool-1>, <ip-pool-2>]
```

Signed-off-by: Doug Davis <dug@us.ibm.com>

### DIFF
--- a/calico_cni_ipam_test.go
+++ b/calico_cni_ipam_test.go
@@ -83,6 +83,87 @@ var _ = Describe("Calico IPAM Tests", func() {
 		})
 	})
 
+	Describe("Run IPAM plugin - Verify IP Pools", func() {
+		Context("Pass valid pools", func() {
+			It("successfully networks the namespace", func() {
+				netconf := fmt.Sprintf(`
+			        {
+			          "name": "net1",
+			          "type": "calico",
+			          "etcd_endpoints": "http://%s:2379",
+			          "ipam": {
+			            "type": "%s",
+			            "assign_ipv4": "true",
+						"ipv4_pools": [ "192.168.0.0/16" ]
+			          }
+			        }`, os.Getenv("ETCD_IP"), plugin)
+				_, _, _, _, _, _, err := CreateContainer(netconf)
+				Expect(err).ShouldNot(HaveOccurred())
+			})
+		})
+
+		Context("Pass more than one pool", func() {
+			It("successfully networks the namespace", func() {
+				PreCreatePool("192.169.1.0/24")
+				netconf := fmt.Sprintf(`
+			        {
+			          "name": "net1",
+			          "type": "calico",
+			          "etcd_endpoints": "http://%s:2379",
+			          "ipam": {
+			            "type": "%s",
+			            "assign_ipv4": "true",
+						"ipv4_pools": [ "192.169.1.0/24", "192.168.0.0/16" ]
+			          }
+			        }`, os.Getenv("ETCD_IP"), plugin)
+				_, _, session, _, _, _, err := CreateContainer(netconf)
+				if err != nil {
+					fmt.Printf("Session Err: %v\n", string(session.Err.Contents()))
+				}
+				Expect(err).ShouldNot(HaveOccurred())
+			})
+		})
+
+		Context("Pass an invalid pool", func() {
+			It("fails to network the namespace", func() {
+				// Put the bogus pool last in the array
+				netconf := fmt.Sprintf(`
+			        {
+			          "name": "net1",
+			          "type": "calico",
+			          "etcd_endpoints": "http://%s:2379",
+			          "ipam": {
+			            "type": "%s",
+			            "assign_ipv4": "true",
+						"ipv4_pools": [ "192.168.0.0/16", "192.169.1.0/24" ]
+			          }
+			        }`, os.Getenv("ETCD_IP"), plugin)
+				_, _, session, _, _, _, err := CreateContainer(netconf)
+				Expect(err).Should(HaveOccurred())
+				Expect(session.Err.Contents()).Should(ContainSubstring("192.169.1.0/24) does not exist"))
+			})
+
+			It("fails to network the namespace", func() {
+				// Put the bogus pool first in the array
+				netconf := fmt.Sprintf(`
+			        {
+			          "name": "net1",
+			          "type": "calico",
+			          "etcd_endpoints": "http://%s:2379",
+			          "ipam": {
+			            "type": "%s",
+			            "assign_ipv4": "true",
+						"ipv4_pools": [ "192.169.1.0/24", "192.168.0.0/16" ]
+			          }
+			        }`, os.Getenv("ETCD_IP"), plugin)
+				_, _, session, _, _, _, err := CreateContainer(netconf)
+				Expect(err).Should(HaveOccurred())
+				Expect(session.Err.Contents()).Should(ContainSubstring("192.169.1.0/24) does not exist"))
+			})
+		})
+
+	})
+
 	Describe("Run IPAM plugin", func() {
 		netconf := fmt.Sprintf(`
 					{"name": "net1",
--- a/calico_cni_ipam_test.go
+++ b/calico_cni_ipam_test.go
@@ -25,7 +25,7 @@ var _ = Describe("Calico IPAM Tests", func() {
 			DescribeTable("Request different numbers of IP addresses",
 				func(expectedIPv4, expectedIPv6 bool, netconf string) {
 
-					result, _ := RunIPAMPlugin(netconf, "ADD", "")
+					result, _, _ := RunIPAMPlugin(netconf, "ADD", "")
 
 					if expectedIPv4 {
 						Expect(result.IP4.IP.Mask.String()).Should(Equal("ffffffff"))
@@ -36,7 +36,7 @@ var _ = Describe("Calico IPAM Tests", func() {
 					}
 
 					// I can't find any testable side effects for this
-					_, _ = RunIPAMPlugin(netconf, "DEL", "")
+					_, _, _ = RunIPAMPlugin(netconf, "DEL", "")
 				},
 				Entry("IPAM with no configuration", true, false, fmt.Sprintf(`
 			{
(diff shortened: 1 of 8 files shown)


caseydavenport:
@tomdee I don't think this is a blocker for v2.0.0 since it is a feature add.

duglin:
Marking this as WIP because I added support for hooking this into K8s via annotation and I haven't had a chance to test it yet.  However, if people could look at the code in k8s/k8s.go to see if it looks right I'd appreciate it.

duglin:
ok I think this is ready for some eyes on it

utils/utils.go
```diff
@@ -232,3 +232,23 @@ func CreateContextLogger(workload string) *log.Entry {
...
+		_, cidr, err := net.ParseCIDR(p)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing pool %q: %s", p, err)
+		}
+		ip := cidr.IP
+		if isv4 && ip.To4() == nil {
+			return nil, fmt.Errorf("%q isn't a IPv4 address", ip)
+		}
+		if !isv4 && ip.To4() != nil {
+			return nil, fmt.Errorf("%q isn't a IPv16 address", ip)
```
tomdee: v16? Now there's a very futuristic IP protocol!

k8s/k8s.go
```diff
@@ -103,6 +103,54 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		if conf.Policy.PolicyType == "k8s" {
+			var err error
+
+			labels, annot, err = getK8sLabelsAnnotations(client, k8sArgs)
+			if err != nil {
+				// Cleanup IP allocation and return the error.
+				utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
+				return nil, err
+			}
+			logger.WithField("labels", labels).Info("Fetched K8s labels")
```
tomdee: If there are a large number of labels or annotations this could result is many large logs - I wonder if debug would be a better level?

cmluciano:
I can work with @duglin to write some tests for k8tes if you can share your framework.

calico_cni_ipam_test.go
```diff
@@ -83,6 +83,87 @@ var _ = Describe("Calico IPAM Tests", func() {
...
+			It("successfully networks the namespace", func() {
+				netconf := fmt.Sprintf(`
+			        {
+			          "name": "net1",
+			          "type": "calico",
+			          "etcd_endpoints": "http://%s:2379",
+			          "ipam": {
+			            "type": "%s",
+			            "assign_ipv4": "true",
+						"ipv4_pools": [ "192.168.0.0/16" ]
```
duglin: fixed

k8s/k8s.go
duglin: I can't expect there would be that many to cause an issue, but switched anyway.

utils/utils.go
duglin: I think that's called future proofing. :-)
Fixed.

duglin:
@tomdee comments addressed and rebased

tomdee:
@duglin / @cmluciano I've just up a PR for an initial k8s test #235 

Hopefully I can get it tidied/merged soon, then you'll be able to add test(s) for your new and updated code paths.

tomdee:
Looks like there are some conflicts in your glide.lock

duglin:
hmm, git and github aren't complaining - where are you seeing that?

tomdee:
Hmmm, maybe I was dreaming it. It looks OK now.

tomdee:
@duglin / @cmluciano - My k8s test framework changes are merged, so you should be able to add a test for your code now. I expect you to be able to add a test that's similar to this - https://github.com/projectcalico/calico-cni/blob/master/calico_cni_k8s_test.go#L77 - but to vary the annotations on the pod to hit your code.

duglin:
@tomdee thanks!  Still need to catch-up on the other thread too about possibly changing what's passed in.....

duglin:
@tomdee I'm running into a problem. I added the start of a k8s test, following your guidance - thanks.  But, when trying to test it I'm not seeing what I expected to see. In particular, my annotations seemed to have no effect, so I started to debug it and got to the point where I added a panic() at https://github.com/projectcalico/cni-plugin/pull/223/files#diff-b396ca928d5bd0e161acd283b56ead9fR114 just to see if that if-statement was being called at all - and I don't think it is. In fact, it doesn't look like that code in k8s/k8s.go is called at all.  Another interesting thing is that when I print out the Pod we get back from the Pod.Create() call, all of the IP fields are empty.  Does the testing infrastructure invoke k8s and the calico stuff or does it fake some of it?

duglin:
Oh wait - it could be timing thing.... checking....

duglin:
ok - it may be a timing issue but I never see the Pod leave the "Pending" state. When I look at the docker containers running I see the API server and etcd. Shouldn't there be others running - like the kubelet?

caseydavenport:
```
cni.calico/ipv4pools: [<ip-pool-1>, <ip-pool-2>]
```

One thought - Kubernetes best practice indicates that annotations should be part of a domain owned by the project, so this should probably be `cni.projectcalico.org/ipv4pools`, or maybe even `ipam.cni.projectcalico.org/ipv4pool`

@tomdee WDYT?

tomdee:
@duglin The k8s testing that I added is quite "interesting". the kubelet doesn't actually invoke the CNI plugin, we still do that manually. In fact, there's not even a kubelet running. However, the test framework does run a kubernetes API server, and manually injects the podspec into it. I _think_ this shouil dbe enough to still enable some useful testing. It does explain why you see pods stuck in pending state though, since there's nothing to actually run the pod, they'll never move out of pending.

tomdee:
@caseydavenport do you have a link to the k8s best practices. I can't see anything relevant about annotations in https://kubernetes.io/docs/user-guide/config-best-practices/

In general, I agree with your suggestion of putting the domain in the annotation though. Do you know if it's "normal" to reverse it for k8s annotations though (e.g. `org.projectcalico.abc`)

calico_cni_k8s_test.go
```diff
@@ -53,11 +53,10 @@ var _ = Describe("CalicoCni", func() {
 			    "subnet": "10.0.0.0/8"
 			  },
 				"kubernetes": {
-					        "k8s_api_root": "http://127.0.0.1:8080"
-
```
tomdee: Broken indent

calico_cni_k8s_test.go
```diff
@@ -53,11 +53,10 @@ var _ = Describe("CalicoCni", func() {
 			    "subnet": "10.0.0.0/8"
 			  },
 				"kubernetes": {
-					        "k8s_api_root": "http://127.0.0.1:8080"
-									    
+                  "k8s_api_root": "http://127.0.0.1:8080"
 				},
 				"policy": {"type": "k8s"},
-	"log_level":"info"
```
tomdee: Broken indent

calico_cni_k8s_test.go
```diff
@@ -82,6 +84,26 @@ var _ = Describe("CalicoCni", func() {
...
 					panic(err)
 				}
+
+				// Now create a K8s pod passing in an IP pool
+				name2 := name + "-pool"
+				pod, err := clientset.Pods(K8S_TEST_NS).Create(&v1.Pod{
+					ObjectMeta: v1.ObjectMeta{
+						Name: name2,
+						Annotations: map[string]string{
+							"cni.calico/ipv4pools": "192.169.1.0/24",
```
tomdee: This annotation won't be used since you're calling CreateContainer below with `name` not `name2`

glide.lock
```diff
@@ -1,5 +1,5 @@
 hash: 1efa10f7d9f9793128d78496760c641fb6ee82b2686a4fd476ba83a624f1506c
-updated: 2016-12-13T15:51:57.719804486-08:00
```
tomdee: Do you really need to update the glide.lock?

caseydavenport:
@tomdee Interestingly, neither can I.  However, this section on label syntax does mention the format that I was thinking of: https://kubernetes.io/docs/user-guide/labels/#syntax-and-character-set

It's easy get a picture of the standard notation from other well-known annotations, e.g:

```
scheduler.alpha.kubernetes.io/critical-pod:
scheduler.alpha.kubernetes.io/tolerations:
net.alpha.kubernetes.io/network-policy:
```

calico_cni_k8s_test.go
duglin: fixed

calico_cni_k8s_test.go
duglin: "log_level" looks ok to me, but I'll recheck after the next push

calico_cni_k8s_test.go
duglin: fixed, I hope. Indentation in backticks is gonna kill me ;-)

glide.lock
duglin: I think I had to update the glide.yaml at one point, which forced the lock file to change. But I think that requirement is now behind me so yea I'll try to remove this from the PR.

duglin:
re: annotations and forward/reverse : I've seen it both ways in Kube, and been told to do it both ways too.  For now I've picked `ipam.cni.projectcalico.org/...`  let me know if you want it changed.

@tomdee do you have some sample code I can use as a guide for how to run the CNI plug-in manually in the k8s testing setup?  In the meantime, I've disable the code that waits for the pod to get into the right state, so the test will fail for now.

calico_cni_k8s_test.go
```diff
@@ -83,6 +85,7 @@ var _ = Describe("CalicoCni", func() {
 					panic(err)
 				}
 				containerID, netnspath, session, contVeth, contAddresses, contRoutes, err := CreateContainer(netconf, name)
```
duglin: @tomdee this CreateContainer() call is confusing me. Didn't the call to clientset.Pods().Create() create the Pod and the underlying container already?

tomdee: clientset.Pods().Create() just creates the API object in the kubernetes API server,
CreateContainer() actually invokes the CNI plugin binary directly. It's NOT the kublet that invokes. This test code invokes it directly passing in similar information to what kubernetes would.

caseydavenport:
@tomdee @duglin what do we need to do to get this guy merged?

duglin:
first, let me apologize for this looking like its lingering... I've just been swamped. I think the outstanding issue here is the testing side of things. I think I have the basic structure there, but I've not be able to get the CNI code to be invoked to verify my code works. I'll try to find some time over the next day or so to see if I can push this over the finish line.

caseydavenport:
> first, let me apologize for this looking like its lingering... I've just been swamped

No worries, we're pretty swamped too :)  

Just want to make sure you're not waiting on us for something and that if you are, we prioritize unblocking you.

duglin:
I don't think I'm blocked on anything other than a possible lack of education on how to get the CNI code invoked, but before I cry for more help I feel I need to play around with it some more first.

djosborne:
Are there any other docs changes planned for this feature besides the ones here: https://github.com/projectcalico/calico/pull/275

Do there need to be? Is there a Kubernetes use case this was added for?

tomdee:
@djosborne I'm not aware of any plans and I don't think we'd ask @duglin to add any more. I can see that this would be a great feature to add more docs for though so maybe we should raise a docs issue to track doing that?

k8s/k8s.go
```diff
@@ -103,6 +103,54 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		// if the policy type has been set to "k8s". This allows users to
+		// run the plugin under Kubernetes without needing it to access the
+		// Kubernetes API
+		if conf.Policy.PolicyType == "k8s" {
+			var err error
+
+			labels, annot, err = getK8sLabelsAnnotations(client, k8sArgs)
+			if err != nil {
+				// Cleanup IP allocation and return the error.
+				utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
```
gunjan5: Why do we need to ReleaseIP here? `ADD` happens later so not sure I see why we need to release IP here.

k8s/k8s.go
```diff
@@ -103,6 +103,54 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
 
+		// Only used by K8s so if its null then we're not doing k8s stuff
+		var labels map[string]string
+		var annot map[string]string
+
+		// Only attempt to fetch the labels and annotations from Kubernetes
+		// if the policy type has been set to "k8s". This allows users to
+		// run the plugin under Kubernetes without needing it to access the
+		// Kubernetes API
+		if conf.Policy.PolicyType == "k8s" {
```
gunjan5: I think we should also add a check `conf.IPAM.Type == "host-local"` here to make sure this code only gets called for `calico-ipam` so it doesn't set fields that are not relevant to other IPAMs.

k8s/k8s.go
```diff
@@ -103,6 +103,54 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			logger.WithField("labels", labels).Debug("Fetched K8s labels")
+			logger.WithField("annotations", annot).Debug("Fetched K8s annotations")
+
+			v4pools := annot["ipam.cni.projectcalico.org/ipv4pools"]
+			v6pools := annot["ipam.cni.projectcalico.org/ipv6pools"]
+
+			if len(v4pools) != 0 || len(v6pools) != 0 {
+				var stdinData map[string]interface{}
+				if err := json.Unmarshal(args.StdinData, &stdinData); err != nil {
+					utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
```
gunjan5: same as above. Not sure if we need to release IP here

k8s/k8s.go
```diff
@@ -103,6 +103,54 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+
+				if len(v4pools) > 0 {
+					fmt.Fprintf(os.Stderr, "Calico CNI setting ipv4pools to %q", v4pools)
+				}
+				if len(v6pools) > 0 {
+					fmt.Fprintf(os.Stderr, "Calico CNI setting ipv6pools to %q", v6pools)
+				}
+				newData, err := json.Marshal(stdinData)
+				if err != nil {
+					utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
```
gunjan5: and here

k8s/k8s.go
```diff
@@ -103,6 +103,54 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			v4pools := annot["ipam.cni.projectcalico.org/ipv4pools"]
+			v6pools := annot["ipam.cni.projectcalico.org/ipv6pools"]
+
+			if len(v4pools) != 0 || len(v6pools) != 0 {
+				var stdinData map[string]interface{}
+				if err := json.Unmarshal(args.StdinData, &stdinData); err != nil {
+					utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
+					return nil, err
+				}
+				stdinData["ipam"].(map[string]interface{})["ipv4pools"] = v4pools
```
gunjan5: shouldn't this be `ipv4_pools` instead of `ipv4pools` since NetConf's `IPv4Pools` field processes json by the string `ipv4_pools` 
```
`IPv4Pools     []string `json:"ipv4_pools"`
```

Same with `ipv6pools` in the line below.

gunjan5:
Changes moved to #260 . Closing this

PR #225, opened 2016-11-30
Fix version marker checks for make release

heschlie:
Check was always hitting the exit 1 causing it to fail


tomdee:
The change looks fine, but there' shouldn't be 3 commits in this PR

PR #228, opened 2016-12-06
Builds/Tests: Remove calicoctl and etcdctl

tomdee:
Simplify the build by not relying on the calicoctl binary anymore
And simplify the tests by using libcalico-go for assertions

### DIFF
--- a/test_utils/utils.go
+++ b/test_utils/utils.go
@@ -2,6 +2,7 @@ package test_utils
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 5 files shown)


tomdee:
Fixes #206

test_utils/utils.go
```diff
@@ -16,13 +17,58 @@ import (
...
+		panic(err)
+	}
+
+	// Create a new client.
+	calicoClient, err := client.New(*clientConfig)
+	if err != nil {
+		panic(err)
+	}
+
+	_, ipNet, err := calnet.ParseCIDR(pool)
```
robbrockbank: FWIW - there are testutils in libcalico-go which includes a MustParseCIDR and MustParseIP that return the appropriate types for the Calico API.

robbrockbank: We also have a few "hokey" etcd methods that used to use etcdctl to clean etcd.  Might have been nice to update the test utilities.

PR #229, opened 2016-12-06
Builds: Disable circle CI

tomdee:
Running on Circle as well as semaphore brings very little benefit


matthewdupre:
By very little, do you mean they run the same tests?  If so SGTM.

PR #233, opened 2016-12-07
Support multiple dirs for install script

caseydavenport:
Allows a user put two directories in the self-hosted manifest, and we'll install to whichever ones are writeable.

This makes the manifest more portable across multiple clusters where the directory might be read-only on some OSes


tomdee:
LGTM - do you want to squash then we can merge?

PR #239, opened 2016-12-13
Smarter fetching of 3rd party CNI binaries

heschlie:
Added target for fetching CNI binaries, and marker to avoid re-download when they exist.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,14 +102,11 @@ $(DEPLOY_CONTAINER_MARKER): Dockerfile build-containerized fetch-cni-bins
 	touch $@
 
 .PHONY: fetch-cni-bins
-fetch-cni-bins:
+fetch-cni-bins: dist/flannel dist/loopback dist/host-local
+
+dist/flannel dist/loopback dist/host-local:
 	mkdir -p dist
-	mkdir -p tmp-cni
-	$(CURL) -L --retry 5 https://github.com/containernetworking/cni/releases/download/v0.3.0/cni-v0.3.0.tgz | tar -xz -C tmp-cni/
-	mv tmp-cni/flannel dist/flannel
-	mv tmp-cni/loopback dist/loopback
-	mv tmp-cni/host-local dist/host-local
-	rm -rf tmp-cni/
+	$(CURL) -L --retry 5 https://github.com/containernetworking/cni/releases/download/v0.4.0/cni-amd64-v0.4.0.tgz | tar -xz -C dist ./flannel ./loopback ./host-local
 
 # Run the tests in a container (as root). Useful for CI but currently slow for 
 # local development because the .go-pkg-cache can't be used (since tests run as root)
@@ -183,11 +180,6 @@ static-checks: vendor
 
 install:
 	CGO_ENABLED=0 go install github.com/projectcalico/cni-plugin
-
-# Retrieve a host-local plugin for use in the tests
-dist/host-local:
-	mkdir -p $(@D)
-	$(CURL) -L https://github.com/containernetworking/cni/releases/download/v0.2.2/cni-v0.2.2.tgz | tar -zxv -C dist
 
 # Retrieve an old version of the Python CNI plugin for use in tests
 dist/calico-python:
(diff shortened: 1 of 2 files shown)


tomdee:
host-local is still being downloaded in two places with two versions - https://github.com/heschlie/calico-cni/blob/b08a4bcc822fdc722223f039b035eae764607800/Makefile#L191

tomdee:
I'm still a little unsure that yo're using the best release for CNI - I think if you check what kubernetes uses you'll find they use a non official release

tomdee:
I would prefer it if we avoided using an additional marker file for this. The existence of the plugins under `dist/` already tells us that they've been downloaded. What's your justification for needing a marker?

heschlie:
I does look like kubernetes/hyperkube is using their own build

```
ARCH?=amd64
TEMP_DIR:=$(shell mktemp -d -t hyperkubeXXXXXX)
CNI_RELEASE=07a8a28637e97b22eb8dfe710eeae1344f69d16e
...
https://storage.googleapis.com/kubernetes-release/network-plugins/cni-${ARCH}-${CNI_RELEASE}.tar.gz
```

it points to a particular PR which was accepted after the latest release, we could just pull using their link I suppose?

As to my reasoning for using a marker is mostly my ignorance on how to do stuff with Makefiles, I'll do some digging and make it mo'betta.

Do we need to keep that host-local at the version it has?  Right now it is pulling v0.2.2 but the one we might want to pull for hyperkube uses a PR post v0.3.0 unless I'm mistaken (very possible)

tomdee:
The existing host-local for the tests can be changed, the version we use for that is _much_ less important than the version we ship onto people's systems when they use our install scripts.

caseydavenport:
@tomdee @heschlie what's the status of this? Is it being worked on or should it be closed?

tomdee:
It's with @heschlie still.

heschlie:
@tomdee thanks for the help this morning, finally some more changes for your review!

Makefile
```diff
@@ -102,14 +103,15 @@ $(DEPLOY_CONTAINER_MARKER): Dockerfile build-containerized fetch-cni-bins
 	touch $@
 
 .PHONY: fetch-cni-bins
-fetch-cni-bins:
+fetch-cni-bins: dist/flannel dist/loopback dist/host-local
+
+dist/flannel dist/loopback dist/host-local:
 	mkdir -p dist
 	mkdir -p tmp-cni
```
tomdee: I don't think you need this temp dir. Just extract the files to the right place

`tar -xzf cni-amd64-v0.4.0.tgz -C dist ./flannel ./loopback ./host-local` (but using curl of course)

Makefile
```diff
@@ -33,7 +34,7 @@ docker-image: $(DEPLOY_CONTAINER_MARKER)
 
 .PHONY: clean
 clean:
-	rm -rf dist vendor $(DEPLOY_CONTAINER_MARKER) .go-pkg-cache
+	rm -rf dist vendor $(DEPLOY_CONTAINER_MARKER) .go-pkg-cache tmp-cni
```
tomdee: Can avoid tmp-cni (see below)

Makefile
```diff
@@ -18,6 +18,7 @@ MAKE_SURE_DIST_EXIST := $(shell mkdir -p dist)
 CALICO_BUILD?=calico/go-build
 DEPLOY_CONTAINER_NAME=calico/cni
 DEPLOY_CONTAINER_MARKER=cni_deploy_container.created
+FETCH_CNI_BINS_MARKER=fetch_cni_bins.created
```
tomdee: Are you using this marker?

Makefile
heschlie: oh nice, I like that

heschlie:
@tomdee I removed that marker, there was no need for it anymore, and am directly untaring into dist, much cleaner.

tomdee:
Cool, please squash then I can merge.

heschlie:
OK @tomdee they are squashed... in a manner of speaking... thanks for the help, again :)

PR #247, opened 2017-01-05
Supply a default CNI config in the calico/cni container.

### LINKED ISSUE #243: Include a default CNI config
We could simplify our self-hosted manifests a lot by including a default CNI config file.

For most deployments I think the same config can be used.  People could still override by mounting in a ConfigMap.

caseydavenport:
Fixes #243


tomdee:
LGTM

PR #248, opened 2017-01-14
Makefile: Use calico/go-build

tomdee:
- For vendoring
- For static checks
- For build binaries (and get a nice speedup from package cache)
- For tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+# Disable make's implicit rules, which are not useful for golang, and slow down the build
+# considerably.
+.SUFFIXES:
+
 SRCFILES=calico.go $(wildcard utils/*.go) $(wildcard k8s/*.go) ipam/calico-ipam.go
 TEST_SRCFILES=$(wildcard test_utils/*.go) $(wildcard calico_cni_*.go)
 LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
@@ -11,14 +15,13 @@ CALICO_CNI_VERSION?=$(shell git describe --tags --dirty)
 
 # Ensure that the dist directory is always created
 MAKE_SURE_DIST_EXIST := $(shell mkdir -p dist)
-
-GO_CONTAINER_NAME?=dockerepo/glide
-BUILD_CONTAINER_NAME=calico/cni_build_container
-BUILD_CONTAINER_MARKER=cni_build_container.created
+CALICO_BUILD?=calico/go-build
 DEPLOY_CONTAINER_NAME=calico/cni
 DEPLOY_CONTAINER_MARKER=cni_deploy_container.created
 
 LIBCALICOGO_PATH?=none
+
+LOCAL_USER_ID?=$(shell id -u $$USER)
 
 .PHONY: all binary plugin ipam
 default: all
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -59,26 +62,28 @@ endif
...
-      --entrypoint /bin/sh $(GO_CONTAINER_NAME) -e -c ' \
-	cd /go/src/github.com/projectcalico/cni-plugin && \
-	glide install -strip-vendor && \
-	chown $(shell id -u):$(shell id -u) -R vendor'
+	  -v $(CURDIR):/go/src/github.com/projectcalico/cni-plugin:rw $$EXTRA_DOCKER_BIND \
+		-v $(HOME)/.glide:/home/user/.glide:rw \
+		-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
+		$(CALICO_BUILD) /bin/sh -c ' \
+			cd /go/src/github.com/projectcalico/cni-plugin && \
+			glide install -strip-vendor'
```
gunjan5: looks like we're downloading glide `v0.12.x` to our build container, and `--strip-vendor` is not necessary for `v0.11+`  i.e. it's the default option (I think, can't find the reference to it anymore)

gunjan5: also, shouldn't it be 2 `-`'s for `--strip-vendor`?

Makefile
```diff
@@ -59,26 +62,28 @@ endif
...
 	mkdir -p $(@D)
-	CGO_ENABLED=0 go build -v -o dist/calico \
+	CGO_ENABLED=0 go build -v -i -o dist/calico \
 	-ldflags "-X main.VERSION=$(CALICO_CNI_VERSION) -s -w" calico.go
 
 # Build the Calico ipam plugin
 dist/calico-ipam: $(SRCFILES) vendor
 	mkdir -p $(@D)
-	CGO_ENABLED=0 go build -v -o dist/calico-ipam  \
+	CGO_ENABLED=0 go build -v -i -o dist/calico-ipam  \
```
gunjan5: curious what the `-i` option changes here? According to the docs:
> The -i flag installs the packages that are dependencies of the target.

wondering how it was working without this option before

Makefile
```diff
@@ -284,7 +269,7 @@ run-kube-proxy:
 	-docker rm -f calico-kube-proxy
 	docker run --name calico-kube-proxy -d --net=host --privileged gcr.io/google_containers/hyperkube:v$(K8S_VERSION) /hyperkube proxy --master=http://127.0.0.1:8080 --v=2
 
-ci: clean static-checks-containerized test-containerized docker-image
+ci: clean static-checks test-containerized docker-image
```
gunjan5: not totally related, but can we rename the `docker-image` target to something more relevant?

gunjan5:
few minor comments/questions. Tested it on my mac and it seems to work fine

Makefile
tomdee: I think strip vendor _is_ still needed. ANd one dash works just fine (but I can update it to two)

```
docker run --rm calico/go-build glide install --help 
Starting with UID : 9001
NAME:
   glide install - Install a project's dependencies

USAGE:
   glide install [command options] [arguments...]

DESCRIPTION:
   This uses the native VCS of each packages to install
   the appropriate version. There are two ways a projects dependencies can
   be installed. When there is a glide.yaml file defining the dependencies but
   no lock file (glide.lock) the dependencies are installed using the "update"
   command and a glide.lock file is generated pinning all dependencies. If a
   glide.lock file is already present the dependencies are installed or updated
   from the lock file.

OPTIONS:
   --force             If there was a change in the repo or VCS switch to new one. Warning: changes will be lost.
   --strip-vendor, -v  Removes nested vendor and Godeps/_workspace directories.
   --skip-test         Resolve dependencies in test files.
```

Makefile
tomdee: -i ensures that that the `.a` files are placed in the `go-pkg-cache` directory. It's only needed to speed up builds.

Makefile
tomdee: We can, but not in this PR :smile:

PR #255, opened 2017-01-31
Bugfix: IP leak. Clean up IP address with host-local IPAM

### LINKED ISSUE #254: Fail to cleanup IP addresses when using "usePodCidr"
I noticed that we're not substituting the "subnet: usePodCidr" on DEL when using the host-local IPAM plugin, which means we end up leaking IPs.

```
time="2017-01-30T05:16:05Z" level=error msg="invalid CIDR address: usePodCidr"
```

gunjan5:
Fixes #254

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 
 	"github.com/vishvananda/netlink"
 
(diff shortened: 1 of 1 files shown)


calico.go
```diff
@@ -284,6 +285,28 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		dummyPodCidr := "192.0.2.0/24"
+		var stdinData map[string]interface{}
+
+		if err := json.Unmarshal(args.StdinData, &stdinData); err != nil {
+			return nil, err
+		}
+
+		logger.WithField("podCidr", dummyPodCidr).Info("Using a dummy podCidr to release the IP")
+		stdinData["ipam"].(map[string]interface{})["subnet"] = dummyPodCidr
+		fmt.Fprintf(os.Stderr, "Calico CNI passing dummyPodCidr to host-local IPAM for releasing the pod IP address: %s\n", dummyPodCidr)
```
caseydavenport: Why not use log here?

calico.go
```diff
@@ -284,6 +285,28 @@ func cmdDel(args *skel.CmdArgs) error {
 	fmt.Fprintf(os.Stderr, "Calico CNI releasing IP address\n")
 	logger.WithFields(log.Fields{"paths": os.Getenv("CNI_PATH"),
 		"type": conf.IPAM.Type}).Debug("Looking for IPAM plugin in paths")
+
+	// We need to replace "usePodCidr" with a valid, but dummy podCidr string with "host-local" IPAM.
+	if conf.IPAM.Type == "host-local" && strings.EqualFold(conf.IPAM.Subnet, "usePodCidr") {
+
+		// Use a dummy CIDR from IETF example IP range.
```
caseydavenport: We should give more detail as to why we need to do this and why it's OK.

caseydavenport: We probably shouldn't use the example range now that I think about it, given they are meant for examples.

Does "0.0.0.0/0" work?

caseydavenport:
I'm going to merge so we can get this in v2.0.2, but @gunjan5 can you please ensure we get the necessary test coverage for this?

Issue #258, opened 2017-01-31
Support specifying specific IP address for pods


gunjan5:
Continuing the conversation from #213:

As per @caseydavenport's comment:

-------------

After discussing with @tomdee, we're thinking about two annotations now to represent the two different use cases above.

```
cni.projectcalico.org/ipAddrs: "['10.0.0.1']"
```

This annotation will be passed to the configured IPAM plugin using the `IP=...` environment variable as described in the CNI conventions.  The CNI plugin will be expected to assign the provided address or return an error.

```
cni.projectcalico.org/ipAddrsNoIpam: "['10.0.0.1']"
```

If specified, this annotation ignores all other IPAM settings and will be stored in the Calico WorkloadEndpoint without calling any IPAM plugin, no questions asked.  This allows non-CNI things to be the source of truth for IPAM.

For now, these annotations will only support a single IP in each list, but will be formatted as lists for future-proofing since we know that CNI will support multiple IP addresses in the future.  When the time comes then multiple IPs will be supported.

-------------

caseydavenport:
A few extensions / thoughts:

- For now, only a single IP addresses is supported in the CNI result, so we should enforce that only a single IPv4 and/or a single IPv6 address is provided.

- If `cni.projectcalico.org/ipAddrs` is provided in conjunction with any IP Pool config from either the CNI config or an annotation, then we should assert that the provided addresses fall within the configured pools and error if they do not.

- If `cni.projectcalico.org/ipamOverrides` is provided then the IPAM plugin is not called, and `cni.projectcalico.org/ipAddrs` is ignored.

- These annotations are compatible with both IPv4 and IPv6 addresses. i.e a single annotation represents both.

- `cni.projectcalico.org/ipAddrs: "[ ]"` should return a "no IPs specified" error.
- `cni.projectcalico.org/ipamOverrides: "[ ]"` should return a "no IPs specified" error.

caseydavenport:
> The use of CNI_ARGS is deprecated and "args" should be used instead.

^ From the CNI conventions doc.

I think this means we should not be using the `IP=...` CNI_ARG currently supported in the Calico IPAM plugin, and instead should be looking to do something else.

CC @tomdee @gunjan5

tomdee:
@gunjan5 Can this be closed now?

gunjan5:
Fixed by #264 and #267

PR #259, opened 2017-02-02
[WIP] Support specifying IP with ipamOverride annotation

gunjan5:
Part of issue #258 

-------------------
```
cni.projectcalico.org/ipamOverrides: "['10.0.0.1']"
```

If specified, this annotation ignores all other IPAM settings and will be stored in the Calico WorkloadEndpoint without calling any IPAM plugin, no questions asked.  This allows non-CNI things to be the source of truth for IPAM.

--------------------

### DIFF
--- a/calico_cni_ipam_test.go
+++ b/calico_cni_ipam_test.go
@@ -83,6 +83,87 @@ var _ = Describe("Calico IPAM Tests", func() {
 		})
 	})
 
+	Describe("Run IPAM plugin - Verify IP Pools", func() {
+		Context("Pass valid pools", func() {
+			It("successfully networks the namespace", func() {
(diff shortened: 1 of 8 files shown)


cmluciano:
@gunjan5 this seems to fix the build failures and addresses the comments in the dependent PR. Is it fine to merge his PR, knowing that the tests will be fixed by this PR?

gunjan5:
@cmluciano I just had a chat with @tomdee about this. What we can do is I can create a new PR with @duglin's commits pulled in and then add my commits to fix the tests and add some more tests and apply review changes for the ipPools feature. I will keep duglin's original commits (squashed) in that PR.

gunjan5:
Moving this work to #264

PR #262, opened 2017-02-06
Fix up install script missing env var

caseydavenport:
#247 added a default CNI config to the install container.  However, the shell script which decided whether or not to override that didn't properly handle the CNI_NETWORK_CONFIG env var being unset. 

This PR allows for CNI_NETWORK_CONFIG to be unset, defaulting to the config in the container.


caseydavenport:
@tomdee added this to v2.1.0 since it's a bug in a new feature as of v2.1.0

PR #264, opened 2017-02-07
Support Pod /32 IP request via k8s annotation (ipAddrsNoIpam)

gunjan5:
Part of issue #258 

-------------------
```
cni.projectcalico.org/ipAddrsNoIpam: "[\"10.0.0.1\"]"
```

If specified, this annotation ignores all other IPAM settings and will be stored in the Calico WorkloadEndpoint without calling any IPAM plugin, no questions asked.  This allows non-CNI things to be the source of truth for IPAM.

--------------------

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -234,6 +234,81 @@ var _ = Describe("CalicoCni", func() {
 					Expect(ipPoolCIDR.Contains(podIP)).To(BeTrue())
 				})
 			})
+
+			Context("using ipAddrsNoIpam annotation to assign IP address to a pod, bypassing IPAM", func() {
+				It("should successfully assigns the annotated IP address", func() {
+					netconfCalicoIPAM := fmt.Sprintf(`
+				{
+			      "name": "net2",
+				  "type": "calico",
+				  "etcd_endpoints": "http://%s:2379",
+			 	  "ipam": {},
+					"kubernetes": {
+					  "k8s_api_root": "http://127.0.0.1:8080"
+					 },
+					"policy": {"type": "k8s"},
+					"log_level":"info"
+				}`, os.Getenv("ETCD_IP"))
+
+					assignIP := net.IPv4(10, 0, 0, 1).To4()
+
+					config, err := clientcmd.DefaultClientConfig.ClientConfig()
+					Expect(err).NotTo(HaveOccurred())
+
+					clientset, err := kubernetes.NewForConfig(config)
+					Expect(err).NotTo(HaveOccurred())
+
+					// Now create a K8s pod passing in an IP address.
+					name := fmt.Sprintf("run%d-ip", rand.Uint32())
+					pod, err := clientset.Pods(K8S_TEST_NS).Create(&v1.Pod{
+						ObjectMeta: v1.ObjectMeta{
+							Name: name,
+							Annotations: map[string]string{
+								"cni.projectcalico.org/ipAddrsNoIpam": "[\"10.0.0.1\"]",
+							},
+						},
+						Spec: v1.PodSpec{Containers: []v1.Container{{
+							Name:  fmt.Sprintf("container-%s", name),
+							Image: "ignore",
+						}}},
+					})
+					Expect(err).NotTo(HaveOccurred())
+
+					logger.Infof("Created POD object: %v", pod)
+
+					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name)
+					Expect(err).NotTo(HaveOccurred())
+					mac := contVeth.Attrs().HardwareAddr
+
+					podIP := contAddresses[0].IP
+					logger.Infof("All container IPs: %v", contAddresses)
+					logger.Infof("Container got IP address: %s", podIP)
+					Expect(podIP).Should(Equal(assignIP))
+
+					interfaceName := k8s.VethNameForWorkload(fmt.Sprintf("%s.%s", K8S_TEST_NS, name))
+
+					// The endpoint is created
+					endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(endpoints.Items).Should(HaveLen(1))
+					Expect(endpoints.Items[0].Metadata).Should(Equal(api.WorkloadEndpointMetadata{
+						Node:         hostname,
+						Name:         "eth0",
+						Workload:     fmt.Sprintf("test.%s", name),
+						Orchestrator: "k8s",
+						Labels:       map[string]string{"calico/k8s_ns": "test"},
+					}))
+					Expect(endpoints.Items[0].Spec).Should(Equal(api.WorkloadEndpointSpec{
+						InterfaceName: interfaceName,
+						IPNetworks: []cnet.IPNet{cnet.IPNet{net.IPNet{
+							IP:   assignIP,
+							Mask: net.CIDRMask(32, 32),
+						}}},
+						MAC:      &cnet.MAC{HardwareAddr: mac},
+						Profiles: []string{"k8s_ns.test"},
+					}))
+				})
+			})
 		})
 	})
 })
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -171,13 +171,25 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
 			}
 		}
 
-		// Run the IPAM plugin
-		logger.Debugf("Calling IPAM plugin %s", conf.IPAM.Type)
-		result, err = ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
-		if err != nil {
-			return nil, err
+		ipAddrsNoIpam := annot["cni.projectcalico.org/ipAddrsNoIpam"]
+
+		// Call IPAM plugin if ipAddrsNoIpam annotation is not present.
+		if ipAddrsNoIpam == "" {
+			// Run the IPAM plugin
+			logger.Debugf("Calling IPAM plugin %s", conf.IPAM.Type)
+			result, err = ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
+			if err != nil {
+				return nil, err
+			}
+			logger.Debugf("IPAM plugin returned: %+v", result)
+		} else {
+			// ipAddrsNoIpam annotation is set so bypass IPAM, and set the IPs manually.
+			result, err = overrideIPAMResult(ipAddrsNoIpam, logger)
+			if err != nil {
+				return nil, err
+			}
+			logger.Debugf("Bypassing IPAM to set the result to: %+v", result)
 		}
-		logger.Debugf("IPAM plugin returned: %+v", result)
 
 		// Create the endpoint object and configure it.
 		endpoint = api.NewWorkloadEndpoint()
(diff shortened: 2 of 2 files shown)


calico_cni_k8s_test.go
```diff
@@ -234,6 +234,55 @@ var _ = Describe("CalicoCni", func() {
...
+						}}},
+					})
+					Expect(err).NotTo(HaveOccurred())
+
+					logger.Infof("Created POD object: %v", pod)
+
+					_, _, _, _, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name)
+					Expect(err).NotTo(HaveOccurred())
+
+					podIP := contAddresses[0].IP
```
tomdee: What about checking what IP was stored in the workload endpoint in etcd?

k8s/k8s.go
```diff
@@ -238,6 +247,74 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+
+		// It's an IPv6 address if ip.To4 is nil.
+		if ipAddr.To4() == nil {
+			// We only allow one IPv4 and one IPv6 at the moment.
+			// So if we see more than one of IPv4 or IPv6 then we throw an error.
+			// If/when CNI spec supports more than one IP, we can loosen this requirement.
+			if visited6 >= 1 {
+				logger.Fatal("Can not have more than one IPv6 addresses in ipAddrsNoIpam annotation")
+			} else {
+				result.IP6.IP.IP = ipAddr
```
tomdee: Do you really need the `visited` stuff? Why not just checkout result.IPx.IP or whatever?

k8s/k8s.go
```diff
@@ -238,6 +247,74 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) *types.Result {
+	var ips []string
+	var visited4, visited6 int
+
+	err := json.Unmarshal([]byte(ipAddrsNoIpam), &ips)
+	if err != nil {
+		logger.WithField("annotation", ipAddrsNoIpam).Fatal("Invalid JSON")
+	}
+
+	result := types.Result{
```
tomdee: Do you really need all of this for creating the `Result`?

k8s/k8s.go
```diff
@@ -238,6 +247,74 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+	if len(ips) == 0 {
+		logger.WithField("annotation", "cni.projectcalico.org/ipAddrsNoIpam").Fatal("No IPs specified")
+	}
+
+	// Go through all the IPs passed in as annotation value and populate
+	// the result variable with IP4 and/or IP6 IPs.
+	// We also make sure there is only one IPv4 and/or one IPv6 passed in,
+	// since CNI spec only supports one of each right now.
+	for _, ip := range ips {
+
```
tomdee: Extra blank line

k8s/k8s.go
```diff
@@ -171,13 +171,22 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		if ipAddrsNoIpam == "" {
+			// Run the IPAM plugin
+			logger.Debugf("Calling IPAM plugin %s", conf.IPAM.Type)
+			result, err = ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
+			if err != nil {
+				return nil, err
+			}
+			logger.Debugf("IPAM plugin returned: %+v", result)
+		} else {
+			// ipAddrsNoIpam annotation is set so bypass IPMA, and set the IPs manually.
```
caseydavenport: typo: IPMA

k8s/k8s.go
```diff
@@ -171,13 +171,22 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			result = overrideIPAMResult(ipAddrsNoIpam, logger)
+			logger.Debugf("Bypassing IPAM to set the IP config to: %+v", result)
```
caseydavenport: IP config -> result to be consistent.

k8s/k8s.go
```diff
@@ -238,6 +247,74 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+// but sets IP field manually since IPAM is bypassed with this annotation.
+// Example annotation:
+// cni.projectcalico.org/ipAddrsNoIpam: "[\"10.0.0.1\", \"2001:db8::1\"]"
+func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) *types.Result {
+	var ips []string
+	var visited4, visited6 int
+
+	err := json.Unmarshal([]byte(ipAddrsNoIpam), &ips)
+	if err != nil {
+		logger.WithField("annotation", ipAddrsNoIpam).Fatal("Invalid JSON")
```
caseydavenport: Could you make the error here more descriptive?  It's user facing, so something like "failed to parse ipAddrsNoIpam as json"

utils/types.go
```diff
@@ -46,13 +46,14 @@ type NetConf struct {
...
-		IPv4Pools  []string `json:"ipv4_pools,omitempty"`
-		IPv6Pools  []string `json:"ipv6_pools,omitempty"`
+		Name          string
+		Type          string   `json:"type"`
+		Subnet        string   `json:"subnet"`
+		AssignIpv4    *string  `json:"assign_ipv4"`
+		AssignIpv6    *string  `json:"assign_ipv6"`
+		IPv4Pools     []string `json:"ipv4_pools,omitempty"`
+		IPv6Pools     []string `json:"ipv6_pools,omitempty"`
+		IPAddrsNoIpam []string `json:"ip_addrs_noipam,omitempty"`
```
caseydavenport: I don't think this needs to be part of the CNI config file.

k8s/k8s.go
```diff
@@ -238,6 +247,74 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
 	return result, nil
 }
 
+// overrideIPAMResult generates types.Result like the one produced by IPAM plugin,
+// but sets IP field manually since IPAM is bypassed with this annotation.
+// Example annotation:
+// cni.projectcalico.org/ipAddrsNoIpam: "[\"10.0.0.1\", \"2001:db8::1\"]"
```
caseydavenport: The whole annotation isn't passed to this function, right?  We should document the format of the string that gets passed to this function.

k8s/k8s.go
```diff
@@ -171,13 +171,22 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		// Call IPAM plugin if ipAddrsNoIpam annotation is not present.
+		if ipAddrsNoIpam == "" {
+			// Run the IPAM plugin
+			logger.Debugf("Calling IPAM plugin %s", conf.IPAM.Type)
+			result, err = ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
+			if err != nil {
+				return nil, err
+			}
+			logger.Debugf("IPAM plugin returned: %+v", result)
+		} else {
```
caseydavenport: Are we worried about the fact that we do IPAM related stuff up above even when this annotation is specified?

e.g https://github.com/projectcalico/cni-plugin/pull/264/files#diff-b396ca928d5bd0e161acd283b56ead9fR84

and https://github.com/projectcalico/cni-plugin/pull/264/files#diff-b396ca928d5bd0e161acd283b56ead9fR124?

Probably not because it's a kinda weird case, but it does mean we'll be doing more work than we need to.

k8s/k8s.go
gunjan5: Not sure I understand what you mean here. Can you elaborate?

gunjan5: I think I get what you're saying now. Done!

k8s/k8s.go
gunjan5: (discussed this offline. Going to leave this as it is for now. Will need to refactor how we call k8s APIs at some point.

gunjan5:
@tomdee @caseydavenport addressed your comments. PTAL

k8s/k8s.go
```diff
@@ -238,6 +247,67 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
 
+// overrideIPAMResult generates types.Result like the one produced by IPAM plugin,
+// but sets IP field manually since IPAM is bypassed with this annotation.
+// Example annotation value string: "[\"10.0.0.1\", \"2001:db8::1\"]"
+func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) *types.Result {
+	var ips []string
+
+	err := json.Unmarshal([]byte(ipAddrsNoIpam), &ips)
+	if err != nil {
+		logger.WithField("annotation", ipAddrsNoIpam).Fatal("Failed to parse ipAddrsNoIpam as json")
```
caseydavenport: Should we be doing a log.Fatal here? Shouldn't this function return an error so we can pass back an error over CNI?

caseydavenport: Same below.

k8s/k8s.go
```diff
@@ -238,6 +247,67 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+	}
+
+	result := types.Result{}
+
+	// annotation value can't be empty.
+	if len(ips) == 0 {
+		logger.WithField("annotation", "cni.projectcalico.org/ipAddrsNoIpam").Fatal("No IPs specified")
```
caseydavenport: I always forget... should this actually be `ipAddrsNoIPAM`?

gunjan5: I think that's what you and @lxpollitt decided, but then later you had a suggestion of using `ipAddrsSkipIPAM` which I personally like better

k8s/k8s.go
```diff
@@ -238,6 +251,72 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+// overrideIPAMResult generates types.Result like the one produced by IPAM plugin,
+// but sets IP field manually since IPAM is bypassed with this annotation.
+// Example annotation value string: "[\"10.0.0.1\", \"2001:db8::1\"]"
+func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) (*types.Result, error) {
+	var ips []string
+
+	err := json.Unmarshal([]byte(ipAddrsNoIpam), &ips)
+	if err != nil {
+		logger.WithField("annotation", ipAddrsNoIpam).Error("Failed to parse ipAddrsNoIpam as json")
+		return nil, err
```
caseydavenport: Would be nice to wrap these errors with context before returning. 

e.g `return nil, fmt.Errorf("Failed to parse ipAddrsNoIpam: %s", err)`

gunjan5: done! BTW, this is why I think we should be using https://github.com/pkg/errors 🙂

k8s/k8s.go
```diff
@@ -238,6 +250,72 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+
+	// Go through all the IPs passed in as annotation value and populate
+	// the result variable with IP4 and/or IP6 IPs.
+	// We also make sure there is only one IPv4 and/or one IPv6 passed in,
+	// since CNI spec only supports one of each right now.
+	for _, ip := range ips {
+		ipAddr := net.ParseIP(ip)
+		if ipAddr == nil {
+			logger.WithField("IP", ip).Error("Invalid IP format")
+			return nil, fmt.Errorf("Invalid IP format: %s", err)
```
caseydavenport: It's this line that should include the IP in the returned error so that it's clear what the misconfig is.

k8s/k8s.go
```diff
@@ -238,6 +250,72 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		}
+
+		// It's an IPv6 address if ip.To4 is nil.
+		if ipAddr.To4() == nil {
+			// We only allow one IPv4 and one IPv6 at the moment.
+			// So if we see more than one of IPv4 or IPv6 then we throw an error.
+			// If/when CNI spec supports more than one IP, we can loosen this requirement.
+			if result.IP6 != nil {
+				logger.Error("Can not have more than one IPv6 addresses in ipAddrsNoIpam annotation")
+				return nil, fmt.Errorf("Can not have more than one IPv6 addresses in ipAddrsNoIpam annotation")
```
caseydavenport: typo: _more than one IPv6 addresses_

k8s/k8s.go
```diff
@@ -238,6 +250,72 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+					IP:   ipAddr,
+					Mask: net.CIDRMask(128, 128),
+				},
+			}
+			logger.Debugf("Adding IPv6: %s to result", ipAddr.String())
+		} else {
+			// It's an IPv4 address.
+			if result.IP4 != nil {
+				logger.Error("Can not have more than one IPv4 addresses in ipAddrsNoIpam annotation")
+				return nil, fmt.Errorf("Can not have more than one IPv4 addresses in ipAddrsNoIpam annotation")
```
caseydavenport: more than one IPv4 addresses (shouldn't be plural)

PR #267, opened 2017-02-09
Support ipAddrs annotation to assign a specific IP to a pod through IPAM

gunjan5:
Second annotation for https://github.com/projectcalico/cni-plugin/issues/258

Example annotation:
```yaml
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: my-nginx-mixed
spec:
  replicas: 1
  template:
    metadata:
      labels:
        run: my-nginx-mixed
      annotations:
        "cni.projectcalico.org/ipAddrs": "[\"192.168.0.65\", \"2001:db8::65\"]"
    spec:
      containers:
      - name: my-nginx-mixed
        image: nginx
        ports:
        - containerPort: 80
```

```bash
# ip a
...
4: eth0@if28: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
    link/ether c6:3b:c3:8e:a3:17 brd ff:ff:ff:ff:ff:ff
    inet 192.168.0.65/32 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 2001:db8::65/128 scope global
       valid_lft forever preferred_lft forever
    inet6 fe80::c43b:c3ff:fe8e:a317/64 scope link
       valid_lft forever preferred_lft forever
```

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -239,7 +239,7 @@ var _ = Describe("CalicoCni", func() {
 				It("should successfully assigns the annotated IP address", func() {
 					netconfCalicoIPAM := fmt.Sprintf(`
 				{
-			      "name": "net2",
+			      "name": "net3",
 				  "type": "calico",
 				  "etcd_endpoints": "http://%s:2379",
 			 	  "ipam": {},
@@ -288,6 +288,93 @@ var _ = Describe("CalicoCni", func() {
 					interfaceName := k8s.VethNameForWorkload(fmt.Sprintf("%s.%s", K8S_TEST_NS, name))
 
 					// The endpoint is created
+					endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(endpoints.Items).Should(HaveLen(1))
+					Expect(endpoints.Items[0].Metadata).Should(Equal(api.WorkloadEndpointMetadata{
+						Node:         hostname,
+						Name:         "eth0",
+						Workload:     fmt.Sprintf("test.%s", name),
+						Orchestrator: "k8s",
+						Labels:       map[string]string{"calico/k8s_ns": "test"},
+					}))
+					Expect(endpoints.Items[0].Spec).Should(Equal(api.WorkloadEndpointSpec{
+						InterfaceName: interfaceName,
+						IPNetworks: []cnet.IPNet{cnet.IPNet{net.IPNet{
+							IP:   assignIP,
+							Mask: net.CIDRMask(32, 32),
+						}}},
+						MAC:      &cnet.MAC{HardwareAddr: mac},
+						Profiles: []string{"k8s_ns.test"},
+					}))
+				})
+			})
+
+			Context("using ipAddrs annotation to assign IP address to a pod, through Calico IPAM", func() {
+
+				It("should successfully assigns the annotated IP address", func() {
+					netconfCalicoIPAM := fmt.Sprintf(`
+				{
+			      "name": "net4",
+				  "type": "calico",
+				  "etcd_endpoints": "http://%s:2379",
+				  "ipam": {
+					   "type": "calico-ipam",
+					   "assign_ipv4": "true",
+					   "assign_ipv6": "true"
+				   },
+					"kubernetes": {
+					  "k8s_api_root": "http://127.0.0.1:8080"
+					 },
+					"policy": {"type": "k8s"},
+					"log_level":"info"
+				}`, os.Getenv("ETCD_IP"))
+
+					assignIP := net.IPv4(20, 0, 0, 111).To4()
+
+					// Create a new ipPool.
+					ipPool := "20.0.0.0/24"
+					c, _ := testutils.NewClient("")
+					testutils.CreateNewIPPool(*c, ipPool, false, false, true)
+					_, _, err := net.ParseCIDR(ipPool)
+					Expect(err).NotTo(HaveOccurred())
+
+					config, err := clientcmd.DefaultClientConfig.ClientConfig()
+					Expect(err).NotTo(HaveOccurred())
+
+					clientset, err := kubernetes.NewForConfig(config)
+					Expect(err).NotTo(HaveOccurred())
+
+					// Now create a K8s pod passing in an IP address.
+					name := fmt.Sprintf("run%d-ip", rand.Uint32())
+					pod, err := clientset.Pods(K8S_TEST_NS).Create(&v1.Pod{
+						ObjectMeta: v1.ObjectMeta{
+							Name: name,
+							Annotations: map[string]string{
+								"cni.projectcalico.org/ipAddrs": "[\"20.0.0.111\"]",
+							},
+						},
+						Spec: v1.PodSpec{Containers: []v1.Container{{
+							Name:  fmt.Sprintf("container-%s", name),
+							Image: "ignore",
+						}}},
+					})
+					Expect(err).NotTo(HaveOccurred())
+
+					logger.Infof("Created POD object: %v", pod)
+
+					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name)
+					Expect(err).NotTo(HaveOccurred())
+					mac := contVeth.Attrs().HardwareAddr
+
+					podIP := contAddresses[0].IP
+					logger.Infof("All container IPs: %v", contAddresses)
+					logger.Infof("Container got IP address: %s", podIP)
+					Expect(podIP).Should(Equal(assignIP))
+
+					interfaceName := k8s.VethNameForWorkload(fmt.Sprintf("%s.%s", K8S_TEST_NS, name))
+
+					// Make sure WorkloadEndpoint is created and has the requested IP in the datastore.
 					endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
 					Expect(err).ShouldNot(HaveOccurred())
 					Expect(endpoints.Items).Should(HaveLen(1))
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -172,23 +172,38 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
 		}
 
 		ipAddrsNoIpam := annot["cni.projectcalico.org/ipAddrsNoIpam"]
+		ipAddrs := annot["cni.projectcalico.org/ipAddrs"]
 
-		// Call IPAM plugin if ipAddrsNoIpam annotation is not present.
-		if ipAddrsNoIpam == "" {
-			// Run the IPAM plugin
+		// switch based on which annotations are passed or not passed.
+		switch {
+		case ipAddrs == "" && ipAddrsNoIpam == "":
+			// Call IPAM plugin if ipAddrsNoIpam or ipAddrs annotation is not present.
 			logger.Debugf("Calling IPAM plugin %s", conf.IPAM.Type)
 			result, err = ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
 			if err != nil {
 				return nil, err
 			}
 			logger.Debugf("IPAM plugin returned: %+v", result)
-		} else {
+		case ipAddrs != "" && ipAddrsNoIpam != "":
+			// Can't have both ipAddrs and ipAddrsNoIpam annotations at the same time.
+			e := fmt.Errorf("Can't have both annotations: 'ipAddrs' and 'ipAddrsNoIpam' in use at the same time")
+			logger.Error(e)
+			return nil, e
+		case ipAddrsNoIpam != "":
 			// ipAddrsNoIpam annotation is set so bypass IPAM, and set the IPs manually.
 			result, err = overrideIPAMResult(ipAddrsNoIpam, logger)
 			if err != nil {
 				return nil, err
 			}
 			logger.Debugf("Bypassing IPAM to set the result to: %+v", result)
+		case ipAddrs != "":
+			// When ipAddrs annotation is set, we call out to the configured IPAM plugin
+			// requesting the specific IP addresses included in the annotation.
+			result, err = ipAddrsResult(ipAddrs, conf, args, logger)
+			if err != nil {
+				return nil, err
+			}
+			logger.Debugf("IPAM result set to: %+v", result)
 		}
 
 		// Create the endpoint object and configure it.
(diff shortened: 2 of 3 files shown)


calico_cni_k8s_test.go
```diff
@@ -309,6 +309,93 @@ var _ = Describe("CalicoCni", func() {
 					}))
 				})
 			})
+
+			Context("using ipAddrs annotation to assign IP address to a pod, through Calico IPAM", func() {
```
tomdee: I would favour a little more testing here. Could this be made into a table driven test that tests for example a single IPv4 address, a single IPv6 address, mixed IPv4 and IPv6 and error cases like too many addresses?

gunjan5: I tried adding those but our (dind?) test framework seems to have some issues with the netlink library and IPv6 when it tries to `DoNetworking` for IPv6 address

k8s/k8s.go
```diff
@@ -172,23 +172,36 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
 		ipAddrsNoIpam := annot["cni.projectcalico.org/ipAddrsNoIpam"]
-
-		// Call IPAM plugin if ipAddrsNoIpam annotation is not present.
-		if ipAddrsNoIpam == "" {
-			// Run the IPAM plugin
+		ipAddrs := annot["cni.projectcalico.org/ipAddrs"]
+
+		// switch based on which annotations are passed or not passed.
+		switch {
+		// Can't have both ipAddrs and ipAddrsNoIpam annotations at the same time.
```
caseydavenport: Be consistent with the placement of your comments in the switch statement.  My preference is for the comments to go within each case.

k8s/k8s.go
```diff
@@ -172,23 +172,36 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		case ipAddrs != "" && ipAddrsNoIpam != "":
+			logger.Error("Can't have both annotations: 'ipAddrs' and 'ipAddrsNoIpam' in use at the same time")
+			return nil, fmt.Errorf("Can't have both annotations: 'ipAddrs' and 'ipAddrsNoIpam' in use at the same time")
+		case ipAddrs == "" && ipAddrsNoIpam == "":
```
caseydavenport: Personally I'd like to see this case first, give it's the "mainline".

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		logger.WithField("annotation", ipAddrs).Error("Failed to parse ipAddrs as json")
+		return nil, fmt.Errorf("Failed to parse ipAddrs: %s", err)
+	}
+
+	result := types.Result{}
+
+	// annotation value can't be empty.
+	if len(ips) == 0 {
+		logger.WithField("annotation", "cni.projectcalico.org/ipAddrs").Error("No IPs specified")
+		return nil, fmt.Errorf("No IPs specified with annotation \"cni.projectcalico.org/ipAddrs\"")
```
caseydavenport: This would be more clear if it was worded like `"Annotation \"cni.projectcalico.org/ipAddrs\" specified but empty"`

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+
+	// Go through all the IPs passed in as annotation value and call IPAM plugin
+	// for each, and populate the result variable with IP4 and/or IP6 IPs returned
+	// from the IPAM plugin. We also make sure there is only one IPv4 and/or one IPv6
+	// passed in, since CNI spec only supports one of each right now.
+	for _, ip := range ips {
+		ipAddr := net.ParseIP(ip)
+		if ipAddr == nil {
+			logger.WithField("IP", ip).Error("Invalid IP format")
+			return nil, fmt.Errorf("Invalid IP format: %s", err)
```
caseydavenport: Does this error log out the invalid IP?

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			// We only allow one IPv4 and one IPv6 at the moment.
+			// So if we see more than one of IPv4 or IPv6 then we throw an error.
+			// If/when CNI spec supports more than one IP, we can loosen this requirement.
+			if result.IP6 != nil {
+				logger.Error("Can not have more than one IPv6 addresses in ipAddrsNoIpam annotation")
+				return nil, fmt.Errorf("Can not have more than one IPv6 addresses in ipAddrsNoIpam annotation")
+			}
+
+			// Calll setIPArg to set CNI_ARGS and call the IPAM plugin with it, and then unset it.
+			r, err := setIPArg(ipAddr, conf, args, logger)
```
caseydavenport: `setIPArg` is a misleading name for this function, since it implies it's just setting an environment variable, when in fact it's actually doing a lot more than that.  

Something like `callIPAMWithIP` or similar would be more accurate and make the code easier to read.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			return nil, fmt.Errorf("Invalid IP format: %s", err)
+		}
+
+		// It's an IPv6 address if ip.To4 is nil.
+		if ipAddr.To4() == nil {
+			// We only allow one IPv4 and one IPv6 at the moment.
+			// So if we see more than one of IPv4 or IPv6 then we throw an error.
+			// If/when CNI spec supports more than one IP, we can loosen this requirement.
+			if result.IP6 != nil {
+				logger.Error("Can not have more than one IPv6 addresses in ipAddrsNoIpam annotation")
```
caseydavenport: IPv6 addresses should be singular

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+				return nil, fmt.Errorf("Error getting IP from IPAM: %s", err)
+			}
+
+			// Set the IP6 part of the result from the type.Result returned by the IPAM plugin.
+			result.IP6 = r.IP6
+			logger.Debugf("Adding IPv6: %s to result", ipAddr.String())
+		} else {
+			// It's an IPv4 address.
+			if result.IP4 != nil {
+				logger.Error("Can not have more than one IPv4 addresses in ipAddrsNoIpam annotation")
```
caseydavenport: IP4 addresses should be singular.

caseydavenport: Also, this isn't the `ipAddrsNoIpam` annotation I don't think :)

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+				return nil, fmt.Errorf("Can not have more than one IPv4 addresses in ipAddrsNoIpam annotation")
+			}
+
+			// Calll setIPArg to set CNI_ARGS and call the IPAM plugin with it, and then unset it.
```
caseydavenport: Three `l`s

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			// Set the IP4 part of the result from the type.Result returned by the IPAM plugin.
+			result.IP4 = r.IP4
+			logger.Debugf("Adding IPv4: %s to result", ipAddr.String())
+		}
+	}
+
+	return &result, nil
+}
+
+// setIPArg sets CNI_ARGS with the IP requested from the IPAM plugin, then it calls
```
caseydavenport: nit: the IP is not requested from the IPAM plugin. Rather, it sets CNI_ARGS with the requested IP and then calls the IPAM plugin.

k8s/k8s.go
gunjan5: I thought I changed it. oops

k8s/k8s.go
gunjan5: hmm I fixed these things, plurals, errors etc, but I guess missed the update in rebasing

calico_cni_k8s_test.go
gunjan5: Discussed this offline. Seems like an issue with our testing framework where it calls netlink only with IPv4 address. Not worth spending time on it for now, but would be good to fix that for future tests.

gunjan5:
Updated the PR based on your review. PTAL

k8s/k8s.go
```diff
@@ -172,23 +172,36 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			return nil, fmt.Errorf("Can't have both annotations: 'ipAddrs' and 'ipAddrsNoIpam' in use at the same time")
+		case ipAddrsNoIpam != "":
 			// ipAddrsNoIpam annotation is set so bypass IPAM, and set the IPs manually.
 			result, err = overrideIPAMResult(ipAddrsNoIpam, logger)
 			if err != nil {
 				return nil, err
 			}
 			logger.Debugf("Bypassing IPAM to set the result to: %+v", result)
+		case ipAddrs != "":
+			// ipAddrs annotation is set.
```
caseydavenport: This is a pretty useless comment :)

I think a bit more detail here would be really useful.  e.g "when this annotation is set, we call out to the configured IPAM plugin requesting the specific IP addresses included in the annotation".

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
 	return result, nil
 }
 
+// ipAddrsResult parses the ipAddrs annotation and calls IPAM plugin for
```
caseydavenport: calls _the configured_ IPAM plugin.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+// each IP passed to it by setting CNI_ARGS IP=ip_addr, and returns the type.Result.
```
caseydavenport: returns the _result of calling the IPAM plugin_

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+// Example annotation value string: "[\"10.0.0.1\", \"2001:db8::1\"]"
+func ipAddrsResult(ipAddrs string, conf utils.NetConf, args *skel.CmdArgs, logger *log.Entry) (*types.Result, error) {
+	var ips []string
+
+	err := json.Unmarshal([]byte(ipAddrs), &ips)
```
caseydavenport: This code to parse a string into a list of IP addresses is also in the overrides function below.

It would be nice if both `ipAddrsResult` and `overrideIPAMResult` took `[]net.IP` as an arg rather than a string.  We could make the parsing code common in a utility function.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+
+	// Set CNI_ARGS to the new value.
+	err := os.Setenv("CNI_ARGS", newArgs)
+	if err != nil {
+		return nil, fmt.Errorf("Error setting CNI_ARGS environment variable: %v", err)
+	}
+
+	// Run the IPAM plugin.
+	logger.Debugf("Calling IPAM plugin %s", conf.IPAM.Type)
+	r, err := ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
```
caseydavenport: If we hit an error here, we don't unset the CNI_ARGS.

Might be best to use a `defer` so that we are guaranteed to reset the value.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+// so it doesn't pollute the subsequent requests.
+func callIPAMWithIP(ip net.IP, conf utils.NetConf, args *skel.CmdArgs, logger *log.Entry) (*types.Result, error) {
+
+	// Save the original value of the CNI_ARGS ENV var for backup.
+	originalArgs := os.Getenv("CNI_ARGS")
+	logger.Debugf("Original CNI_ARGS=%s", originalArgs)
+
+	// Append IP=ip_address to CNI_ARGS.
+	// See: https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#cni_args for more info.
+	newArgs := originalArgs + ";IP=" + ip.String()
```
caseydavenport: We might want to check that `IP=` isn't already set in the args.

I'm not sure _how_ that could actually happen, but given it's a documented convention we should probably handle that case.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+// callIPAMWithIP sets CNI_ARGS with the IP and calls the IPAM plugin with it
+// to get types.Result and then it unsets the IP field from CNI_ARGS ENV var,
+// so it doesn't pollute the subsequent requests.
+func callIPAMWithIP(ip net.IP, conf utils.NetConf, args *skel.CmdArgs, logger *log.Entry) (*types.Result, error) {
+
+	// Save the original value of the CNI_ARGS ENV var for backup.
+	originalArgs := os.Getenv("CNI_ARGS")
+	logger.Debugf("Original CNI_ARGS=%s", originalArgs)
+
+	// Append IP=ip_address to CNI_ARGS.
```
caseydavenport: What is `ip_address`?

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+	r, err := ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
+	if err != nil {
+		return nil, err
+	}
+	logger.Debugf("IPAM plugin returned: %+v", r)
+
+	// Restore the CNI_ARGS ENV var to it's original value,
+	// so the subsequent calls don't get polluted by the old IP value.
+	err = os.Setenv("CNI_ARGS", originalArgs)
+	if err != nil {
```
caseydavenport: If we hit an error here, we need to clean up the IP allocation we just made.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+		if ipAddr.To4() == nil {
+			// We only allow one IPv4 and one IPv6 at the moment.
+			// So if we see more than one of IPv4 or IPv6 then we throw an error.
+			// If/when CNI spec supports more than one IP, we can loosen this requirement.
+			if result.IP6 != nil {
+				logger.Error("Can not have more than one IPv6 address in ipAddrs annotation")
+				return nil, fmt.Errorf("Can not have more than one IPv6 address in ipAddrs annotation")
+			}
+
+			// Calll callIPAMWithIP to set CNI_ARGS and call the IPAM plugin with it, and then unset it.
```
caseydavenport: `Calll` should only have two `l`s

caseydavenport: This comment doesn't need to talk about the implementation of `callIPAMWithIP`.

i.e we don't want to have to change this comment when we decide to start using the `args` field instead of `CNI_ARGS`.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+	for _, ip := range ips {
+		ipAddr := net.ParseIP(ip)
+		if ipAddr == nil {
+			logger.WithField("IP", ip).Error("Invalid IP format")
+			return nil, fmt.Errorf("Invalid IP format: %s", ip)
+		}
+
+		// It's an IPv6 address if ip.To4 is nil.
+		if ipAddr.To4() == nil {
+			// We only allow one IPv4 and one IPv6 at the moment.
```
caseydavenport: It's not us that only allows one IPv4 / IPv6, it's the CNI spec.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			result.IP6 = r.IP6
+			logger.Debugf("Adding IPv6: %s to result", ipAddr.String())
+		} else {
+			// It's an IPv4 address.
+			if result.IP4 != nil {
+				logger.Error("Can not have more than one IPv4 address in ipAddrs annotation")
+				return nil, fmt.Errorf("Can not have more than one IPv4 address in ipAddrs annotation")
+			}
+
+			// Call callIPAMWithIP to set CNI_ARGS and call the IPAM plugin with it, and then unset it.
```
caseydavenport: Same comment re: leaking implementation for this comment.

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+func ipAddrsResult(ipAddrs string, conf utils.NetConf, args *skel.CmdArgs, logger *log.Entry) (*types.Result, error) {
+	var ips []string
+
+	err := json.Unmarshal([]byte(ipAddrs), &ips)
+	if err != nil {
+		logger.WithField("annotation", ipAddrs).Error("Failed to parse ipAddrs as json")
+		return nil, fmt.Errorf("Failed to parse ipAddrs: %s", err)
+	}
+
+	result := types.Result{}
```
caseydavenport: Could you move this closer to where it's used?  Maybe line 291?

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+
+	// Go through all the IPs passed in as annotation value and call IPAM plugin
+	// for each, and populate the result variable with IP4 and/or IP6 IPs returned
+	// from the IPAM plugin. We also make sure there is only one IPv4 and/or one IPv6
+	// passed in, since CNI spec only supports one of each right now.
+	for _, ip := range ips {
+		ipAddr := net.ParseIP(ip)
+		if ipAddr == nil {
+			logger.WithField("IP", ip).Error("Invalid IP format")
+			return nil, fmt.Errorf("Invalid IP format: %s", ip)
```
caseydavenport: This should _also_ include the error that was hit.

e.g

```
fmt.Errorf("Invalid IP %s: %s", ip, err)
```

k8s/k8s.go
gunjan5: i meant `<insert-IP-address-here>`. I will change it to `<IP_ADDR>`

k8s/k8s.go
gunjan5: That is what this comment says, in line 3: `If/when CNI spec supports more than one IP, we can loosen this requirement.`

k8s/k8s.go
gunjan5: `net.ParseIP` doesn't return an error. It just sets the result to `nil` if it's not valid

k8s/k8s.go
```diff
@@ -250,6 +263,118 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+
+	err := json.Unmarshal([]byte(ipAddrs), &ips)
+	if err != nil {
+		logger.WithField("annotation", ipAddrs).Error("Failed to parse ipAddrs as json")
+		return nil, fmt.Errorf("Failed to parse ipAddrs: %s", err)
+	}
+
+	result := types.Result{}
+
+	// annotation value can't be empty.
```
gunjan5: we decided to throw an error when it's empty. See comment: https://github.com/projectcalico/cni-plugin/issues/258#issuecomment-276505762

> - `cni.projectcalico.org/ipAddrs: "[ ]"` should return a "no IPs specified" error.
> - `cni.projectcalico.org/ipAddrsNoIpam: "[ ]"` should return a "no IPs specified" error.

k8s/k8s.go
caseydavenport: My point was more that it's not referencing anything in the code.  <IP_ADDR> isn't any better.

The comment should say something like "Request the provided IP address using the IP CNI_ARG"

k8s/k8s.go
tomdee: OK, fair enough

k8s/k8s.go
caseydavenport: My comment was more about this line:

> We only allow one

k8s/k8s.go
```diff
@@ -316,6 +445,29 @@ func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) (*types.Result,
 	return &result, nil
 }
 
+// parseIPAddrs is a utility function that parses string of IPs in json format that are
+// passed in as a string and returns a slice of string with IPs.
+// It also makes sure the slice isn't empty.
+func parseIPAddrs(annotation, ipAddrsStr string, logger *log.Entry) ([]string, error) {
```
caseydavenport: Rather than pass the annotation in here just for logging purposes, I'd much rather do the context logging outside of this function.

gunjan5: hmm I don't know if I agree, there are a few places we log info/failure inside the function, and having the annotation there would help a lot. 
Like: `logger.WithField("annotation", annotation).Errorf("Failed to parse '%s' as JSON", annotation)` knowing which annotation this was helps a lot. 

Maybe im misunderstanding what you're saying

k8s/k8s.go
```diff
@@ -250,26 +264,141 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+			res = nil
+			finalErr = fmt.Errorf("Error setting CNI_ARGS environment variable: %v", err)
+		}
+	}()
+
+	// Run the IPAM plugin.
+	logger.Debugf("Calling IPAM plugin %s", conf.IPAM.Type)
+	r, err := ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
+	if err != nil {
+		return nil, err
```
caseydavenport: Not a huge fan of the mixed explicit returns / `finalErr` stuff.  I think it makes this function harder to follow.

gunjan5: I had to do named return because defere'd closure func in Go gets ignored, so only way to pass it to the parent function's return is to use named returns. See: https://groups.google.com/forum/#!msg/golang-nuts/qTTBENO_Em0/Y4MCVZZ3c5sJ (second message)

k8s/k8s.go
```diff
@@ -250,26 +264,141 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, hostname string, calicoCl
...
+	defer func() {
+		err := os.Setenv("CNI_ARGS", originalArgs)
+		if err != nil {
+			// Need to clean up IP allocation if this step doesn't succeed.
+			utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
+
+			// Returned values from a defered closure func get ignored in Go, so have to use named return.
+			// Following two lines are equivalent of return nil, fmt.Errorf("Error setting CNI_ARGS environment variable: %v", err)
+			res = nil
+			finalErr = fmt.Errorf("Error setting CNI_ARGS environment variable: %v", err)
```
caseydavenport: See comment below about not liking the mixed return stuff.

I think we should just not care about the finalErr / res stuff.

k8s/k8s.go
caseydavenport: Yep, I understand why, but let's just not return anything in the deferred function and then we don't need the named returns.

k8s/k8s.go
caseydavenport: If we log the annotation before calling this then it's moot, and makes this code much better encapsulated than it is now.

The logs would look something like this:

```
INFO Parsing annotation 'cni.projectcalico.org/ipAddrs': [..., ...]
ERROR Failed to parse [..., ...] as json: <error msg here>
```

We don't need to log out the annotation in every log - the context comes from the logs surrounding it.

caseydavenport: The point I'm trying to make is one about breaking the code into logical components.  As it is, the lines between this code and the rest are quite blurred because the code in this function is based on our knowledge that the calling code is always going to be from an annotation. 

What about if we want to call this code based on something from the CNI network config, which isn't an annotation?  What value would we pass for `annotation`?

gunjan5:
@caseydavenport I've removed the `defer` handling and consolidated error strings as discussed. PTAL

k8s/k8s.go
```diff
@@ -316,6 +451,27 @@ func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) (*types.Result,
...
+// It also makes sure the slice isn't empty.
+func parseIPAddrs(ipAddrsStr string, logger *log.Entry) ([]string, error) {
+	var ips []string
+
+	err := json.Unmarshal([]byte(ipAddrsStr), &ips)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse '%s' as JSON: %s", ipAddrsStr, err)
+	}
+
+	// annotation value can't be empty.
```
caseydavenport: Minor thing, but I'd recommend moving this `len()` check out of this function.  There's nothing about this helper that says it needs to be parsing an annotation, and it's really the calling code that knows if `len() == 0` is an error or not.

caseydavenport:
LGTM.

Squash and merge away!

PR #270, opened 2017-02-14
add tests for host-local IPAM IP leak bugfix

### LINKED ISSUE #257: Tests for hostlocal CNI IP leak bugfix
Tests for #255

gunjan5:
Fixes #257

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -89,7 +89,7 @@ var _ = Describe("CalicoCni", func() {
 				if err != nil {
 					panic(err)
 				}
-				containerID, netnspath, session, contVeth, contAddresses, contRoutes, err := CreateContainer(netconf, name)
+				containerID, netnspath, session, contVeth, contAddresses, contRoutes, err := CreateContainer(netconf, name, "")
 
 				Expect(err).ShouldNot(HaveOccurred())
 				Eventually(session).Should(gexec.Exit())
@@ -225,7 +225,7 @@ var _ = Describe("CalicoCni", func() {
 
 					logger.Infof("Created POD object: %v", pod)
 
-					_, _, _, _, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name)
+					_, _, _, _, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
 					Expect(err).NotTo(HaveOccurred())
 
 					podIP := contAddresses[0].IP
@@ -276,7 +276,7 @@ var _ = Describe("CalicoCni", func() {
 
 					logger.Infof("Created POD object: %v", pod)
 
-					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name)
+					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
 					Expect(err).NotTo(HaveOccurred())
 					mac := contVeth.Attrs().HardwareAddr
 
@@ -363,7 +363,7 @@ var _ = Describe("CalicoCni", func() {
 
 					logger.Infof("Created POD object: %v", pod)
 
-					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name)
+					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
 					Expect(err).NotTo(HaveOccurred())
 					mac := contVeth.Attrs().HardwareAddr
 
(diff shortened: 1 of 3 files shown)


calico_cni_k8s_test.go
```diff
@@ -396,6 +396,95 @@ var _ = Describe("CalicoCni", func() {
...
+						Spec: v1.NodeSpec{
+							PodCIDR: "10.0.0.0/24",
+						},
+					})
+					Expect(err).NotTo(HaveOccurred())
+
+					// Step 1: Create the first pod on node we created.
+					name := fmt.Sprintf("run%d", rand.Uint32())
+
+					// Create a K8s pod with the node we created above.
```
caseydavenport: For clarity, I'd delete this comment and the empty line above it so that the "// Step 1" comment is right above the block it describes.

calico_cni_k8s_test.go
```diff
@@ -396,6 +396,95 @@ var _ = Describe("CalicoCni", func() {
...
+
+					// Step 2: Delete the pod we created earlier, so it should release the IP address.
+					session, err = DeleteContainer(netconfHostLocalIPAM, netnspath, name)
+					Expect(err).ShouldNot(HaveOccurred())
+					Eventually(session).Should(gexec.Exit())
+
+					// Step 3: Create a second pod on the same node, it should get the same IP first pod got.
+					name2 := fmt.Sprintf("run2%d", rand.Uint32())
+
+					// Create another K8s pod with the node we created above.
```
caseydavenport: Same here w.r.t the comments.  This is just a duplicate of the "Step 3:" comment.

calico_cni_k8s_test.go
```diff
@@ -396,6 +396,95 @@ var _ = Describe("CalicoCni", func() {
...
+						Spec: v1.PodSpec{Containers: []v1.Container{{
+							Name:  fmt.Sprintf("container-%s", name),
+							Image: "ignore",
+						}},
+							NodeName: hostname,
+						},
+					})
+					Expect(err).NotTo(HaveOccurred())
+
+					// host-local IPAM always assigns the IPs sequentialls except the first 2 that are reserved
```
caseydavenport: typo: "sequentialls"

caseydavenport: I'm not a huge fan of relying on this internal understanding of how the host-local IPAM plugin assigns IP addresses.

This could easily change in the future and would break this test.  Can we find another way of asserting that the plugin released the IP address?

gunjan5:
`test-containerized` passes on my mac, but it's failing on semaphore for some reason. I'm investigating

gunjan5:
@caseydavenport applied your feedback. Made some changes to the framework to accept IP so I can request a specific IP. Tests seem to pass now

calico_cni_k8s_test.go
```diff
@@ -396,6 +396,91 @@ var _ = Describe("CalicoCni", func() {
...
+					// Create a K8s Node object with PodCIDR and name equal to hostname.
+					_, err = clientset.Nodes().Create(&v1.Node{
+						ObjectMeta: v1.ObjectMeta{Name: hostname},
+						Spec: v1.NodeSpec{
+							PodCIDR: "10.0.0.0/24",
+						},
+					})
+					Expect(err).NotTo(HaveOccurred())
+
+					// Step 1: Create the first pod on node we created, and request a specific IP.
```
caseydavenport: Does it make sense to put each of these steps in a `By(` so we have added context when a test fails?

calico_cni_k8s_test.go
```diff
@@ -396,6 +396,91 @@ var _ = Describe("CalicoCni", func() {
...
+
+					requestedIP := "10.0.0.42"
+					expectedIP := net.IPv4(10, 0, 0, 42).To4()
+
+					_, netnspath, session, _, contAddresses, _, err := CreateContainer(netconfHostLocalIPAM, name, requestedIP)
+					Expect(err).NotTo(HaveOccurred())
+
+					podIP := contAddresses[0].IP
+					logger.Infof("All container IPs: %v", contAddresses)
+					logger.Infof("Container got IP address: %s", podIP)
```
caseydavenport: I think the second log here is redundant.

caseydavenport:
@gunjan5 two more minor comments, then LGTM.

calico_cni_k8s_test.go
gunjan5: That's what I was thinking of initially, but this comment doesn't really fit the `By()` BDD style, but I guess it makes sense to have it so it's clear when the tests fail

caseydavenport: `By(creating a pod with a specific IP address)`

test_utils/utils.go
```diff
@@ -142,7 +142,7 @@ func CreateContainer(netconf string, k8sName string) (container_id, netnspath st
 	//TODO pass in the env properly
 	var k8s_env = ""
 	if k8sName != "" {
-		k8s_env = fmt.Sprintf("CNI_ARGS=\"K8S_POD_NAME=%s;K8S_POD_NAMESPACE=test;K8S_POD_INFRA_CONTAINER_ID=whatever\"", k8sName)
+		k8s_env = fmt.Sprintf("CNI_ARGS=\"K8S_POD_NAME=%s;K8S_POD_NAMESPACE=test;K8S_POD_INFRA_CONTAINER_ID=whatever;IP=%s\"", k8sName, ip)
```
caseydavenport: Rather than always append this, could you have a separate IF statement which adds it if `ip` is passed?

Otherwise we're changing all the other tests to get passed `IP=`

caseydavenport:
LGTM

PR #279, opened 2017-02-27
Insert serviceaccount token in the generated kubeconfig.

caseydavenport:
Haven't tested this yet, but intention is to start moving all the API connection details out of the network config and into the kubeconfig (given that specifying them in the CNI network is deprecated as per our docs).

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -106,6 +106,8 @@ clusters:
     insecure-skip-tls-verify: true
 users:
 - name: calico 
+  user:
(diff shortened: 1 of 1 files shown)


tomdee:
@caseydavenport Do you need to find someone to test this? Is there a corresponding docs PR to go with it too?

caseydavenport:
@tomdee yeah, I need to get some testing in on this.  I'd also like to move `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` in there as well.

There will be some corresponding docs changes.  I don't have a PR yet.

Issue #226, opened 2016-12-05
Switch from tags -> labels


caseydavenport:
The CNI plugin should be writing profiles that use labels rather than tags, since tags are deprecated and will be removed in the future.

PR #285, opened 2017-03-14
Use labels instead of tags + tests

gunjan5:
`Tags` are deprecated, so in order to phase `Tags` out, we want to replace tags with `Labels` in all of our plugins where default `Profile` is created by the plugins (libnetwork-plugin and cni-plugin)


tomdee:
@gunjan5 Can you update the PR description to describe why we're making this change? At the moment the PR is a little cryptic.

tomdee:
See #226

PR #289, opened 2017-03-16
Don't error if things don't exist

### LINKED ISSUE #287: Don't error on teardown for pod that never received a setup
Seen using k8s `v1.6.0-alpha.3`.

The DNS pod gets stuck in ContainerCreating because we're receiving a `TearDown` for the sandbox container which never received a `Setup`, because the container was started before CNI was installed.

We should be doing a best-effort teardown rather than failing in this case.

CC @tomdee

djosborne:
Fixes #287 

This PR enables the CNI plugin to gracefully "Delete" a container that it it never "Setup". It does this by not erroring out if the resources it was called against do not exist, be it resources in the etcd datastore, or devices in the namespace it was given.

This resolves a condition that has popped up in K8s 1.6 where Kubernetes seems to lock a container in "ContainerCreating" status when Calico errors a CNI teardown request. This request was errored because Calico was never given a Create request.

This PR should also help scenarios in the future where Calico may have only half-completed a SetUp request or was called to teardown  a dirty environment.

WIP. Please don't review yet, will clean up tomorrow a.m.

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -325,24 +325,39 @@ func cmdDel(args *skel.CmdArgs) error {
 		return err
 	}
 
-	if err := calicoClient.WorkloadEndpoints().Delete(api.WorkloadEndpointMetadata{
+	ep := api.WorkloadEndpointMetadata{
 		Name:         args.IfName,
 		Node:         nodename,
 		Orchestrator: orchestrator,
-		Workload:     workload}); err != nil {
-		return err
+		Workload:     workload}
+	if err = calicoClient.WorkloadEndpoints().Delete(ep); err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
+			logger.WithField("endpoint", ep).Info("Endpoint object does not exist, no need to clean up.")
+		} else {
+			return err
+		}
 	}
 
 	// Only try to delete the device if a namespace was passed in.
 	if args.Netns != "" {
-		fmt.Fprintf(os.Stderr, "Calico CNI deleting device in netns %s\n", args.Netns)
-		err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-			_, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+		logger.Debug("Checking namespace & device exist.")
+		devErr := ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+			_, err := netlink.LinkByName(args.IfName)
 			return err
 		})
 
-		if err != nil {
-			return err
+		if devErr == nil {
+			fmt.Fprintf(os.Stderr, "Calico CNI deleting device in netns %s\n", args.Netns)
+			err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+				_, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+				return err
+			})
+
+			if err != nil {
+				return err
+			}
+		} else {
+			logger.Info("veth does not exist, no need to clean up.")
 		}
 	}
 
--- a/calico_cni_test.go
+++ b/calico_cni_test.go
@@ -128,9 +128,8 @@ var _ = Describe("CalicoCni", func() {
 						Type:      syscall.RTN_UNICAST,
 					})))
 
-				session, err = DeleteContainer(netconf, netnspath, "")
+				_, err = DeleteContainer(netconf, netnspath, "")
 				Expect(err).ShouldNot(HaveOccurred())
-				Eventually(session).Should(gexec.Exit())
 
 				// Make sure there are no endpoints anymore
 				endpoints, err = calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
@@ -190,9 +189,8 @@ var _ = Describe("CalicoCni", func() {
 					Orchestrator: "cni",
 				}))
 
-				session, err = DeleteContainer(netconf, netnspath, "")
+				_, err = DeleteContainer(netconf, netnspath, "")
 				Expect(err).ShouldNot(HaveOccurred())
-				Eventually(session).Should(gexec.Exit())
 			})
 		})
 	})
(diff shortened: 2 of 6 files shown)


calico.go
```diff
@@ -319,24 +319,39 @@ func cmdDel(args *skel.CmdArgs) error {
...
 		return err
 	}
 
 	// Only try to delete the device if a namespace was passed in.
 	if args.Netns != "" {
-		fmt.Fprintf(os.Stderr, "Calico CNI deleting device in netns %s\n", args.Netns)
-		err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-			_, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+		// Ensure the namespace & device exist.
+		nilWhenDevExists := ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
```
djosborne: @tomdee We have to manually check if the interface exists here because CNI library doesn't return different error types depending on the error encountered during [`DelLinkByNameAddr`](https://github.com/containernetworking/cni/blob/master/pkg/ip/link.go#L157-L173). If that function did a `ErrDeviceNotFound`, `ErrNamespaceNotFound`, `ErrCouldntDeleteDevice`, we could gracefully decide if we needed to error or not.

calico.go
```diff
@@ -319,24 +319,40 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		}
 	}
 
 	// Only try to delete the device if a namespace was passed in.
 	if args.Netns != "" {
-		fmt.Fprintf(os.Stderr, "Calico CNI deleting device in netns %s\n", args.Netns)
-		err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-			_, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+		logger.Info("Checking namespace & device exist.")
+		nilWhenDevExists := ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
```
tomdee: I think something like devErr would be a better name

test_utils/utils.go
```diff
@@ -116,6 +116,27 @@ func RunIPAMPlugin(netconf, command, args string) (types.Result, types.Error, in
 	return result, error, exitCode
 }
 
+func CreateContainerNamespace() (containerId, netnspath string, err error) {
```
tomdee: Why haven't you made CreateContainer() use this function?

test_utils/utils.go
```diff
@@ -187,6 +208,41 @@ func CreateContainer(netconf string, k8sName string, ip string) (container_id, n
 	return
 }
 
+// Executes the Calico CNI plugin and return the error code of the command.
+func CmdDel(netconf, netnspath, name string) (exitCode int, err error) {
```
tomdee: Do both CmdDel and DeleteContainer need to exist? There's a lot of common code between them, should one call the other? Should other tests be updated to use CmdDel?

calico_cni_test.go
```diff
@@ -154,4 +154,76 @@ var _ = Describe("CalicoCni", func() {
...
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(exitCode).To(Equal(0))
+				})
+
+				Context("with an interface", func() {
+					It("deletes the interface and succesfully exits", func() {
+						_, netnspath, err := CreateContainerNamespace()
+						Expect(err).ShouldNot(HaveOccurred())
+
+						// TODO: create an interface
```
tomdee: TODO

calico_cni_test.go
```diff
@@ -154,4 +154,76 @@ var _ = Describe("CalicoCni", func() {
...
+
+						exitCode, err := CmdDel(netconf, netnspath, "")
+						Expect(err).ShouldNot(HaveOccurred())
+						Expect(exitCode).To(Equal(0))
+
+						// TODO: assert interface was deleted
```
tomdee: TODO

calico.go
```diff
@@ -319,24 +319,40 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	ep := api.WorkloadEndpointMetadata{
 		Name:         args.IfName,
 		Node:         hostname,
 		Orchestrator: orchestrator,
-		Workload:     workload}); err != nil {
-		return err
+		Workload:     workload}
+	if err = calicoClient.WorkloadEndpoints().Delete(ep); err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
+			logger.WithField("endpoint", ep).Warn("Asked to delete endpoint object, but it doesn't exist. Ignoring.")
```
caseydavenport: If we know this is mainline operation, is a Warning really the right log level?

calico.go
```diff
@@ -319,24 +319,40 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			return err
+		}
 	}
 
 	// Only try to delete the device if a namespace was passed in.
 	if args.Netns != "" {
-		fmt.Fprintf(os.Stderr, "Calico CNI deleting device in netns %s\n", args.Netns)
-		err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-			_, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+		logger.Info("Checking namespace & device exist.")
```
caseydavenport: Could we stick this at Debug level?

calico.go
```diff
@@ -319,24 +319,40 @@ func cmdDel(args *skel.CmdArgs) error {
...
+				_, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+				return err
+			})
+
+			if err != nil {
+				return err
+			}
+		} else {
+			// This commonly occurs if Calico was issued a teardown but never received the SetUp signal.
+			logger.Warn("Asked to clean up container namespace, but container device (or namespace) does not exist. Ignoring.")
```
caseydavenport: I think these shouldn't be Warning logs - probably Info is right so they're not as scary but still visible.

caseydavenport: something more benign sounding would be nice too: "veth does not exist, no need to clean up"

test_utils/utils.go
```diff
@@ -203,14 +209,21 @@ func DeleteContainer(netconf, netnspath, name string) (session *gexec.Session, e
...
 	io.WriteString(stdin, "\n")
 	stdin.Close()
 
-	session, err = gexec.Start(subProcess, ginkgo.GinkgoWriter, ginkgo.GinkgoWriter)
+	session, err := gexec.Start(subProcess, ginkgo.GinkgoWriter, ginkgo.GinkgoWriter)
+	if err != nil {
+		return
+	}
+
+	// Wait should supposedly induce a test failure if it times out?
```
djosborne: Yea that was a tad lazy of me. I added a 6 second timeout and can confirm the test prints a failure in an obvious manner. Will update the comment.

PR #297, opened 2017-03-22
Read Datastore config from file specified in CNI config

tmjd:
Part of work for https://github.com/projectcalico/calico/issues/563

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -180,7 +180,7 @@ func CreateClient(conf NetConf) (*client.Client, error) {
 	log.Infof("Configured environment: %+v", os.Environ())
 
 	// Load the client config from the current environment.
-	clientConfig, err := client.LoadClientConfig("")
+	clientConfig, err := client.LoadClientConfig(conf.DatastoreFile)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


utils/utils.go
```diff
@@ -180,7 +180,7 @@ func CreateClient(conf NetConf) (*client.Client, error) {
 	log.Infof("Configured environment: %+v", os.Environ())
 
 	// Load the client config from the current environment.
-	clientConfig, err := client.LoadClientConfig("")
+	clientConfig, err := client.LoadClientConfig(conf.DatastoreFile)
```
djosborne: Would be nice to log out the values _after_ client config has been read in. That way users can see what values were picked up.

tmjd: I think that is a good idea too.  I already added that to the LoadClientConfig function in my changes that are in a libcalico-go PR.  I just double checked and it wasn't in the other exposed functions loading config so I added it to them too (LoadClientConfigFromBytes and LoadClientConfigFromEnvironment).

caseydavenport:
I'm going to close this for now - we can certainly re-open if/when we pick this back up.

PR #298, opened 2017-03-22
Bundle newer CNI binaries

djosborne:
Tested with a kubernetes 1.6 cluster kicked up with kubeadm.

Confirmed that updated flannel now handles teardowns correctly for pods it didn't setup.


djosborne:
@tomdee another quick look before I hit it?

PR #303, opened 2017-03-30
On install, only modify CNI config if cfg is specified

### LINKED ISSUE #299: Kubernetes install script assumes calico.conf.tmp always exists
When the `CNI_NETWORK_CONFIG` environment variable is not set, the `calico.conf.tmp` file is not created. However, this file is used down the line.

The following are the logs for an execution of the `install-cni` container:
```
k logs calico-node-6368r install-cni
CNI plugin version: v1.6.1
sed: calico.conf.tmp: No such file or directory
sed: calico.conf.tmp: No such file or directory
sed: calico.conf.tmp: No such file or directory
sed: calico.conf.tmp: No such file or directory
sed: calico.conf.tmp: No such file or directory
sed: calico.conf.tmp: No such file or directory
Wrote CNI config: {
    "name": "k8s-pod-network",
    "type": "calico",
    "etcd_endpoints": "https://ip-10-0-3-176:6666",
    "etcd_cert_file": "/etc/kubernetes/kubenode.pem",
    "etcd_key_file": "/etc/kubernetes/kubenode-key.pem",
    "etcd_ca_cert_file": "/etc/kubernetes/ca.pem",
    "log_level": "info",
    "ipam": {
        "type": "calico-ipam"
    },
    "kubernetes": {
        "kubeconfig": "/var/lib/kubelet/kubeconfig"
    }
}
sed: calico.conf.tmp: No such file or directory
sed: calico.conf.tmp: No such file or directory
sed: calico.conf.tmp: No such file or directory
sed: calico.conf.tmp: No such file or directory
mv: can't rename 'calico.conf.tmp': No such file or directory
```

tmjd:
Fixes #299

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -5,17 +5,17 @@
 # - Expects the host CNI network config path to be mounted at /host/etc/cni/net.d.
 # - Expects the desired CNI config in the CNI_NETWORK_CONFIG env variable.
 
-# Ensure all variables are defined. 
-set -u 
+# Ensure all variables are defined.
+set -u
 
-# The directory on the host where CNI networks are installed. Defaults to 
+# The directory on the host where CNI networks are installed. Defaults to
 # /etc/cni/net.d, but can be overridden by setting CNI_NET_DIR.  This is used
 # for populating absolute paths in the CNI network config to assets
 # which are installed in the CNI network config directory.
 HOST_CNI_NET_DIR=${CNI_NET_DIR:-/etc/cni/net.d}
 HOST_SECRETS_DIR=${HOST_CNI_NET_DIR}/calico-tls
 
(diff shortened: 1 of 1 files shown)


caseydavenport:
Hm, I don't think this is quite right. There is a default `calico.conf.tmp` that still needs the sed replacements.

https://github.com/projectcalico/cni-plugin/blob/master/k8s-install/scripts/calico.conf.default

tmjd:
Is that file named incorrectly then?  Should it be `calico.conf.tmp` instead of `calico.conf.default`?
Because otherwise I don't see how that default file is ever used.

tmjd:
I don't think this is the correct fix anymore. It looks like the script is meant to read the default if the Env Var is not set but where the script is ran and where the default file is do not match up.  I'll put the fix up in a different PR.

PR #309, opened 2017-04-06
Change tmp cfg file path to be absolute

### LINKED ISSUE #299: Kubernetes install script assumes calico.conf.tmp always exists

tmjd:
Fixes #299

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -5,17 +5,17 @@
 # - Expects the host CNI network config path to be mounted at /host/etc/cni/net.d.
 # - Expects the desired CNI config in the CNI_NETWORK_CONFIG env variable.
 
-# Ensure all variables are defined. 
-set -u 
+# Ensure all variables are defined.
+set -u
 
-# The directory on the host where CNI networks are installed. Defaults to 
+# The directory on the host where CNI networks are installed. Defaults to
 # /etc/cni/net.d, but can be overridden by setting CNI_NET_DIR.  This is used
 # for populating absolute paths in the CNI network config to assets
 # which are installed in the CNI network config directory.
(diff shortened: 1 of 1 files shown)


tmjd:
I tested this by removing the cni_network_config and the env var using it in a calico manifest. Ran it with the current cni image checked the logs to see the failure, then updated the manifest to use an image I built and ran that and checked the logs again and saw success.

Issue #310, opened 2017-04-10
Include container ID in WorkloadEndpoint Key


caseydavenport:
We need to encode the container ID into the identifying information of the WorkloadEndpoint to fix https://github.com/projectcalico/calicoctl/issues/1592.

#### Design, as discussed with @tomdee:

- Add a new field called `ContainerID` to the WorkloadEndointMetadata struct in libcalico.

For example: 

```go
// ContainerID, when populated, contains an identifier for the container networked for 
// this WorkloadEndpoint. ContainerID is an optional field.
ContainerID string `json:containerID,omitempty" value:"omitempty,name"`
```

This field will need to be supported in the etcd datastore driver, but does not need to be populated by the Kubernetes datastore driver, since the KDD doesn't perform CRUD on WorkloadEndpoints. 

- The CNI ADD call will be modified to populate the new ContainerID field on the WorkloadEndpoints it creates.

- The CNI DEL call will be modified to respect the ContainerID field in the following ways.
  - If ContainerID != "" and ContainerID does not match CNI_CONTAINERID, return without deleting.
  - If ContainerID != "" and ContainerID matches CNI_CONTAINERID, then execute the delete.
  - If ContainerID == "", then execute the delete.

#### Testing

#1592 primarily affects the etcd datastore driver, and so testing should primarily be done against the etcd datastore, with simple smoke testing against the KDD.

We should add a unit test which does:
- CNI ADD with container ID X
- CNI DEL with container ID X
- CNI ADD with container ID Y
- CNI DEL with **_container  ID X_**

The end result of the test will be that a WorkloadEndpoint exists with container ID Y.

#### Documentation

Likely just a release note, with no user documentation updates required. 

#### Releases

This should go into v2.2, and be ported back to v2.1 as well.

fasaxc:
Would it work to consider `pod A with container ID X` and `pod A with container ID Y` as different workloads.  I.e just encode the pod and container ID into the Workload field?

The datamodel was meant to be orthogonal to whether we're networking a container or a VM, for example, so the intention was the orchestrator would put whatever it needed in the Workload and Endpoint fields.

tomdee:
Changing the workload "ID" is tricky for upgrade/downgrade so we judged it not to be the best design approach

immesys:
Is there a hotfix?

fasaxc:
@tomdee but where are you going to store this in etcd?  Where does it go in the backend key/data?

[...]

PR #312, opened 2017-04-18
Check ContainerID before Deleting a pod

gunjan5:
Fixes https://github.com/projectcalico/cni-plugin/issues/310

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -20,16 +20,11 @@ import (
 	"fmt"
 	"os"
 	"runtime"
-	"strings"
-
-	"github.com/vishvananda/netlink"
 
 	"net"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/containernetworking/cni/pkg/ip"
 	"github.com/containernetworking/cni/pkg/ipam"
-	"github.com/containernetworking/cni/pkg/ns"
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
 	"github.com/containernetworking/cni/pkg/types/current"
@@ -306,78 +301,40 @@ func cmdDel(args *skel.CmdArgs) error {
 		"Node":         nodename,
 	}).Info("Extracted identifiers")
 
-	// Always try to release the address. Don't deal with any errors till the endpoints are cleaned up.
-	fmt.Fprintf(os.Stderr, "Calico CNI releasing IP address\n")
-	logger.WithFields(log.Fields{"paths": os.Getenv("CNI_PATH"),
-		"type": conf.IPAM.Type}).Debug("Looking for IPAM plugin in paths")
-
-	// We need to replace "usePodCidr" with a valid, but dummy podCidr string with "host-local" IPAM.
-	if conf.IPAM.Type == "host-local" && strings.EqualFold(conf.IPAM.Subnet, "usePodCidr") {
-
-		// host-local IPAM releases the IP by ContainerID, so podCidr isn't really used to release the IP.
-		// It just needs a valid CIDR, but it doesn't have to be the CIDR associated with the host.
-		dummyPodCidr := "0.0.0.0/0"
-		var stdinData map[string]interface{}
-
-		if err := json.Unmarshal(args.StdinData, &stdinData); err != nil {
-			return err
-		}
-
-		logger.WithField("podCidr", dummyPodCidr).Info("Using a dummy podCidr to release the IP")
-		stdinData["ipam"].(map[string]interface{})["subnet"] = dummyPodCidr
-
-		args.StdinData, err = json.Marshal(stdinData)
-		if err != nil {
-			return err
-		}
-		logger.WithField("stdin", args.StdinData).Debug("Updated stdin data for Delete Cmd")
-	}
-
-	ipamErr := ipam.ExecDel(conf.IPAM.Type, args.StdinData)
-
-	if ipamErr != nil {
-		logger.Error(ipamErr)
-	}
-
 	calicoClient, err := CreateClient(conf)
 	if err != nil {
 		return err
 	}
 
-	ep := api.WorkloadEndpointMetadata{
+	wep := api.WorkloadEndpointMetadata{
 		Name:         args.IfName,
 		Node:         nodename,
 		Orchestrator: orchestrator,
-		Workload:     workload}
-	if err = calicoClient.WorkloadEndpoints().Delete(ep); err != nil {
+		Workload:     workload,
+	}
+
+	// Handle k8s specific bits of handling the DEL.
+	if orchestrator == "k8s" {
+		return k8s.CmdDelK8s(calicoClient, wep, args, conf, logger)
+	}
+
+	// Release the IP address by calling the configured IPAM plugin.
+	ipamErr := CleanUpIPAM(conf, args, logger)
+
+	// Delete the WorkloadEndpoint object from the datastore.
+	if err = calicoClient.WorkloadEndpoints().Delete(wep); err != nil {
 		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
-			logger.WithField("endpoint", ep).Info("Endpoint object does not exist, no need to clean up.")
+			// Log and proceed with the clean up if WEP doesn't exist.
+			logger.WithField("endpoint", wep).Info("Endpoint object does not exist, no need to clean up.")
 		} else {
 			return err
 		}
 	}
 
-	// Only try to delete the device if a namespace was passed in.
-	if args.Netns != "" {
-		logger.Debug("Checking namespace & device exist.")
-		devErr := ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-			_, err := netlink.LinkByName(args.IfName)
-			return err
-		})
-
-		if devErr == nil {
-			fmt.Fprintf(os.Stderr, "Calico CNI deleting device in netns %s\n", args.Netns)
-			err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-				_, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
-				return err
-			})
-
-			if err != nil {
-				return err
-			}
-		} else {
-			logger.Info("veth does not exist, no need to clean up.")
-		}
+	// Clean up namespace by removing the interfaces.
+	err = CleanUpNamespace(args, logger)
+	if err != nil {
+		return err
 	}
 
 	// Return the IPAM error if there was one. The IPAM error will be lost if there was also an error in cleaning up
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -114,12 +114,16 @@ var _ = Describe("CalicoCni", func() {
 				endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
 				Expect(err).ShouldNot(HaveOccurred())
 				Expect(endpoints.Items).Should(HaveLen(1))
+
+				// Set the Revision to nil since we can't assert it's exact value.
+				endpoints.Items[0].Metadata.Revision = nil
 				Expect(endpoints.Items[0].Metadata).Should(Equal(api.WorkloadEndpointMetadata{
-					Node:         hostname,
-					Name:         "eth0",
-					Workload:     fmt.Sprintf("test.%s", name),
-					Orchestrator: "k8s",
-					Labels:       map[string]string{"calico/k8s_ns": "test"},
+					Node:             hostname,
+					Name:             "eth0",
+					Workload:         fmt.Sprintf("test.%s", name),
+					ActiveInstanceID: containerID,
+					Orchestrator:     "k8s",
+					Labels:           map[string]string{"calico/k8s_ns": "test"},
 				}))
 				Expect(endpoints.Items[0].Spec).Should(Equal(api.WorkloadEndpointSpec{
 					InterfaceName: interfaceName,
@@ -312,7 +316,7 @@ var _ = Describe("CalicoCni", func() {
 
 					logger.Infof("Created POD object: %v", pod)
 
-					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
+					containerID, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
 					Expect(err).NotTo(HaveOccurred())
 					mac := contVeth.Attrs().HardwareAddr
 
@@ -327,12 +331,16 @@ var _ = Describe("CalicoCni", func() {
 					endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
 					Expect(err).ShouldNot(HaveOccurred())
 					Expect(endpoints.Items).Should(HaveLen(1))
+
+					// Set the Revision to nil since we can't assert it's exact value.
+					endpoints.Items[0].Metadata.Revision = nil
 					Expect(endpoints.Items[0].Metadata).Should(Equal(api.WorkloadEndpointMetadata{
-						Node:         hostname,
-						Name:         "eth0",
-						Workload:     fmt.Sprintf("test.%s", name),
-						Orchestrator: "k8s",
-						Labels:       map[string]string{"calico/k8s_ns": "test"},
+						Node:             hostname,
+						Name:             "eth0",
+						Workload:         fmt.Sprintf("test.%s", name),
+						ActiveInstanceID: containerID,
+						Orchestrator:     "k8s",
+						Labels:           map[string]string{"calico/k8s_ns": "test"},
 					}))
 					Expect(endpoints.Items[0].Spec).Should(Equal(api.WorkloadEndpointSpec{
 						InterfaceName: interfaceName,
@@ -400,7 +408,7 @@ var _ = Describe("CalicoCni", func() {
 
 					logger.Infof("Created POD object: %v", pod)
 
-					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
+					containerID, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
 					Expect(err).NotTo(HaveOccurred())
 					mac := contVeth.Attrs().HardwareAddr
 
@@ -415,12 +423,16 @@ var _ = Describe("CalicoCni", func() {
 					endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
 					Expect(err).ShouldNot(HaveOccurred())
 					Expect(endpoints.Items).Should(HaveLen(1))
+
+					// Set the Revision to nil since we can't assert it's exact value.
+					endpoints.Items[0].Metadata.Revision = nil
 					Expect(endpoints.Items[0].Metadata).Should(Equal(api.WorkloadEndpointMetadata{
-						Node:         hostname,
-						Name:         "eth0",
-						Workload:     fmt.Sprintf("test.%s", name),
-						Orchestrator: "k8s",
-						Labels:       map[string]string{"calico/k8s_ns": "test"},
+						Node:             hostname,
+						Name:             "eth0",
+						Workload:         fmt.Sprintf("test.%s", name),
+						ActiveInstanceID: containerID,
+						Orchestrator:     "k8s",
+						Labels:           map[string]string{"calico/k8s_ns": "test"},
 					}))
 					Expect(endpoints.Items[0].Spec).Should(Equal(api.WorkloadEndpointSpec{
 						InterfaceName: interfaceName,
@@ -486,7 +498,7 @@ var _ = Describe("CalicoCni", func() {
 
 					logger.Infof("Created POD object: %v", pod)
 
-					_, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
+					containerID, _, _, contVeth, contAddresses, _, err := CreateContainer(netconfCalicoIPAM, name, "")
 					Expect(err).NotTo(HaveOccurred())
 					mac := contVeth.Attrs().HardwareAddr
 
@@ -501,12 +513,16 @@ var _ = Describe("CalicoCni", func() {
 					endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
 					Expect(err).ShouldNot(HaveOccurred())
 					Expect(endpoints.Items).Should(HaveLen(1))
+
+					// Set the Revision to nil since we can't assert it's exact value.
+					endpoints.Items[0].Metadata.Revision = nil
 					Expect(endpoints.Items[0].Metadata).Should(Equal(api.WorkloadEndpointMetadata{
-						Node:         hostname,
-						Name:         "eth0",
-						Workload:     fmt.Sprintf("test.%s", name),
-						Orchestrator: "k8s",
-						Labels:       map[string]string{"calico/k8s_ns": "test"},
+						Node:             hostname,
+						Name:             "eth0",
+						Workload:         fmt.Sprintf("test.%s", name),
+						ActiveInstanceID: containerID,
+						Orchestrator:     "k8s",
+						Labels:           map[string]string{"calico/k8s_ns": "test"},
 					}))
 					Expect(endpoints.Items[0].Spec).Should(Equal(api.WorkloadEndpointSpec{
 						InterfaceName: interfaceName,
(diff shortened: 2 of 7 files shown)


gunjan5:
not sure why Semaphore can't pull my libcalicogo fork, all tests pass locally (containerized)
```bash
Ran 28 of 28 Specs in 15.778 seconds
SUCCESS! -- 28 Passed | 0 Failed | 0 Pending | 0 Skipped PASS
```

calico.go
```diff
@@ -178,6 +178,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 			endpoint = api.NewWorkloadEndpoint()
 			endpoint.Metadata.Name = args.IfName
 			endpoint.Metadata.Node = nodename
+			endpoint.Metadata.ActiveInstanceID = args.ContainerID
```
tomdee: What's your reasoning for including this line?

calico.go
```diff
@@ -306,6 +307,33 @@ func cmdDel(args *skel.CmdArgs) error {
...
+
+	ep := api.WorkloadEndpointMetadata{
+		Name:             args.IfName,
+		Node:             nodename,
+		ActiveInstanceID: args.ContainerID,
+		Orchestrator:     orchestrator,
+		Workload:         workload,
+	}
+
+	wep, err := calicoClient.WorkloadEndpoints().Get(ep)
```
tomdee: Does this `Get()` require the ActiveInstanceID to be present? What happens if it's missing (e.g. on upgrade?)

tomdee: I was expecting more a of compare-and-delete approach here.

You at least need some comments explaining the complex flow we have going on here...

calico_cni_k8s_test.go
```diff
@@ -602,6 +607,137 @@ var _ = Describe("CalicoCni", func() {
 					Expect(pod2IP).Should(Equal(expectedIP))
 				})
 			})
+
+			Context("ADD a continaer with a ContainerID and DEL it with the same ContainerID then ADD a new container with a different ContainerID, and send a DEL for the old ContainerID", func() {
```
tomdee: typo

tomdee: Add a comment with a link to the specific issues so we know why this test was added

calico_cni_test.go
```diff
@@ -52,16 +51,16 @@ var _ = Describe("CalicoCni", func() {
...
-			  "cniVersion": "%s",
-			  "name": "net1",
-			  "type": "calico",
-			  "etcd_endpoints": "http://%s:2379",
-			  "ipam": {
-			    "type": "host-local",
-			    "subnet": "10.0.0.0/8"
-			  }
-			}`, cniVersion, os.Getenv("ETCD_IP"))
+            {
```
tomdee: Why change the indent here?

calico_cni_test.go
```diff
@@ -297,4 +299,155 @@ var _ = Describe("CalicoCni", func() {
 			})
 		})
 	})
+
+	Describe("ADD a continaer with ContainerID and DEL it with the same ContainerID", func() {
```
tomdee: typo

calico_cni_test.go
```diff
@@ -297,4 +299,155 @@ var _ = Describe("CalicoCni", func() {
...
+
+				// The endpoint should not exist in the backend datastore.
+				endpoints, err = calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(endpoints.Items).Should(HaveLen(0))
+			})
+		})
+	})
+
+	Describe("ADD a continaer with a ContainerID and DEL it with the same ContainerID then ADD a new container with a different ContainerID, and send a DEL for the old ContainerID", func() {
```
tomdee: typo

tomdee: What value do you see in having a k8s and non-k8s version of this test?

calico.go
robbrockbank: @tomdee @gunjan5 

Whose responsibility is it to police the ActiveInstanceID.  If its CNI then I'm worried there is a mighty big timing window here:

-  Old instance queries Endpoint gets ID - looks OK
-  New instance applies updated Endpoint data (updates the ActiveInstanceID)
-  Old instance goes - yup looks fine, delete.

At the moment, the consumer facing API does not include revision info so there is currently no CompareAndSwap.  That was something that I'd added in my libcalico-go refactor.  So there is no way for CNI to protect against this kind of operation.

Unless we can guarantee the delete is *always* called after a successful create/update then I worry this window will be quite large.

If you moved the ActiveInstanceID checking into libcalico-go then you do have the ability to do the CaS operation.

robbrockbank: maybe nightly big is an exaggeration - but it might lead for some subtle hard to debug issues if the window is possible.

calico.go
gunjan5: we populate that field in `k8s.CmdAddK8s`, we want the same behavior for non-k8s orchestrators, this is for that.

calico.go
gunjan5: ActiveInstanceID is part of the Metadata, but not the key, so other 4 fields that make up the key are still unchanged

calico_cni_test.go
gunjan5: must have been my editor

calico_cni_test.go
gunjan5: right now it's just k8s being too "aggressive" in sending DELs, we could have other non-k8s doing the same and since both of these take different code paths (for ADD) I wanted to make sure we get the same behavior. In fact, I found an issue with this test, as the result I had L181 in calico.go :)

calico.go
tomdee: Why do we want the same behaviour?

gunjan5: The reason why I am thinking this is what if some other orchestrator decides to be "aggressive" with DELs in future, we should be populating `ActiveInstanceID` similarly across all orchestrators so it's not confusing when someone looks at the code. Another reason is we go through different code paths for ADD (for k8s and non-k8s), but same code path for DEL and DEL code checks for `ActiveInstanceID` to avoid double-delete situation, which is unique to k8s for now, but could arise in future for any other orchestrators we support

calico.go
robbrockbank: ActiveID is not required on the Get since it isn't an index.

calico.go
```diff
@@ -339,16 +367,6 @@ func cmdDel(args *skel.CmdArgs) error {
...
-	if err != nil {
-		return err
-	}
-
-	ep := api.WorkloadEndpointMetadata{
-		Name:         args.IfName,
-		Node:         nodename,
-		Orchestrator: orchestrator,
-		Workload:     workload}
 	if err = calicoClient.WorkloadEndpoints().Delete(ep); err != nil {
```
robbrockbank: To do a atomic delete (in the sense no one else has modified it since you did your Get), you need to pass in the Metadata that was returned from the Get query - this will include revision data that etcd will use to determine whether the underlying entry has since been modified. 

So:  `calicoClient.WorkloadEndpoints().Delete(wep.Metadata)`

robbrockbank: You probably want to explicitly handle an error type of `errors.ErrorResourceUpdateConflict` - I'm not sure if you need to retry the operation (I can't think that you'd need to), but you may want to log out that we're no-oping due to a concurrent update.

gunjan5: I don't think we want to retry, I don't see where/why we need to do that. I'm logging it and returning the error in that case.

calico.go
tomdee: But `ActiveInstanceID` is exactly the workload ID, which is already part of the key. It seems strange to store it twice.
The DEL code path should handle already it not being there.

calico.go
```diff
@@ -306,6 +307,32 @@ func cmdDel(args *skel.CmdArgs) error {
...
+
+	ep := api.WorkloadEndpointMetadata{
+		Name:         args.IfName,
+		Node:         nodename,
+		Orchestrator: orchestrator,
+		Workload:     workload,
+	}
+
+	wep, err := calicoClient.WorkloadEndpoints().Get(ep)
+	if err != nil {
```
tomdee: This is a behaviour change. Before, if a endpoint couldn't be found, we'd still try to clean up the IP address. Now we won't.

gunjan5:
@tomdee I did some thinking about different cases and what clean up needs to be done for each case, my scribble below:
image

I updated the code, and added a lot of comments around the DEL logic.

PTAL

calico.go
```diff
@@ -306,57 +306,84 @@ func cmdDel(args *skel.CmdArgs) error {
...
 
-		logger.WithField("podCidr", dummyPodCidr).Info("Using a dummy podCidr to release the IP")
-		stdinData["ipam"].(map[string]interface{})["subnet"] = dummyPodCidr
+	var wep *api.WorkloadEndpoint
+	var endpoint api.WorkloadEndpointMetadata
 
-		args.StdinData, err = json.Marshal(stdinData)
+	// The following logic only applies to kubernetes since it sends multiple DELs for the same endpoint.
+	// No need to get WEP from the datastore and compare it's ActiveInstanceID with ContainerID for non-k8s orchestrators
+	if orchestrator == "k8s" {
```
tomdee: Can this block live in k8s.go since it's kubernetes specific?

calico.go
```diff
@@ -306,57 +306,84 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	}
 
-		logger.WithField("podCidr", dummyPodCidr).Info("Using a dummy podCidr to release the IP")
-		stdinData["ipam"].(map[string]interface{})["subnet"] = dummyPodCidr
+	var wep *api.WorkloadEndpoint
+	var endpoint api.WorkloadEndpointMetadata
 
-		args.StdinData, err = json.Marshal(stdinData)
+	// The following logic only applies to kubernetes since it sends multiple DELs for the same endpoint.
+	// No need to get WEP from the datastore and compare it's ActiveInstanceID with ContainerID for non-k8s orchestrators
```
tomdee: ...why? This comment would be more useful if you say why. e.g. ...because the ActiveInstanceID and ContainerID are the same value.

calico.go
```diff
@@ -306,57 +306,84 @@ func cmdDel(args *skel.CmdArgs) error {
...
-		"type": conf.IPAM.Type}).Debug("Looking for IPAM plugin in paths")
-
-	// We need to replace "usePodCidr" with a valid, but dummy podCidr string with "host-local" IPAM.
-	if conf.IPAM.Type == "host-local" && strings.EqualFold(conf.IPAM.Subnet, "usePodCidr") {
-
-		// host-local IPAM releases the IP by ContainerID, so podCidr isn't really used to release the IP.
-		// It just needs a valid CIDR, but it doesn't have to be the CIDR associated with the host.
-		dummyPodCidr := "0.0.0.0/0"
-		var stdinData map[string]interface{}
+	calicoClient, err := CreateClient(conf)
```
tomdee: Are you sure that creating a client doesn't actually connect to the datastore?

calico.go
gunjan5: we don't have a DEL func specific to k8s in k8s.go like we do for ADD, but I can move this part there if you'd like

calico.go
```diff
@@ -306,57 +306,84 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	// No need to get WEP from the datastore and compare it's ActiveInstanceID with ContainerID for non-k8s orchestrators
+	if orchestrator == "k8s" {
+		wep, err = calicoClient.WorkloadEndpoints().Get(ep)
 		if err != nil {
-			return err
+			if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
+
+				// We can talk to the datastore but WEP doesn't exist in there,
+				// but we still want to go ahead with the clean up. So, log a warning and clean up.
+				logger.WithField("WorkloadEndpoint", ep).Warningf("Error getting workloadendpoint: %v", err)
```
tomdee: A warning log, with "error" in the text? That's probably not a good idea. I think it would be good if the log also say that processing will continue.

calico.go
```diff
@@ -306,57 +306,84 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		} else {
+			endpoint = ep
+		}
 
-	if ipamErr != nil {
-		logger.Error(ipamErr)
+		// Check if ActiveInstanceID is populated (it will be an empty string "" if it was populated
+		// before this field was added to the API), and if it is there then compare it with ContainerID
+		// passed by the orchestrator to make sure they are the same, return without deleting if they aren't.
+		if endpoint.ActiveInstanceID != "" && args.ContainerID != endpoint.ActiveInstanceID {
```
tomdee: This code assumes that an ADD and DEL for the same workload ID (but different container ID) can't happen at the same time. If that is correct, that you should document that assumption in a comment.

calico.go
```diff
@@ -306,57 +306,84 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			logger.WithField("WorkloadEndpoint", wep).Warning("CNI_ContainerID does not match WorkloadEndpoint ActiveInstanceID so ignoring the DELETE cmd.")
+			return nil
+		}
 	}
 
-	calicoClient, err := CreateClient(conf)
-	if err != nil {
-		return err
+	// Store WEP Metadata in endpoint only if it is not nil to avoid nil pointer dereference
```
tomdee: Are you sure you need this? I don't think you do...

gunjan5: yes, it's needed because in case of non-k8s the value of `endpoint` is an empty struct of type `api.WorkloadEndpoint` which won't work, so we need to use the initial `ep` we created to delete the WEP

calico.go
gunjan5: It just creates the client, doesn't actually connect to the datastore. This was there before my PR, so no change in behavior here

calico.go
tomdee: There is something not right about this code. This block is duplicated with the section above, and the comment doesn't really explain why this is going on. Having `ep`, `wep` and `endpoint` doesn't help the situation.

calico.go
```diff
@@ -306,57 +306,84 @@ func cmdDel(args *skel.CmdArgs) error {
...
-			logger.WithField("endpoint", ep).Info("Endpoint object does not exist, no need to clean up.")
-		} else {
+	// Delete the WorkloadEndpoint object from the datastore.
+	if err = calicoClient.WorkloadEndpoints().Delete(endpoint); err != nil {
+		switch err := err.(type) {
+		case errors.ErrorResourceDoesNotExist:
+			// Log and proceed with the clean up if WEP doesn't exist.
+			logger.WithField("endpoint", endpoint).Info("Endpoint object does not exist, no need to clean up.")
+		case errors.ErrorResourceUpdateConflict:
+			// This case means the WEP object was modified between the time we did the Get and now,
```
tomdee: Why _don't_ you need to retry here?

calico.go
gunjan5: I've explained the "why" inside the code block in comments where I'm doing specific operations, but I'll add a line here too

calico.go
```diff
@@ -306,57 +306,84 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			return fmt.Errorf("Error deleting endpoint: endpoint was modified before it could be deleted: %v", err)
+		default:
 			return err
 		}
 	}
 
+	// Release the IP address by calling the configured IPAM plugin.
+	ipamErr := cleanUpIPAM(conf, args, logger)
+
 	// Only try to delete the device if a namespace was passed in.
```
tomdee: Since you've pulled cleanUpIpam into its own function, does it make sense to put the namespace cleanup in its own function too?

calico_cni_k8s_test.go
```diff
@@ -602,6 +619,147 @@ var _ = Describe("CalicoCni", func() {
 					Expect(pod2IP).Should(Equal(expectedIP))
 				})
 			})
+
+			// This specific test case is for an issue where k8s would send extra DELs being "aggrassive". See: https://github.com/kubernetes/kubernetes/issues/44100
```
tomdee: typo - aggrassive

calico_cni_test.go
tomdee: So can you change it back?

calico.go
gunjan5: I don't think retrying makes sense from CNI side since if some data changed while we were trying to Compare-and-Delete then retrying with the same data will still fail. 
It is the orchestrator's (k8s) job to send another DEL with the correct/different ContainerID if CNI plugin returns an error when k8s sent the DEL. k8s will do the retry here

calico.go
tomdee: Yes please

calico.go
```diff
@@ -306,57 +306,64 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		Node:         nodename,
+		Orchestrator: orchestrator,
+		Workload:     workload,
+	}
 
-		// host-local IPAM releases the IP by ContainerID, so podCidr isn't really used to release the IP.
-		// It just needs a valid CIDR, but it doesn't have to be the CIDR associated with the host.
-		dummyPodCidr := "0.0.0.0/0"
-		var stdinData map[string]interface{}
+	endpoint := &ep
```
tomdee: Why do you have an endpoint and an `ep`?

gunjan5: endpoint's value is ep for non-k8s orch and for k8s when we can't get WEP, so the Delete operation doesn't error out, and can still clean up. endpoint's value is wep.Metadata so if we got WEP back from the datastore then we want to use it to have the Revision to safely do a Compare-and-delete

tomdee: This isn't very clear from the naming of the variables. You need to name the variables better and/or comment it better.

k8s/k8s.go
```diff
@@ -294,6 +296,45 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
...
 
+// ContainerIDMismatchErr is for when ContainerID does not match ActiveInstanceID during Deletion process.
+var ContainerIDMismatchErr = errors.New("ContainerID does not match ActiveInstanceID")
+
+// CmdDelK8s performs some special checks for the "DEL" operation on a kubernetes pod.
+// The following logic only applies to kubernetes since it sends multiple DELs for the same endpoint.
+// We store CNI_CONTAINERID as ActiveInstanceID in WEP Metadata for k8s,
+// so we need to make sure we check if ContainerID and ActiveInstanceID are the same before deleting the pod.
+func CmdDelK8s(c *calicoclient.Client, ep api.WorkloadEndpointMetadata, args *skel.CmdArgs, logger *log.Entry) (*api.WorkloadEndpointMetadata, error) {
+
```
tomdee: I think we've had this before, but I don't like like the first line of a function to be blank.

k8s/k8s.go
```diff
@@ -294,6 +296,45 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
...
+		} else {
+			// Could not connect to datastore (connection refused, unauthorized, etc.)
+			// so we have no way of knowing/checking ActiveInstanceID. To protect the endpoint
+			// from false DEL, we return the error without deleting/cleaning up.
+			return nil, err
+		}
+	}
+
+	// Only compare ContainerID with ActiveInstanceID if we got WEP from the datastore.
+	if wep != nil {
```
tomdee: Would this be better as an `else` to the block above?

k8s/k8s.go
```diff
@@ -294,6 +296,45 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
...
 }
 
+// ContainerIDMismatchErr is for when ContainerID does not match ActiveInstanceID during Deletion process.
+var ContainerIDMismatchErr = errors.New("ContainerID does not match ActiveInstanceID")
+
+// CmdDelK8s performs some special checks for the "DEL" operation on a kubernetes pod.
+// The following logic only applies to kubernetes since it sends multiple DELs for the same endpoint.
+// We store CNI_CONTAINERID as ActiveInstanceID in WEP Metadata for k8s,
+// so we need to make sure we check if ContainerID and ActiveInstanceID are the same before deleting the pod.
+func CmdDelK8s(c *calicoclient.Client, ep api.WorkloadEndpointMetadata, args *skel.CmdArgs, logger *log.Entry) (*api.WorkloadEndpointMetadata, error) {
```
tomdee: Does this name describe what the function is doing? To me it seems that it's not deleting anything, it's just fetching a WEPMeta?

calico.go
```diff
@@ -306,57 +306,68 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			if err == k8s.ContainerIDMismatchErr {
+				// CNI_ContainerID does not match WorkloadEndpoint ActiveInstanceID so ignoring the DELETE cmd.
+				return nil
+			}
 			return err
 		}
 
-		logger.WithField("podCidr", dummyPodCidr).Info("Using a dummy podCidr to release the IP")
-		stdinData["ipam"].(map[string]interface{})["subnet"] = dummyPodCidr
+		if wep != nil {
```
tomdee: I think it's slightly abnormal for a function to return a nil error *and* and nil value, so probably worth a comment to note that.

calico.go
tomdee: I think this comment got lost

calico.go
tomdee: The important point is that we don't care about all the data changing - we only really care about the containerID/activeworkloadID. Are you sure that k8s will retry and we won't leak endpoints? This is related to my comment elsewhere about concurrent ADD/DEL requests

k8s/k8s.go
gunjan5: both cases in the `if` above that `return` so there is no point of adding an else, it just makes the code harder to read. Also the linter yells :)

glide.yaml
```diff
@@ -15,7 +15,7 @@ import:
   subpackages:
   - gexec
 - package: github.com/projectcalico/libcalico-go
-  version: v1.1.0
+  version: 0a0ff81244b014c0d4076573edcc9e73694e673a
```
tomdee: Isn't there a release with the fixes in?

k8s/k8s.go
tomdee: No they don't both return, which is why you need the extra if check.

glide.yaml
gunjan5: it's an RC not an actual release yet

calico.go
```diff
@@ -306,57 +306,68 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	endpoint := &ep
 
-		if err := json.Unmarshal(args.StdinData, &stdinData); err != nil {
+	// Handle k8s specific bits of handling the DEL.
+	if orchestrator == "k8s" {
+		wep, err := k8s.CmdDelK8s(calicoClient, ep, args, logger)
+		if err != nil {
+			if err == k8s.ContainerIDMismatchErr {
+				// CNI_ContainerID does not match WorkloadEndpoint ActiveInstanceID so ignoring the DELETE cmd.
+				return nil
```
robbrockbank: When k8s recreates the container - does it use the same IP address, or should we still be going through the IPAM deletion processing.

calico.go
```diff
@@ -306,57 +306,68 @@ func cmdDel(args *skel.CmdArgs) error {
...
-		stdinData["ipam"].(map[string]interface{})["subnet"] = dummyPodCidr
+		if wep != nil {
+			endpoint = wep
+		}
+	}
 
-		args.StdinData, err = json.Marshal(stdinData)
-		if err != nil {
+	// Delete the WorkloadEndpoint object from the datastore.
+	if err = calicoClient.WorkloadEndpoints().Delete(*endpoint); err != nil {
```
robbrockbank: Can we just move the actual deletion of the endpoint into the k8s code too - so basically once you call CmdDelK8s it will actually have deleted the WEP.  I think that would be more symmetric with the ADD no?

Yes it means you have two bits of code to handle the delete, one in k8s and one here for non-k8s - but that means wouldn't need to worry about resource update conflicts here (just in the k8s code where it makes more sense).

robbrockbank: Also, although I can't think of an actual scenario where it is likely to happen, in the k8s code when you do have the resource update conflict I think you should retry as Tom suggests.  You need to retry from the Get() so that the revision and active instance ID are up to date.  If the WEP has had the active instance ID updated then you'll know about it and be able to log that with certainty.

k8s/k8s.go
robbrockbank: If you added a `return nil, nil` to the ErrorResourceDoesNotExist branch then you could lose the `web != nil` check and unindent this whole section.

glide.yaml
robbrockbank: Yup:  v1.2.1-rc4

k8s/k8s.go
```diff
@@ -294,6 +296,45 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
...
+
+	wep, err := c.WorkloadEndpoints().Get(ep)
+	if err != nil {
+		if _, ok := err.(cerrors.ErrorResourceDoesNotExist); ok {
+			// We can talk to the datastore but WEP doesn't exist in there,
+			// but we still want to go ahead with the clean up. So, log a warning and clean up.
+			logger.WithField("WorkloadEndpoint", ep).Warning("WorkloadEndpoint does not exist in the datastore, moving forward with the clean up")
```
robbrockbank: I _think_ you mean to return nil, nil here (hence your comment below).  If that is the case then:

You can lose the next `else` and you can drop the `if wep != nil {` below and unindent that whole section.

k8s/k8s.go
```diff
@@ -294,6 +296,45 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
 	return result, nil
 }
 
+// ContainerIDMismatchErr is for when ContainerID does not match ActiveInstanceID during Deletion process.
+var ContainerIDMismatchErr = errors.New("ContainerID does not match ActiveInstanceID")
```
robbrockbank: I think the concept of the Container ID mismatch should be confined to the k8s code.  If you performed the remainder of the k8s WEP delete processing in CmdDelK8s you could achieve that.

k8s/k8s.go
gunjan5: yeah that makes sense

calico.go
gunjan5: k8s doesn't recreate with the same IP, it does the ADD as if it's a new pod, so we still need to do the IPAM clean up

calico.go
gunjan5: Retrying in CNI doesn't make sense for a couple of reasons:
1. calling CNI to setUp/tearDown pod is like a one time job, so if it fails kubelet retries with the latest CNI ADD/DEL parameters, so CNI keep retrying with what info we have (which might change from k8s Pov) defeats that purpose
2. This would be a change in behavior, CNI returns an error if it can't delete the wep instead of retrying (i know it's under different case this time since we do a get, and CaD, but it's still a change in behavior)

I'm mostly saying no because of the first reason I stated above

calico.go
```diff
@@ -306,78 +301,40 @@ func cmdDel(args *skel.CmdArgs) error {
...
-		Workload:     workload}
-	if err = calicoClient.WorkloadEndpoints().Delete(ep); err != nil {
+		Workload:     workload,
+	}
+
+	// Handle k8s specific bits of handling the DEL.
+	if orchestrator == "k8s" {
+		return k8s.CmdDelK8s(calicoClient, wep, args, conf, logger)
+	}
+
```
tomdee: Previously, the IPAM delete happened _before_ the WEP delete. Why have you moved it?

calico_cni_test.go
```diff
@@ -297,4 +308,61 @@ var _ = Describe("CalicoCni", func() {
 			})
 		})
 	})
+
+	Describe("ADD a container with ContainerID and DEL it with the same ContainerID", func() {
```
tomdee: How is this test different from the first test in this file? Do we really need both?

k8s/k8s.go
```diff
@@ -294,6 +296,69 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
...
+	} else if wep.Metadata.ActiveInstanceID != "" && args.ContainerID != wep.Metadata.ActiveInstanceID {
+		logger.WithField("WorkloadEndpoint", wep).Warning("CNI_ContainerID does not match WorkloadEndpoint ActiveInstanceID so ignoring the DELETE cmd.")
+		return nil
+	}
+
+	// Delete the WorkloadEndpoint object from the datastore.
+	// In case of k8s, where we are deleting the WEP we got from the Datastore,
+	// this Delete is a Compare-and-Delete, so if *any* field in the WEP changed from
+	// the time we get WEP until here then the Delete operation will fail.
+	if err = c.WorkloadEndpoints().Delete(wep.Metadata); err != nil {
```
tomdee: Why do we try to delete even if the get just failed?

k8s/k8s.go
```diff
@@ -294,6 +296,69 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
...
+	// Clean up namespace by removing the interfaces.
+	err = utils.CleanUpNamespace(args, logger)
+	if err != nil {
+		return err
+	}
+
+	// Return the IPAM error if there was one. The IPAM error will be lost if there was also an error in cleaning up
+	// the device or endpoint, but crucially, the user will know the overall operation failed.
+	return ipamErr
+
```
tomdee: remove this blank line

utils/utils.go
```diff
@@ -40,6 +44,75 @@ func min(a, b int) int {
...
+	var err error
+
+	// Always try to release the address. Don't deal with any errors till the endpoints are cleaned up.
+	fmt.Fprintf(os.Stderr, "Calico CNI releasing IP address\n")
+	logger.WithFields(log.Fields{"paths": os.Getenv("CNI_PATH"),
+		"type": conf.IPAM.Type}).Debug("Looking for IPAM plugin in paths")
+
+	// We need to replace "usePodCidr" with a valid, but dummy podCidr string with "host-local" IPAM.
+	if conf.IPAM.Type == "host-local" && strings.EqualFold(conf.IPAM.Subnet, "usePodCidr") {
+
```
tomdee: remove this blank line

utils/utils.go
```diff
@@ -40,6 +44,75 @@ func min(a, b int) int {
...
+	return nil
+}
+
+// CleanUpIPAM calls IPAM plugin to release the IP address.
+// It also contains IPAM plugin specific changes needed before calling the plugin.
+func CleanUpIPAM(conf NetConf, args *skel.CmdArgs, logger *log.Entry) error {
+	var err error
+
+	// Always try to release the address. Don't deal with any errors till the endpoints are cleaned up.
```
tomdee: This comment feels out of place

utils/utils.go
```diff
@@ -40,6 +44,75 @@ func min(a, b int) int {
...
+		}
+	}
+
+	return nil
+}
+
+// CleanUpIPAM calls IPAM plugin to release the IP address.
+// It also contains IPAM plugin specific changes needed before calling the plugin.
+func CleanUpIPAM(conf NetConf, args *skel.CmdArgs, logger *log.Entry) error {
+	var err error
```
tomdee: Why is this up here?

calico.go
gunjan5: that is because if the deleting WEP fails we return an error so DEL operation doesn't go through, I thought moving IPAM deletion after makes sense, but I guess it makes more sense in k8s.go not so much here

k8s/k8s.go
gunjan5: I added an else there so it only tries to delete when get succeeds (forgot to save and commit it)

k8s/k8s.go
```diff
@@ -322,13 +322,12 @@ func CmdDelK8s(c *calicoclient.Client, ep api.WorkloadEndpointMetadata, args *sk
...
 		logger.WithField("WorkloadEndpoint", wep).Warning("CNI_ContainerID does not match WorkloadEndpoint ActiveInstanceID so ignoring the DELETE cmd.")
 		return nil
-	}
 
-	// Delete the WorkloadEndpoint object from the datastore.
-	// In case of k8s, where we are deleting the WEP we got from the Datastore,
-	// this Delete is a Compare-and-Delete, so if *any* field in the WEP changed from
-	// the time we get WEP until here then the Delete operation will fail.
-	if err = c.WorkloadEndpoints().Delete(wep.Metadata); err != nil {
+		// Delete the WorkloadEndpoint object from the datastore.
```
tomdee: These comments are indented too much

tomdee: ...so are the ones on 319:321

utils/utils.go
```diff
@@ -75,22 +75,19 @@ func CleanUpNamespace(args *skel.CmdArgs, logger *log.Entry) error {
...
 	// We need to replace "usePodCidr" with a valid, but dummy podCidr string with "host-local" IPAM.
 	if conf.IPAM.Type == "host-local" && strings.EqualFold(conf.IPAM.Subnet, "usePodCidr") {
-
 		// host-local IPAM releases the IP by ContainerID, so podCidr isn't really used to release the IP.
 		// It just needs a valid CIDR, but it doesn't have to be the CIDR associated with the host.
 		dummyPodCidr := "0.0.0.0/0"
 		var stdinData map[string]interface{}
 
-		if err := json.Unmarshal(args.StdinData, &stdinData); err != nil {
+		err := json.Unmarshal(args.StdinData, &stdinData)
```
tomdee: nit: I don't really understand why you changed these lines but I don't think it matters too much

k8s/k8s.go
gunjan5: that's all `go fmt`

utils/utils.go
gunjan5: this is because of line 97 where args.StdinData: `non-name args.StdinData on left side of :=
` so have to reuse the `err` variable.

k8s/k8s.go
tomdee: 👍

utils/utils.go
tomdee: OK, makes sense. I didn't see that bit of code.

PR #319, opened 2017-04-27
bugfix: parse IP, pass ip version in result + tests

### LINKED ISSUE #317: Calico CNI ignoring IP Version and Gateway info for a second ADD with host-local IPAM
**What's the issue?**:
host-local IPAM returns gateway as part of the `Result` but while Calico CNI stores it to a WEP it doesn't store the gateway, so when an orchestrator calls an `ADD` for an existing endpoint we create a `result` from WEP but it's missing the `Gateway` info this time.

**What is affected?**:
This affects all CNI versions with `host-local` only when an `ADD` is called for an existing pod/container/endpoint. For example, KDD with k8s sending another `ADD` for the same pod would see this issue since it uses Calico CNI with `host-local` IPAM

### LINKED ISSUE #318: IP unmarshalling error with a duplicate ADD for an existing pod/container
**What's the issue?**:
If an orchestrator sends an `ADD` for an existing pod/container with `cniVersion` specified as `0.3.0` or `0.3.1` then Calico CNI will try to produce `result` from the WEP, and it will fail to unmarshal it since the field name has changed from `ip` to `address` in the new CNI spec.

**Who is affected?**:
This is a regression issue so CNI version `v1.7.0` **_only_** when/if the orchestrator sends multiple `ADD`s for the same pod/container, but this should not affect an already running pod, it will just throw all kinds or errors in k8s logs and pod status won't be in `Running` state.

gunjan5:
fixes #317 
fixes #318 
also getting rid of dot imports slowly

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -19,7 +19,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
-	. "github.com/projectcalico/cni-plugin/test_utils"
+	tu "github.com/projectcalico/cni-plugin/test_utils"
 	"github.com/projectcalico/cni-plugin/utils"
 	"github.com/projectcalico/libcalico-go/lib/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/k8s"
@@ -35,9 +35,10 @@ func init() {
 
 var _ = Describe("CalicoCni", func() {
 	hostname, _ := os.Hostname()
+	K8S_TEST_NS := tu.K8S_TEST_NS
 	BeforeEach(func() {
-		WipeK8sPods()
-		WipeEtcd()
+		tu.WipeK8sPods()
+		tu.WipeEtcd()
 	})
 
 	Describe("Run Calico CNI plugin in K8s mode", func() {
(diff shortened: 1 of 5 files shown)


calico_cni_k8s_test.go
```diff
@@ -759,6 +759,97 @@ var _ = Describe("CalicoCni", func() {
...
+			    		"type": "calico-ipam"
+			        	},
+				"kubernetes": {
+					  "k8s_api_root": "http://127.0.0.1:8080"
+					 },
+				"policy": {"type": "k8s"},
+				"log_level":"info"
+				}`, cniVersion, os.Getenv("ETCD_IP"))
+
+				It("nodename takes precedence over hostname", func() {
```
tomdee: What does this mean?

calico_cni_test.go
```diff
@@ -309,4 +309,67 @@ var _ = Describe("CalicoCni", func() {
...
+			  "etcd_endpoints": "http://%s:2379",
+			  "hostname": "namedHostname",
+			  "nodename": "namedNodename",
+			  "ipam": {
+			    "type": "host-local",
+			    "subnet": "10.0.0.0/24"
+			  }
+			}`, cniVersion, os.Getenv("ETCD_IP"))
+
+			It("nodename takes precedence over hostname", func() {
```
tomdee: Again, I think this might just be a copy paste error?

calico_cni_test.go
```diff
@@ -309,4 +309,67 @@ var _ = Describe("CalicoCni", func() {
...
+
+	Describe("ADD for an existing endpoint", func() {
+		Context("Create a container then send another ADD for the same container", func() {
+			netconf := fmt.Sprintf(`
+			{
+			  "cniVersion": "%s",
+			  "name": "net1",
+			  "type": "calico",
+			  "etcd_endpoints": "http://%s:2379",
+			  "hostname": "namedHostname",
```
tomdee: Why both?

test_utils/utils.go
```diff
@@ -250,6 +250,60 @@ func CreateContainerWithId(netconf string, k8sName string, ip string, containerI
 	return
 }
 
+
+// RunCNIPluginWithId calls CNI plugin with a containerID and targetNs passed to it.
+// This is for when you want to call CNI for an existing container.
+func RunCNIPluginWithId(netconf string, k8sName string, ip string, netnspath, containerId string, targetNs ns.NetNS) (session *gexec.Session, contVeth netlink.Link, contAddr []netlink.Addr, contRoutes []netlink.Route, err error) {
```
tomdee: Why doesn't this share more code with CreateContainerWithID?

PR #320, opened 2017-04-27
Add sysctl and route programming to CNI + tests

### LINKED ISSUE #306: Plugin should create host routes (rather than relying on felix to do it)
It should also do the other things that felix does (e.g. setting sysctls on the interface - TBD what that full list is)

gunjan5:
fixes #306 
Calico CNI relies on Felix to configure host side route(s) and sysctls for the host veth for the pod currently, with this Calico CNI should be able to do that itself.

Route programming and interface sysctl config happens during ADD, and gets destroyed with DEL as the interface gets removed, which removes routes and sysctls.

### DIFF
--- a/utils/network.go
+++ b/utils/network.go
@@ -2,7 +2,9 @@ package utils
 
 import (
 	"fmt"
+	"io"
 	"net"
+	"os"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/containernetworking/cni/pkg/ip"
@@ -15,8 +17,9 @@ import (
 // DoNetworking performs the networking for the given config and IPAM result
 func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logger *log.Entry, desiredVethName string) (hostVethName, contVethMAC string, err error) {
 	// Select the first 11 characters of the containerID for the host veth.
-	hostVethName = "cali" + args.ContainerID[:min(11, len(args.ContainerID))]
+	hostVethName = "cali" + args.ContainerID[:Min(11, len(args.ContainerID))]
 	contVethName := args.IfName
+	var hasIPv4, hasIPv6 bool
 
 	// If a desired veth name was passed in, use that instead.
 	if desiredVethName != "" {
@@ -92,7 +95,9 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
 				if err = netlink.AddrAdd(contVeth, &netlink.Addr{IPNet: &addr.Address}); err != nil {
 					return fmt.Errorf("failed to add IP addr to %q: %v", contVethName, err)
 				}
-
+				// Set hasIPv4 to true so sysctls for IPv4 can be programmed when the host side of
+				// the veth finishes moving to the host namespace.
+				hasIPv4 = true
 			}
 
 			// Handle IPv6 routes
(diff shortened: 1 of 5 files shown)


utils/network.go
```diff
@@ -140,6 +149,11 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
 		return "", "", err
 	}
 
+	err = configureSysctls(hostVethName, hasIPv4, hasIPv6)
+	if err != nil {
+		return "", "", fmt.Errorf("error configuring sysctls for the interface: %s, error: %s", hostVethName, err)
```
tomdee: nit: I think you should drop "the" before interface.

utils/network.go
```diff
@@ -151,5 +165,101 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
...
+
+// configureSysctls configures necessary sysctls required for the host side of the veth pair for IPv4 and/or IPv6.
+func configureSysctls(hostVethName string, hasIPv4, hasIPv6 bool) error {
+	var err error
+
+	if hasIPv4 {
+		// Enable proxy ARP, this makes the host respond to all ARP requests with its own
+		// MAC.  This has a couple of advantages:
+		//
+		// - In OpenStack, we're forced to configure the guest's networking using DHCP.
```
tomdee: I don't think this comment has any place here

utils/utils.go
```diff
@@ -37,7 +37,7 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
-func min(a, b int) int {
+func Min(a, b int) int {
```
tomdee: Why did you make this change?

tomdee: Aha, I see you're using it in the tests

gunjan5:
My manual testing results (it passed 🎉 ) :
- Start a k8s cluster with `calico/cni:master`
- start an nginx pod (Felix + CNI will work on it as normal)
- Note down the routes and sysctls it programs, then delete the pod
- Now kill the felix runsv service
- Copy the CNI binary built from this branch of the code into `/opt/cni/bin`
- Start another nginx pod (this time Felix is not running, so CNI should program the routes/sysctls)
- now delete the pod and make sure routes and sysctls go away.

Here is the result:
- With Felix running:
Route:
```bash
root@node-01:/home/vagrant/go/src/github.com/projectcalico$ route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
192.168.190.4   0.0.0.0         255.255.255.255 UH    0      0        0 calic039d95a607
```
Sysctls:
```bash
root@node-01:/home/vagrant/go/src/github.com/projectcalico/cni-plugin# cat /proc/sys/net/ipv4/conf/calic039d95a607/proxy_arp
1
root@node-01:/home/vagrant/go/src/github.com/projectcalico/cni-plugin# cat /proc/sys/net/ipv4/conf/calic039d95a607/forwarding
1
root@node-01:/home/vagrant/go/src/github.com/projectcalico/cni-plugin# cat /proc/sys/net/ipv4/neigh/calic039d95a607/proxy_delay
0
```

- Without Felix running:
Routes:
```bash
root@node-01:/home/vagrant/go/src/github.com/projectcalico$ route -n
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
192.168.190.13  0.0.0.0         255.255.255.255 UH    0      0        0 cali4782ba18a94
```
Sysctls:
```bash
root@node-01:/home/vagrant/go/src/github.com/projectcalico$ cat /proc/sys/net/ipv4/conf/cali4782ba18a94/proxy_arp
1
root@node-01:/home/vagrant/go/src/github.com/projectcalico$ cat /proc/sys/net/ipv4/conf/cali4782ba18a94/forwarding
1
root@node-01:/home/vagrant/go/src/github.com/projectcalico$ cat /proc/sys/net/ipv4/neigh/cali4782ba18a94/proxy_delay
0
```

/CC @tomdee

tomdee:
Discussed offline with @gunjan5 and did some additional testing. We found that to ping between containers when felix has never been up, we need to ensure that iptables allows the traffic (e.g. the FORWARD table passes the traffic) and that the following sysctl is set - `sysctl -w net.ipv4.ip_forward=1`

PR #326, opened 2017-05-03
Release IPs on bad IPAM result

### LINKED ISSUE #325: CNI plugin can leak IP addresses
Seen on Calico CNI v1.8.0
When the calico CNI plugin calls an IPAM plugin, it can end up returning an error to the runtime even though the IPAM plugin didn't return an error. For this to be OK, the networking plugin needs to ensure that it is instructing the IPAM plugin to cleanup, otherwise addresses can be leaked.

e.g. in this code, at least the last two return statements (and maybe the first one too) should be calling ipam.ExecDel.
```
		case ipAddrs == "" && ipAddrsNoIpam == "":
			// Call IPAM plugin if ipAddrsNoIpam or ipAddrs annotation is not present.
			logger.Debugf("Calling IPAM plugin %s", conf.IPAM.Type)
			ipamResult, err := ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
			if err != nil {
				return nil, err
			}
			logger.Debugf("IPAM plugin returned: %+v", ipamResult)

			// Convert IPAM result into current Result.
			// IPAM result has a bunch of fields that are optional for an IPAM plugin
			// but required for a CNI plugin, so this is to populate those fields.
			// See CNI Spec doc for more details.
			result, err = current.NewResultFromResult(ipamResult)
			if err != nil {
				return nil, err
			}

			if len(result.IPs) == 0 {
				return nil, errors.New("IPAM plugin returned missing IP config")
			}
```

caseydavenport:
Fixes #325

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 
 SRCFILES=calico.go $(wildcard utils/*.go) $(wildcard k8s/*.go) ipam/calico-ipam.go
 TEST_SRCFILES=$(wildcard test_utils/*.go) $(wildcard calico_cni_*.go)
-LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
+LOCAL_IP_ENV?=$(shell docker run -ti --rm --net=host busybox ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
 
 # fail if unable to download
 CURL=curl -sSf
@@ -109,6 +109,9 @@ $(DEPLOY_CONTAINER_MARKER): Dockerfile build-containerized fetch-cni-bins
 .PHONY: fetch-cni-bins
 fetch-cni-bins: dist/flannel dist/loopback dist/host-local
 
+dist/echo-ipam:
+	cp utils/echo-ipam dist/
+
 dist/flannel dist/loopback dist/host-local:
 	mkdir -p dist
 	$(CURL) -L --retry 5 https://github.com/containernetworking/cni/releases/download/$(CNI_VERSION)/cni-amd64-$(CNI_VERSION).tgz | tar -xz -C dist ./flannel ./loopback ./host-local
@@ -117,7 +120,7 @@ dist/flannel dist/loopback dist/host-local:
 # .go-pkg-cache can't be used (since tests run as root)
 .PHONY: test-containerized
 ## Run the tests in a container (as root)
-test-containerized: run-etcd run-k8s-apiserver build-containerized dist/host-local
+test-containerized: run-etcd run-k8s-apiserver build-containerized dist/host-local dist/echo-ipam
 	docker run --rm --privileged --net=host \
 	-e ETCD_IP=$(LOCAL_IP_ENV) \
 	-e LOCAL_USER_ID=0 \
@@ -145,7 +148,7 @@ run-test-containerized-without-building: run-etcd run-k8s-apiserver
 ## Run the tests in a container (as root) for different CNI spec versions
 ## to make sure we don't break backwards compatiblity.
 .PHONY: test-containerized-cni-versions
-test-containerized-cni-versions: build-containerized dist/host-local;
+test-containerized-cni-versions: build-containerized dist/host-local dist/echo-ipam;
 	for cniversion in "0.2.0" "0.3.1" ; do \
 		make run-test-containerized-without-building CNI_SPEC_VERSION=$$cniversion; \
 	done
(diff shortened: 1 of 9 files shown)


gunjan5:
@caseydavenport I think we also need to do this at https://github.com/projectcalico/cni-plugin/blob/master/calico.go#L157 and https://github.com/projectcalico/cni-plugin/blob/master/calico.go#L162

calico_cni_k8s_test.go
```diff
@@ -885,6 +885,59 @@ var _ = Describe("CalicoCni", func() {
 					_, err = DeleteContainer(netconf, netnspath, "")
 					Expect(err).ShouldNot(HaveOccurred())
 				})
+
+				Context("when IPAM returns an empty result", func() {
```
caseydavenport: This isn't done yet, didn't mean to add this to the PR yet :)

Makefile
```diff
@@ -4,7 +4,7 @@
 
 SRCFILES=calico.go $(wildcard utils/*.go) $(wildcard k8s/*.go) ipam/calico-ipam.go
 TEST_SRCFILES=$(wildcard test_utils/*.go) $(wildcard calico_cni_*.go)
-LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
+LOCAL_IP_ENV?=$(shell docker run -ti --rm --net=host busybox ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
```
caseydavenport: Just so this now works on mac as well :)

calico_cni_k8s_test.go
```diff
@@ -885,6 +885,55 @@ var _ = Describe("CalicoCni", func() {
...
+							Spec: v1.PodSpec{Containers: []v1.Container{{
+								Name:  fmt.Sprintf("container-%s", name),
+								Image: "ignore",
+							}}},
+						})
+						Expect(err).NotTo(HaveOccurred())
+
+						logger.Infof("Created POD object: %v", pod)
+
+						_, netnspath, _, _, _, _, _, err := CreateContainer(netConf, name, "")
```
caseydavenport: So this seems to be working (can tell from the logs), but I don't know how to assert that the IPs have been actually released by the plugin.

I could update `echo-ipam` to do some fake IP allocation / releasing perhaps using files on disk and then assert that the file is removed in the test?

@tomdee

Makefile
tomdee: Lol, nice workaround.

Makefile
```diff
@@ -109,6 +109,9 @@ $(DEPLOY_CONTAINER_MARKER): Dockerfile build-containerized fetch-cni-bins
 .PHONY: fetch-cni-bins
 fetch-cni-bins: dist/flannel dist/loopback dist/host-local
 
+dist/echo-ipam:
```
tomdee: Why not just store echo-ipam in dist?

calico.go
```diff
@@ -156,10 +156,12 @@ func cmdAdd(args *skel.CmdArgs) error {
 			// See CNI Spec doc for more details.
 			result, err = current.NewResultFromResult(ipamResult)
 			if err != nil {
+				ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
```
tomdee: How much thought have you put into checking all the places where this should be called?

tomdee: And this probably needs test coverage too

calico_cni_k8s_test.go
```diff
@@ -885,6 +885,55 @@ var _ = Describe("CalicoCni", func() {
...
+				Context("when IPAM returns an empty result", func() {
+					It("should try to release any IP addresses", func() {
+						netConf := fmt.Sprintf(`
+				{
+			      "cniVersion": "%s",
+				  "name": "net2",
+				  "type": "calico",
+				  "etcd_endpoints": "http://%s:2379",
+			 	  "ipam": {
+			    	 "type": "echo-ipam",
```
tomdee: formatting (indent)

calico_cni_k8s_test.go
```diff
@@ -885,6 +885,55 @@ var _ = Describe("CalicoCni", func() {
...
+				 "return": "{}"
```
tomdee: I don't really get this line, is it left over from a previous design of your echo plugin?

caseydavenport: It is indeed.

Makefile
caseydavenport: `dist` isn't checked in at the moment.  I could do that if so desired though.  LMK.

robbrockbank:
@caseydavenport :  should this be closed?

PR #328, opened 2017-05-05
Add github templates

tomdee:


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+3.
+4.
+
+## Context
+<!--- How has this issue affected you? What are you trying to accomplish? -->
+<!--- Providing context helps us come up with a solution that is most useful in the real world -->
+
+## Your Environment
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+* CNI plugin version:
```
caseydavenport: This should probably say how to figure that out - people get confused about versions.

e.g:

```
* CNI plugin version (e.g. `/opt/cni/bin/calico --version`)
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+* Orchestrator version (e.g. kubernetes, mesos, rkt):
```
caseydavenport: Maybe "Orchestrator and version".

We want to know both what orchestrator it is, and the version of the orchestrator they're using.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,34 @@
...
+* CNI config:
+* Operating System and version:
```
caseydavenport: Again, is there a specific command we want for this one?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,12 @@
+## Description
+A few sentences describing the overall goals of the pull request's commits. 
+Please include
```
caseydavenport: This should also include links to any issues that this PR fixes.

.github/ISSUE_TEMPLATE.md
tomdee: I can't think of one

PR #329, opened 2017-05-09
Release IPs on bad IPAM result

caseydavenport:
A version of #326 without the hokey tests which aren't working yet.


caseydavenport:
@tomdee I haven't been able to get the tests I wrote in the other PR working.

I think this fix is simple enough that a manual test will do in order to get it into v2.2.

PR #333, opened 2017-05-09
validate if more than one IPv4/v6 are passed with ipAddrsNoIpam/ipAdd…

gunjan5:
…rs annotations

also print logs in string form, not byte. It makes it really hard to debug.

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -104,7 +104,7 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
 			if err != nil {
 				return nil, err
 			}
-			logger.WithField("stdin", args.StdinData).Debug("Updated stdin data")
+			logger.WithField("stdin", string(args.StdinData)).Debug("Updated stdin data")
 		}
 
 		labels := make(map[string]string)
@@ -170,7 +170,7 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
 						return nil, err
 					}
 					args.StdinData = newData
-					logger.WithField("stdin", args.StdinData).Debug("Updated stdin data")
+					logger.WithField("stdin", string(args.StdinData)).Debug("Updated stdin data")
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


k8s/k8s.go
```diff
@@ -503,9 +517,20 @@ func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) (*current.Resul
...
+			// Check if there is already has been an IPv4 in the list, as we only support one IPv4 and/or one IPv6 per interface for now.
+			if hasIPv4 {
+				return nil, fmt.Errorf("cannot have more than one IPv4 address for \"cni.projectcalico.org/ipAddrsNoIpam\" annotation")
+			}
+			hasIPv4 = true
 			version = "4"
 			mask = net.CIDRMask(32, 32)
+
 		} else {
+			// Check if there is already has been an IPv4 in the list, as we only support one IPv4 and/or one IPv6 per interface for now.
```
robbrockbank: IPv4 -> IPv6

k8s/k8s.go
```diff
@@ -388,6 +389,18 @@ func ipAddrsResult(ipAddrs string, conf utils.NetConf, args *skel.CmdArgs, logge
...
+			if hasIPv4 {
+				return nil, fmt.Errorf("cannot have more than one IPv4 address for \"cni.projectcalico.org/ipAddrs\" annotation")
+			}
+			hasIPv4 = true
+		} else {
+			if hasIPv6 {
+				return nil, fmt.Errorf("cannot have more than one IPv6 address for \"cni.projectcalico.org/ipAddrs\" annotation")
+			}
+			hasIPv6 = true
+		}
```
robbrockbank: This will allocate at least one IP address before it fails - are we sure the tidy up will work correctly?

Should we check upfront that there is only a single IPv4 and single IPv6 address in the list *before* doing any assignment.

robbrockbank: Ditto for branch below too.

k8s/k8s.go
```diff
@@ -513,17 +495,65 @@ func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) (*current.Resul
...
 }
 
+// validateAndExtractIPs is a utility function that validates the passed IP list to make sure
+// there is one IPv4 and/or one IPv6 and then returns the slice of IPs.
+func validateAndExtractIPs(ipAddrs string, annotation string, logger *log.Entry) ([]net.IP, error) {
+
+	// Parse IPs from JSON.
+	ips, err := parseIPAddrs(ipAddrs, logger)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse IPs %s for annotation \"cni.projectcalico.org/ipAddrs\": %s", ipAddrs, err)
```
robbrockbank: Need to use `annotation` here.

k8s/k8s.go
```diff
@@ -513,17 +495,65 @@ func overrideIPAMResult(ipAddrsNoIpam string, logger *log.Entry) (*current.Resul
...
+		logger.Debugf("Adding IPv%s: %s to result", ipConf.Version, ip.String())
 	}
 
 	return &result, nil
 }
 
+// validateAndExtractIPs is a utility function that validates the passed IP list to make sure
+// there is one IPv4 and/or one IPv6 and then returns the slice of IPs.
+func validateAndExtractIPs(ipAddrs string, annotation string, logger *log.Entry) ([]net.IP, error) {
+
```
robbrockbank: no blank line here

PR #338, opened 2017-05-31
Add portmap plugin

caseydavenport:
## Description
Adds the portmap plugin from https://github.com/containernetworking/plugins/pull/1 to the cni installer.  Currently, there is no release of the portmap plugin so this PR needs to wait until there is one.

## Todos
- [x] Include portmap plugin, or reference upstream release.
- [ ] Documentation

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,3 +6,4 @@
 !dist/flannel
 !dist/loopback
 !dist/host-local
+!dist/portmap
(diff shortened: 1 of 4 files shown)


tomdee:
I'm not super stoked about having that binary checked in, since it will live in the history forever. But otherwise, the change looks fine.

tomdee:
Actually - please also say (somewhere) exactly what git hash the binary is built from.

gunjan5:
What's wrong with leaving it as a Dockerhub image (with binary in it) until the plugin PR is merged so we don't have to upload the binary to the repo?

caseydavenport:
> What's wrong with leaving it as a Dockerhub image (with binary in it) until the plugin PR is merged so we don't have to upload the binary to the repo?

We could do that.  It would mean that building that version of the calico/cni container wouldn't be reproducible (but we could include the commit hash from which it was built to make that easier?)

Issue #339, opened 2017-06-01
Provide an option to skip installing other cni binaries even when absent


abhinavdahiya:
## Current Behavior

https://github.com/projectcalico/cni-plugin/blob/master/k8s-install/scripts/install-cni.sh#L56-L64
script installs other cni binaries when they are absent.

## Possible Solution

It would be great if there option to disable copying binaries even if they were not present. like:
```bash
INSTALL_OTHER_CNI_BINARIES=${INSTALL_OTHER_CNI_BINARIES:-false}

if [[ ${INSTALL_OTHER_CNI_BINARIES} = "true" && ! -f /host/opt/cni/bin/flannel ]]; then
cp /opt/cni/bin/flannel /host/opt/cni/bin/
fi
```
## Context

For example I am trying to use flannel+calico and want flannel manifest to copy its binary version to correct directory. There might be conflicts depending on who executes first.

caseydavenport:
This sounds useful to me.

Probably on a per-plugin basis? e.g. you might want host-local but not flannel?

abhinavdahiya:
@caseydavenport So there would be INSTALL_FLANNEL_CNI_BINARY, INSTALL_HOSTLOCAL_CNI_BINARY and INSTALL_LOOPBACK_CNI_BINARY options ??

Not sure if individual options would be necessary. But if it doesn't create too much complexity it would be nice.

caseydavenport:
Yeah, separate options would work fine.  Or maybe a single one with multiple clauses?

e.g.

`INSTALL_BINARIES=calico,calico-ipam,flannel,host-local,loopback,portmap`

Or,

`SKIP_BINARIES=flannel,loopback`

tmjd:
Fixed by #341

PR #341, opened 2017-06-05
Use comma sep. string to set install binaries

abhinavdahiya:
addresses #339

Update the install-cni.sh to loop through /opt/cni/bin/* and copy only binaries that are requested in comma string as env variable INSTALL_CNI_BINARIES

Default value set to "calico,calico-ipam,flannel,loopback,host-local" to keep backward compatibility.

/cc: @yifan-gu

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -58,7 +58,7 @@ if [ -w "/host/opt/cni/bin/" ]; then
 	do
 		filename=$(basename $path)
 		tmp=",$filename,"
-		if [ ! "${SKIP_CNI_BINARIES#*$tmp}" != "$SKIP_CNI_BINARIES" ] && [ ! -f /host/opt/cni/bin/$filename ]; then
+		if [ "${SKIP_CNI_BINARIES#*$tmp}" = "$SKIP_CNI_BINARIES" ] && [ ! -f /host/opt/cni/bin/$filename ]; then
 			cp /opt/cni/bin/$filename /host/opt/cni/bin/
 		fi
 	done
@@ -73,7 +73,7 @@ if [ -w "/host/secondary-bin-dir/" ]; then
 	do
 		filename=$(basename $path)
 		tmp=",$filename,"
-		if [ ! "${SKIP_CNI_BINARIES#*$tmp}" != "$SKIP_CNI_BINARIES" ] && [ ! -f /host/opt/cni/bin/$filename ]; then
+		if [ "${SKIP_CNI_BINARIES#*$tmp}" = "$SKIP_CNI_BINARIES" ] && [ ! -f /host/opt/cni/bin/$filename ]; then
 			cp /opt/cni/bin/$filename /host/secondary-bin-dir/
 		fi
 	done
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -49,8 +49,8 @@ then
 fi
 
 # Choose which default cni binaries should be copied
-INSTALL_CNI_BINARIES=${INSTALL_CNI_BINARIES:-calico,calico-ipam,flannel,loopback,host-local}
-INSTALL_CNI_BINARIES=",$INSTALL_CNI_BINARIES,"
+SKIP_CNI_BINARIES=${SKIP_CNI_BINARIES:-""}
+SKIP_CNI_BINARIES=",$SKIP_CNI_BINARIES,"
 
 # Place the new binaries if the directory is writeable.
 if [ -w "/host/opt/cni/bin/" ]; then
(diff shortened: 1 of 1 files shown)


tmjd:
I'm thinking this new env variable should specify the binaries to *not* install.  In the future when new binaries are added or current ones renamed the current proposed list will mean that any upgrades will result in a broken setup.  If it was a list of binaries to exclude then new ones can be added without the need to update the env var.
@caseydavenport WDYT?

abhinavdahiya:
@tmjd 
I think `INSTALL_CNI_BINARIES` allows users to set it to `calico,calico-ipam` and not worry about other binaries. If we use something like `SKIP_CNI_BINARIES` users who want just `calico,calico-ipam` will have to specify list of all other binaries to skip which might keep on changing.

tmjd:
I understand that, but I'm thinking if calico or calico-ipam were renamed or began to depend on a 3rd binary then a setup with INSTALL_CNI_BINARIES would break on upgrade.

If a new binary was added to the installer that was not needed in some setups, then not having it listed in SKIP_CNI_BINARIES would, I think, more than likely not break the setup.

I'm asking the question, which configuration is more likely to break on upgrade?

Or maybe this setting is a specific corner case, and that it breaking during an upgrade is too small to be concerned with.

yifan-gu:
@tmjd Good point. The use case we have is **not to override** some binaries provided by 3rdparty. So `SKIP_CNI_BINARIES` SGTM.

abhinavdahiya:
@caseydavenport @tmjd @yifan-gu If everybody agrees that `SKIP_CNI_BINARIES` is more useful. I'll update the PR.

abhinavdahiya:
@yifan-gu @tmjd @caseydavenport As every one suggested that SKIP_CNI_BINARIES would be a better solution. I have pushed changes to change `INSTALL_CNI_BINARIES` flag to `SKIP_CNI_BINARIES`

k8s-install/scripts/install-cni.sh
```diff
@@ -48,45 +48,35 @@ then
...
-	    cp /opt/cni/bin/host-local /host/opt/cni/bin/
-	fi
-	if [ ! -f /host/opt/cni/bin/portmap ]; then
-	    cp /opt/cni/bin/portmap /host/opt/cni/bin/
-	fi
+	for path in /opt/cni/bin/*
+	do
+		filename=$(basename $path)
+		tmp=",$filename,"
+		if [ ! "${SKIP_CNI_BINARIES#*$tmp}" != "$SKIP_CNI_BINARIES" ] && [ ! -f /host/opt/cni/bin/$filename ]; then
```
tmjd: Is there a reason for the double negative? `[ ! x != y ]`  If not please switch to equality check, otherwise please add a comment.

k8s-install/scripts/install-cni.sh
```diff
@@ -48,45 +48,35 @@ then
...
-	    cp /opt/cni/bin/host-local /host/secondary-bin-dir/
-	fi
-	if [ ! -f /host/secondary-bin-dir/portmap ]; then
-	    cp /opt/cni/bin/portmap /host/secondary-bin-dir/
-	fi
+	for path in /opt/cni/bin/*
+	do
+		filename=$(basename $path)
+		tmp=",$filename,"
+		if [ ! "${SKIP_CNI_BINARIES#*$tmp}" != "$SKIP_CNI_BINARIES" ] && [ ! -f /host/opt/cni/bin/$filename ]; then
```
tmjd: double negative

abhinavdahiya: @tmjd changed the double negative to equality check. Thanks!

tmjd:
@abhinavdahiya Could you describe the testing you have done for this change?

tmjd:
I pulled down the changes and built a new calico/cni container and then did two runs with it.  The first with no SKIP_CNI_BINARIES value set and the second run with `portmap` set.  Both runs worked as expected and when I inspected `/opt/cni/bin` for the 2nd run the portmap binary was removed.

abhinavdahiya:
I ran following test-cases locally:
-1:
`SKIP_CNI_BINARIES=""`
`/opt/cni/bin = calico,calico-ipam,flannel,host-local,loopback,portmap`

-2:
`SKIP_CNI_BINARIES="flannel"`
`/opt/cni/bin = calico,calico-ipam,host-local,loopback,portmap`

-3:
`SKIP_CNI_BINARIES="flannel,portmap"`
`/opt/cni/bin = calico,calico-ipam,host-local,loopback`

-4:
`SKIP_CNI_BINARIES="flannel,non-existing-binary"`
`/opt/cni/bin = calico,calico-ipam,host-local,loopback,portmap`

-5:
`SKIP_CNI_BINARIES="flannel,,"`
`/opt/cni/bin = calico,calico-ipam,host-local,loopback,portmap`

PR #337, opened 2017-05-19
Add IPAM config, annotation tests

gunjan5:

PR #346, opened 2017-06-14
Update glide.yaml pins

caseydavenport:
## Description

- Update pins in glide.yaml to track latest minor release of each dependency.
- Pin client-go to known working commit from libcalico-go


gunjan5:
you will also have to update the tests, see: https://github.com/projectcalico/cni-plugin/pull/337/commits/7e8497fa1f342d99e70f379c73241849f13021e9

gunjan5:
so with the same failing tests, this PR seems a subset of https://github.com/projectcalico/cni-plugin/pull/337/files

PR #356, opened 2017-07-17
Overwrite CNI binaries by default

alvelcom:
`install-cni.sh` doesn't overwrites binaries by default. That creates troubles during upgrade, because I had to remove stuff in `/opt/cni/bin` and then restart a calico/canal pod. Or just put binaries by hand.

Thes PR adds a new environment variable `UPDATE_CNI_BINARIES` which is `true` by default. If `UPDATE_CNI_BINARIES` is `true`, CNI binaries would be overwritten. If it's `false`, existing binaries would stay, like it was before the patch.

## Release Note
```release-note
install-cni.sh will now replace system CNI binaries, use UPDATE_CNI_BINARIES=false to restore old behavior 
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -51,35 +51,36 @@ fi
 # Choose which default cni binaries should be copied
 SKIP_CNI_BINARIES=${SKIP_CNI_BINARIES:-""}
 SKIP_CNI_BINARIES=",$SKIP_CNI_BINARIES,"
+UPDATE_CNI_BINARIES=${UPDATE_CNI_BINARIES:-"true"}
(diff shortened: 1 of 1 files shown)


alvelcom:
Hey! Anything I can help with?

tmjd:
I don't think so, I think @caseydavenport has just been busy.  He's OOO for for a little bit so I just wanted to let you know why you weren't going to see a response for a bit.

I kicked off a rebuild of the semaphoreci test as the failure was from a time out, which they do occasionally.

alvelcom:
Okay, thanks!

tmjd:
LGTM
@caseydavenport WDYT?

PR #361, opened 2017-07-27
Mask k8s auth token from install-cni scrip logs.

### LINKED ISSUE #360: Output from install-cni.sh prints k8s_auth_token field in plaintext to logs
## Expected Behavior

k8s_auth_token field should be masked.

## Current Behavior

k8s_auth_token field is printed in plain text, along with the rest of 10-calico.conf.

## Possible Solution

Mask that field away by piping to sed before using echo.

## Context

In our cluster we log all container logs from `/var/logs/container/*` which then goes into a centralised logging platform. Logging the service account token like this allows anyone with access to the log platform to impersonate the calico service account in our cluster which is an undesirable outcome for cluster security. In some clusters this will be the kube-system service account which has a lot of privileges. Thankfully in our cluster we already made a separate account with limited permissions.

## Your Environment
calico/cni:1.9.1

gunjan5:
Fixes #360 

It looks like this with this PR:

```
Wrote CNI config: {
    "name": "k8s-pod-network",
    "cniVersion": "0.1.0",
    "type": "calico",
    "etcd_endpoints": "http://10.96.232.136:6666",
    "log_level": "info",
    "mtu": 1500,
    "ipam": {
        "type": "calico-ipam"
    },
    "policy": {
        "type": "k8s",
         "k8s_api_root": "https://10.96.0.1:443",
         "k8s_auth_token": "*****"
    },
    "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
    }
}
```

### Release Note:
```release-note
NONE
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -117,7 +117,6 @@ SERVICEACCOUNT_TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
 grep "__KUBERNETES_SERVICE_HOST__" $TMP_CONF && sed -i s/__KUBERNETES_SERVICE_HOST__/${KUBERNETES_SERVICE_HOST}/g $TMP_CONF
 grep "__KUBERNETES_SERVICE_PORT__" $TMP_CONF && sed -i s/__KUBERNETES_SERVICE_PORT__/${KUBERNETES_SERVICE_PORT}/g $TMP_CONF
 sed -i s/__KUBERNETES_NODE_NAME__/${KUBERNETES_NODE_NAME:-$(hostname)}/g $TMP_CONF
-sed -i s/__SERVICEACCOUNT_TOKEN__/${SERVICEACCOUNT_TOKEN:-}/g $TMP_CONF
 sed -i s/__KUBECONFIG_FILENAME__/calico-kubeconfig/g $TMP_CONF
 
 # Use alternative command character "~", since these include a "/".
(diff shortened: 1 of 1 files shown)


k8s-install/scripts/install-cni.sh
```diff
@@ -129,7 +129,7 @@ sed -i s~__ETCD_ENDPOINTS__~${ETCD_ENDPOINTS:-}~g $TMP_CONF
 # Move the temporary CNI config into place.
 FILENAME=${CNI_CONF_NAME:-10-calico.conf}
 mv $TMP_CONF /host/etc/cni/net.d/${FILENAME}
-echo "Wrote CNI config: $(cat /host/etc/cni/net.d/${FILENAME})"
+echo "Wrote CNI config: $(cat /host/etc/cni/net.d/${FILENAME} | sed '/k8s_auth_token/s/\: \"[a-zA-Z0-9\_\-\.\-]*/\: \"\*\*\*\*\*/g')"
```
tmjd: You could simplify the sed a bit with `sed '/k8s_auth_token/s/\: \"[^"]*/\: \"*****/g'`
Might as well just match on things that are not `"` and in the replacement half you don't have to escape the `*`.

tmjd:
LGTM

caseydavenport:
Sorry for the late review, but I'd like to suggest a different approach.

What if we move the substitution until _after_ we log out the config?

Basically, this line: https://github.com/gunjan5/calico-cni/blob/a4ecd6a89be90a02906e842f75c12da4bdc5f8f5/k8s-install/scripts/install-cni.sh#L119

That way, no matter where in the file it is, we'll log the right thing.

caseydavenport:
The logic would be:


- log the TMP file
- sed the token in
- copy it over

gunjan5:
@caseydavenport updated the PR. Tested it, and it looks like this now:
```
$ kl calico-node-h58nq -c install-cni
CNI plugin version: v1.10.0-22-gff88441
         "k8s_api_root": "https://10.96.0.1:__KUBERNETES_SERVICE_PORT__",
CNI config: {
    "name": "k8s-pod-network",
    "cniVersion": "0.1.0",
    "type": "calico",
    "etcd_endpoints": "http://10.96.232.136:6666",
    "log_level": "info",
    "mtu": 1500,
    "ipam": {
        "type": "calico-ipam"
    },
    "policy": {
        "type": "k8s",
         "k8s_api_root": "https://10.96.0.1:443",
         "k8s_auth_token": "__SERVICEACCOUNT_TOKEN__"
    },
    "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
    }
}
Created CNI config 10-calico.conf
```
PTAL

PR #366, opened 2017-08-01
Log tests to ginkgowriter

ozdanborne:
## Description

In current master, a successful test run through with no errors will spit out random log messages from various tests.

This PR logs to the ginkgowriter stream which hides output unless the test fails.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -33,6 +33,7 @@ func init() {
 	rand.Seed(time.Now().UTC().UnixNano())
 	log.SetFormatter(&logutils.Formatter{})
 	log.AddHook(&logutils.ContextHook{})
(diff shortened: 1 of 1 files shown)


gunjan5:
also, have you run the tests with `ginkgo -v` to make sure the logs show up even if the tests are not failing?

ozdanborne:
I've tested that logs do show up when `-v` is passed or the test fails.

I am hesitant to touch the logging in other files, as they seem like they need some larger changes which I'm not confidant in. Basically, it'd seem like they should use `Expect(err)` instead of `if err; log.Fatalf`. 

I'll get to it eventually but I want to get a better grasp on the best practice first.

PR #367, opened 2017-08-04
Added dynamic log level

zopanix:
These changes give the user the possibility to change calico's log level.
The log level can be changed by launching the container with the `LOG_LEVEL` environment variable containing the desired level.

This commit affects the calico.conf.default template to add a placeholder for the log level. It changes the install-cni.sh script and add the sed command to add the log level.

Finally it also provides a default value for the log-level as an env variable in the Dockerfile.

## Todos
- [x] Tests: It tested it locally and the templating stuff works
- [x] Documentation: I don't think documentation is needed, if someone think there should be, please advice me
- [x] Release note: See below

## Changelog
```release-note
#367 - Added configurable log level: Changes default behavior of the container by setting the log level to warn by default. To override this you can set it to info by launching the container with as environment variable: `LOG_LEVEL` to `info` (or whatever other valid level you want)
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -125,7 +125,7 @@ sed -i s~__ETCD_CERT_FILE__~${CNI_CONF_ETCD_CERT:-}~g $TMP_CONF
 sed -i s~__ETCD_KEY_FILE__~${CNI_CONF_ETCD_KEY:-}~g $TMP_CONF
 sed -i s~__ETCD_CA_CERT_FILE__~${CNI_CONF_ETCD_CA:-}~g $TMP_CONF
 sed -i s~__ETCD_ENDPOINTS__~${ETCD_ENDPOINTS:-}~g $TMP_CONF
-sed -i s~__LOG_LEVEL__~${LOG_LEVEL}~g $TMP_CONF
+sed -i s~__LOG_LEVEL__~${LOG_LEVEL:-warn}~g $TMP_CONF
 
 # Move the temporary CNI config into place.
 FILENAME=${CNI_CONF_NAME:-10-calico.conf}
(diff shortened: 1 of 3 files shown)


tmjd:
Thank you for the contribution.
I think the addition to be able to change the log level is great.  Since this does include an end user change, the default going from `info` to `warn`, I think we should add a release note or keep the default.  @caseydavenport WDYT?

zopanix:
hey @caseydavenport,
Sorry about that warn thing. I put the defaults I wanted. I don't care what the choice will be since I can override it. But calico is pretty stable on my side So I decided to put it into warn. But if we do change default behovior, we'll need to put into the change log.

zopanix:
Just added the changelog if you guys decide to stick with warn, if not I'll update the pull request to info. thx

k8s-install/scripts/install-cni.sh
```diff
@@ -125,6 +125,7 @@ sed -i s~__ETCD_CERT_FILE__~${CNI_CONF_ETCD_CERT:-}~g $TMP_CONF
 sed -i s~__ETCD_KEY_FILE__~${CNI_CONF_ETCD_KEY:-}~g $TMP_CONF
 sed -i s~__ETCD_CA_CERT_FILE__~${CNI_CONF_ETCD_CA:-}~g $TMP_CONF
 sed -i s~__ETCD_ENDPOINTS__~${ETCD_ENDPOINTS:-}~g $TMP_CONF
+sed -i s~__LOG_LEVEL__~${LOG_LEVEL}~g $TMP_CONF
```
caseydavenport: I think the defaulting should be done here rather than in the Dockerfile

e.g.

```
${LOG_LEVEL:-warn}
```

caseydavenport:
This LGTM with one change I think we should make to the defaulting.  Thanks!

k8s-install/scripts/install-cni.sh
zopanix: Yeah, I understand why you would say that, I intentionnally put it in the Dockerfile because of how I use docker images in general: ervery time I use a docker image that wasn't made by made, I check the Docker file to see which port it exposes and which env variables are available to me to configure at run time. That's the reason I do this but I can change this if you want to :)
Thanks

zopanix:
@caseydavenport @tmjd : I moved the default value in the build script and out of the Dockerfile. If you want me to change anything else, just tell me. Thx for being so quick

zopanix:
@caseydavenport @tmjd Any news on this ? Thx

tmjd:
Changes look good to me but @caseydavenport needs to sign off as maintainer.  I feel like the release note is mis-leading, the word "dynamic" makes me think the log level changes during operation.  What about `Added configurable log level` instead as the tag line? or `Exposed log level through environment variable`?

zopanix:
@tmjd : Looks fine to me for `Added configurable ...` it is more precise.

caseydavenport:
This LGTM, sorry for the long delay on this!

PR #371, opened 2017-08-14
Add arm64 architecture support for calico cni-plugin(issue #14)

TrevorTaoARM:
Add support for arm64 architecture to Makefile so that cni-plugin
can be ran on arm64 server. The modifications include
1. Using a 'ARCH' variable to differentiate between amd64/arm64;
2. Using the docker images of arm64 of go-build, etcd, hyperkube;
3. Using cni-plugins of arm64, which contains portmap;
4. Modifying the startup command of the various used docker images
   to suite for different ARCH , such as hyperkube, go-build,
   etcd.

this commit resolves issue #14

TrevorTaoARM <trevor.tao@arm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,14 @@ LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
 # fail if unable to download
 CURL=curl -sSf
 
+ARCH?=amd64
+
 K8S_VERSION=1.6.1
 CNI_VERSION=v0.5.2
+
+ifeq ($(ARCH), arm64) 
+CNI_VERSION=v0.6.0-rc1
+endif
 
 CALICO_CNI_VERSION?=$(shell git describe --tags --dirty)
 
@@ -20,6 +26,15 @@ CNI_SPEC_VERSION?=0.3.1
 # Ensure that the dist directory is always created
 MAKE_SURE_DIST_EXIST := $(shell mkdir -p dist)
 CALICO_BUILD?=calico/go-build
+ETCD_IMG?=quay.io/coreos/etcd
+
+# Set docker image for arm64
+ifeq ($(ARCH), arm64)
+CALICO_BUILD=linaro/ci-aarch64-leg-go-build-alpine
+ETCD_IMG=linaro/ci-aarch64-leg-etcd
+endif
+
+
 DEPLOY_CONTAINER_NAME=calico/cni
 DEPLOY_CONTAINER_MARKER=cni_deploy_container.created
 
@@ -111,6 +126,7 @@ fetch-cni-bins: dist/flannel dist/loopback dist/host-local dist/portmap
 
 # For now we've got a checked in version of the plugin.  Eventually
 # this should pull from upstream releases.
+ifeq ($(ARCH), amd64) 
 dist/portmap:
 	mkdir -p dist
 	$(CURL) -L https://github.com/projectcalico/cni-plugin/releases/download/v1.9.0/portmap -o $@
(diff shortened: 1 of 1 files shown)


TrevorTaoARM:
I don't know the root cause for the checks failure as it is my first commit to Calico org,  would you please provide some help for me?

tmjd:
I think the failure was a timeout issue that happens sometimes with Semaphore, I kicked off a rebuild so we'll see if it happens again.

TrevorTaoARM:
@tmjd Thanks a lot for your help

Makefile
```diff
@@ -111,14 +126,23 @@ fetch-cni-bins: dist/flannel dist/loopback dist/host-local dist/portmap
...
 	$(CURL) -L https://github.com/projectcalico/cni-plugin/releases/download/v1.9.0/portmap -o $@
 	chmod +x dist/portmap
 
 dist/flannel dist/loopback dist/host-local:
 	mkdir -p dist
-	$(CURL) -L --retry 5 https://github.com/containernetworking/cni/releases/download/$(CNI_VERSION)/cni-amd64-$(CNI_VERSION).tgz | tar -xz -C dist ./flannel ./loopback ./host-local
+	$(CURL) -L --retry 5 https://github.com/containernetworking/cni/releases/download/$(CNI_VERSION)/cni-$(ARCH)-$(CNI_VERSION).tgz | tar -xz -C dist ./flannel ./loopback ./host-local
+endif
+
+ifeq ($(ARCH), arm64)
```
TrevorTaoARM: @caseydavenport Thanks a lot for your review.
Yes, I also think if amd64 and arm64 can use a single version of cni-plugins, which would simplify the work. Actually I have written a version of Makefile change with the same CNI_VERSION=v0.6.0-rc1 and it works, but for the concern of keeping x86 as the original as possible, I separate the processings for x86 and arm64. 
If you think a single version is better and given a certain version, I would merge the separate processings of the 2 archs.

Btw, I would also update my arm64 enabling work on other Calico components in the near future.  
Hope you can give me more help.

tmjd:
You can find myself and casey in Slack at slack.projectcalico.org.  There is a #calico-dev channel that is probably the most appropriate place to chat about porting Calico.

caseydavenport:
@TrevorTaoARM sorry for the delay! I've been out of the office for a couple of weeks.

I'm still working through my backlog of emails, messages, etc.

Makefile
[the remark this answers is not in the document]
caseydavenport: > Actually I have written a version of Makefile change with the same CNI_VERSION=v0.6.0-rc1 and it works, but for the concern of keeping x86 as the original as possible... If you think a single version is better and given a certain version, I would merge the separate processings of the 2 archs

Yeah! I'd much rather combine the two into a single code path if that works.  Much cleaner.

Also, since I've taken such a long time to finish reviewing this... https://github.com/containernetworking/plugins/releases/tag/v0.6.0

Makefile
```diff
@@ -20,6 +26,15 @@ CNI_SPEC_VERSION?=0.3.1
 # Ensure that the dist directory is always created
 MAKE_SURE_DIST_EXIST := $(shell mkdir -p dist)
 CALICO_BUILD?=calico/go-build
+ETCD_IMG?=quay.io/coreos/etcd
+
+# Set docker image for arm64
+ifeq ($(ARCH), arm64)
+CALICO_BUILD=linaro/ci-aarch64-leg-go-build-alpine
```
caseydavenport: Would you be able to submit a Dockerfile [here](https://github.com/projectcalico/go-build)? Then we could have official calico/go-build releases. 

This is also what we did for [ppc64](https://github.com/projectcalico/go-build/blob/master/Dockerfile.ppc64le)

caseydavenport:
@TrevorTaoARM sorry for the long delay!  I've added two more comments which should hopefully be easy to resolve.  Looking forward to getting this in.

Makefile
TrevorTaoARM: @caseydavenport 
No problem.  I would submit it ASAP.  
And since my company requires an internal review before any submit, I will submit it after that.

Makefile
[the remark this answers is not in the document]
TrevorTaoARM: Yes, it's great v0.6.0 is ok now, and I would also like to combine the two into a single code. 
If needed, I could submit an update for this change.

TrevorTaoARM:
@caseydavenport I have submit the Dockerfile.arm64 to projectcalico/go-build with:
https://github.com/projectcalico/go-build/pull/13

TrevorTaoARM:
@caseydavenport I have used the newest cni version v0.6.0:
https://github.com/containernetworking/cni/releases/tag/v0.6.0
for both amd64 and arm64 to simplify the work, would you please check it again.

caseydavenport:
I think the current Makefile has build support for arm64 now: https://github.com/projectcalico/cni-plugin/blob/master/Makefile#L23-L25

PR #375, opened 2017-09-07
Fixing calico-ipam using hostname instead of nodename

heschlie:
## Description

Fixes projectcalico/calico#826

It looks as though we were still using the deprecated `hostname` from the CNI payload, and not `nodename` which was causing issues assigning affine blocks if you need to use a `NODENAME` that is different from the Nodes `os.Hostname()`.

I tested this on a vagrant cluster launching Calico with systemd to configure a `NODENAME` different than the Nodes hostname, and matched the `NODENAME` env var in the CNI config files `nodename`.  Without this fix the claimed blocks get assigned to the Nodes hostname in etcd, instead of the proper nodename.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico IPAM will now respect the `nodename` variable in the CNI config. This will still check for the old value of `hostname` first if it exists, and will default to the Nodes `os.Hostname()` if neither is set.
```

### DIFF
--- a/ipam/calico-ipam.go
+++ b/ipam/calico-ipam.go
@@ -69,12 +69,25 @@ type ipamArgs struct {
 	IP net.IP `json:"ip,omitempty"`
 }
 
+func determineNodename(conf utils.NetConf) string {
+	nodename, _ := os.Hostname()
+	if conf.Hostname != "" {
+		nodename = conf.Hostname
+		log.Warn("Configuration option 'hostname' is deprecated, use 'nodename' instead.")
+	}
+	if conf.Nodename != "" {
+		nodename = conf.Nodename
+	}
+	return nodename
+}
+
 func cmdAdd(args *skel.CmdArgs) error {
 	conf := utils.NetConf{}
 	if err := json.Unmarshal(args.StdinData, &conf); err != nil {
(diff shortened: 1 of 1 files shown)


bcreane:
lgtm at this point.

ipam/calico-ipam.go
```diff
@@ -69,12 +69,26 @@ type ipamArgs struct {
 	IP net.IP `json:"ip,omitempty"`
 }
 
+func updateNodename(conf utils.NetConf) string {
```
caseydavenport: naming nit: this isn't _updating_ nodename, it's determining it.

Maybe `determineNodename` would be better.

ipam/calico-ipam.go
```diff
@@ -69,12 +69,26 @@ type ipamArgs struct {
...
+}
+
 func cmdAdd(args *skel.CmdArgs) error {
 	conf := utils.NetConf{}
 	if err := json.Unmarshal(args.StdinData, &conf); err != nil {
 		return fmt.Errorf("failed to load netconf: %v", err)
 	}
 
+	nodename := updateNodename(conf)
+
```
caseydavenport: no need for the extra newline.

PR #380, opened 2017-09-15
Enable necessary sysctls in the container and host net namespace for IPv6

gunjan5:
## Description

Add support for IPv6. This PR should make Calico work with kubernetes in IPv6 

## Todos
 - need to add some tests 

## Release Note

```release-note
Fixed bug preventing startup of kubernetes pods with IPv6 addresses
```


gunjan5:
Fixes #345

gunjan5:
@caseydavenport updated the PR as per our discussion. PTAL

utils/network.go
```diff
@@ -109,12 +109,27 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
 
 			// Handle IPv6 routes
 			if addr.Version == "6" {
+				// Make sure ipv6 is enabled in the container/pod network namespace.
+				// Without these sysctls enabled, interfaces will come up but they won't get a link local IPv6 address
+				// which is required to add the default IPv6 route.
+				if err = writeProcSys("/proc/sys/net/ipv6/conf/all/disable_ipv6", "0"); err != nil {
+					return err
```
caseydavenport: Will this error be descriptive enough?

I think we should wrap all of these so there's more context. 

```
return fmt.Errorf("failed to set conf/all/disable_ipv6=0: %s", err)
```

utils/network.go
```diff
@@ -258,6 +273,12 @@ func configureSysctls(hostVethName string, hasIPv4, hasIPv6 bool) error {
 	}
 
 	if hasIPv6 {
+		// Make sure ipv6 is enabled on the hostVeth interface in the host network namespace.
+		// Interfaces won't get a linc local address without this sysctl set to 0.
```
caseydavenport: typo: linc -> link

utils/network.go
gunjan5: I've updated all the sysctl errors. I used the syntax used for working on sysctl commands so it's more user-friendly.

calico_cni_test.go
```diff
@@ -390,63 +390,4 @@ var _ = Describe("CalicoCni", func() {
...
-			  }
-			}`, cniVersion, os.Getenv("ETCD_IP"))
-
-			It("should successfully add and network the container", func() {
-				containerIPStr := "10.0.0.55"
-				_, containerIP, _ := net.ParseCIDR("10.0.0.55/32")
-
-				By("Manually programming the route that CNI plugin is going to add.")
-				netlink.RouteAdd(
-					&netlink.Route{
```
gunjan5: @caseydavenport I'm removing this whole test (unrelated to this PR) because you can't add a route without the interface being present first which is why this test was failing. There's no easy way to reproduce the interface and route being present before we call CNI ADD, so I'm removing this test for now

PR #381, opened 2017-09-19
Disable IP forwarding inside the container netns.

Dirbaio:
## Description

This PR disables ip forwarding inside the container netns. By default it is enabled, because it's inherited from the host netns.

My use case for this is running containers that accept network traffic for analysis.

I'm running a container with Snort IDS in it. I have some iptables rules on the host that duplicate the traffic with `TEE` to the container. The container is seeing the traffic is not destined to it, and is routing it back out the container interface. This then causes martian source errors in the host, and is pointless overall.

I think there are pretty much no downsides to disabling this: Containers only have 1 interface, so only difference this can make is that incoming packets can no longer be routed out the single interface. I really can't think of a reason you'd want this.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Disable IP forwarding inside the container netns.
```

### DIFF
--- a/utils/network.go
+++ b/utils/network.go
@@ -142,15 +142,14 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
 			}
 		}
 
+		if err = configureContainerSysctls(hasIPv4, hasIPv6); err != nil {
+			return fmt.Errorf("error configuring sysctls for the container netns, error: %s", err)
+		}
+
 		// Now that the everything has been successfully set up in the container, move the "host" end of the
 		// veth into the host namespace.
 		if err = netlink.LinkSetNsFd(hostVeth, int(hostNS.Fd())); err != nil {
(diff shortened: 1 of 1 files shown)


utils/network.go
```diff
@@ -148,6 +148,11 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
 			return fmt.Errorf("failed to move veth to host netns: %v", err)
 		}
 
+		err = configureContainerSysctls(hasIPv4, hasIPv6)
```
fasaxc: I think this should go above the block above that moves the host end of the veth.  I think we want to make sure it's configured correctly before exposing it.

caseydavenport:
@gunjan5 you're also doing some sysctl work in this area could you take a quick look?

albertvaka:
@fasaxc comments fixed. Can this be merged now?

caseydavenport:
I'm happy to merge as-is.

PR #382, opened 2017-09-19
Updating to the latest upstream CNI plugin release

### LINKED ISSUE #362: Use upstream portmap plugin releases (when available)
We currently host our own build of the portmap plugin to include in calico/cni. We should switch to the upstream release when v0.6.0 comes out.

heschlie:
## Description

Fixes #362 

Updated to CNI plugins release v0.6.0
Removed target for our fork of portmap in favor of using the version included in the CNI plugins

I tested this locally and ran the E2Es against it with no issues.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
 CURL=curl -sSf
 
 K8S_VERSION=1.6.1
-CNI_VERSION=v0.5.2
+CNI_VERSION=v0.6.0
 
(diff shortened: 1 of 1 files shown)


caseydavenport:
@heschlie this LGTM - which kubernetes versions did you test this against? v1.7?

heschlie:
v1.7.5, I could test against 1.6 and 1.5 if we want to cover the bases.

caseydavenport:
@heschlie I've merged since I think this is useful but we should validate against a v1.8 and v1.6 cluster ideally if it won't take you too long. 

v1.5 doesn't need to be tested.

PR #383, opened 2017-09-20
Always update the active instance ID

caseydavenport:
## Description

We don't always update the ActiveInstanceID on workload endpoints when handling an ADD call.

This results in a scenario like so:

1) ADD for Pod X with container ID A
2) ADD for Pod X with container ID B
3) DEL for Pod X with container ID A

Since step 2 doesn't update the active instance ID, the workload endpoint gets deleted in step 3 even though there is an actively running container / Pod for that wep.

Fixes https://github.com/projectcalico/calico/issues/1117

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Fixed a bug where running Kubernetes Pods could end up without corresponding workload endpoints in the datastore
```

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -799,6 +799,145 @@ var _ = Describe("CalicoCni", func() {
 				})
 			})
 
+			// This specific test case is for an issue where k8s would send two ADDs with different container IDs for the same Pod.
+			Context("Multiple ADD calls with different container IDs followed by DEL calls for stale container IDs", func() {
+				It("Use different CNI_ContainerIDs to ADD and DEL the container", func() {
+					netconf := fmt.Sprintf(`
+				{
+			      "cniVersion": "%s",
+				  "name": "net7",
+				  "type": "calico",
+				  "etcd_endpoints": "http://%s:2379",
+			 	  "ipam": {
+			    	 "type": "calico-ipam"
+			         },
+					"kubernetes": {
+					  "k8s_api_root": "http://127.0.0.1:8080"
+					 },
+					"policy": {"type": "k8s"},
+					"log_level":"info"
+				}`, cniVersion, os.Getenv("ETCD_IP"))
+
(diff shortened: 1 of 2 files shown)


calico_cni_k8s_test.go
```diff
@@ -799,6 +799,161 @@ var _ = Describe("CalicoCni", func() {
...
+
+					// Delete the container with the CNI_CONTAINERID "X".
+					exitCode, err := DeleteContainerWithId(netconf, contNs.Path(), name, cniContainerIDX)
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(exitCode).Should(Equal(0))
+
+					// The endpoint should still exist in the backend datastore, but with container ID Y.
+					endpoints, err = calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{})
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(endpoints.Items).Should(HaveLen(0))
```
gunjan5: i don't expect this to be true. Seems like Line 912 to 915 shouln't be there

caseydavenport: Yep, good spot :)

calico_cni_k8s_test.go
```diff
@@ -799,6 +799,161 @@ var _ = Describe("CalicoCni", func() {
...
+					Expect(endpoints.Items[0].Metadata).Should(Equal(api.WorkloadEndpointMetadata{
+						Node:             hostname,
+						Name:             "eth0",
+						Workload:         fmt.Sprintf("test.%s", name),
+						ActiveInstanceID: cniContainerIDY,
+						Orchestrator:     "k8s",
+						Labels:           map[string]string{"calico/k8s_ns": "test"},
+					}))
+
+					// Delete the container with the CNI_ContainerID "X" again.
```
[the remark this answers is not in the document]
caseydavenport: Yep, agreed.

PR #386, opened 2017-09-28
Adding tests for install-cni.sh

### LINKED ISSUE #374: CNI installation returns successful error code even when it fails miserably
I ran a hosted install that failed to install cni-plugins due to permission errors from the host. Calico thought everything succeeded even though it didn't.

## Expected Behavior

I'd expect the cni container to fail and exit with an appropriate status, and kubernetes should take over repeatedly retrying it for the user to notice in the pod logs.

## Current Behavior

container thought it succeeded and my installation entered the following state:
- all calico pods running / successful
- nodes in NotReady state

## Possible Solution

cni install script should return the right error code
## Steps to Reproduce (for bugs)

This is more easily reproducable in OpenShift where containers are locked down. But any situation which causes an error in the install script should reproduce this.
1.
2.
3.
4.

## Context

```
$ oc logs -n kube-system calico-node-l1j9p -c install-cni
cp: can't create '/host/opt/cni/bin/calico': Permission denied
cp: can't create '/host/opt/cni/bin/calico-ipam': Permission denied
cp: can't create '/host/opt/cni/bin/flannel': Permission denied
cp: can't create '/host/opt/cni/bin/host-local': Permission denied
cp: can't create '/host/opt/cni/bin/loopback': Permission denied
cp: can't create '/host/opt/cni/bin/portmap': Permission denied
CNI plugin version: 
        "k8s_api_root": "https://172.30.0.1:__KUBERNETES_SERVICE_PORT__",
/install-cni.sh: line 67: /host/opt/cni/bin/calico: not found
/install-cni.sh: line 97: can't create /host/etc/cni/net.d/calico-kubeconfig: Permission denied
mv: can't create '/host/etc/cni/net.d/10-calico.conf': Permission denied
cat: can't open '/host/etc/cni/net.d/10-calico.conf': No such file or directory
Wrote CNI config: 
```

### LINKED ISSUE #376: Automated tests for CNI install script
See here: https://github.com/projectcalico/cni-plugin/pull/356#issuecomment-327508619

We don't have any UT/FV coverage of the install CNI script!  We should.

heschlie:
## Description

Fixes #374 
Fixes #376 
I've added several tests to ensure at least basic functionality of the install-cni.sh script. I've also resolved a couple of minor bugs in the script for it not failing when it should.

I still have some minor concerns, I was unable to write a test to check that the script fails when we are unable to write the CNI bin files as we do not have `chattr` in the build container, and since we are running as root we have the capability to clobber anything otherwise, looking for suggestions here.

I am also unsure of this test:
https://github.com/projectcalico/cni-plugin/commit/5d1d18f2bd2ba1bdd0691c2e2cbd52f99b304eca#diff-2cac8d2716f994aa4216d877a9d2c84aR120
In the situation where both the primary and secondary bin dirs are not writable, it feels like we should fail, which is simple enough to fix but I'm unsure if this is set this way for a reason other than checking it exists.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ delstats
 *.pyc
 *.created
 tmp-cni
+k8s-install/scripts/tmp/
+install_cni.test
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ docker-image: $(DEPLOY_CONTAINER_MARKER)
 
 .PHONY: clean
 clean:
-	rm -rf dist vendor $(DEPLOY_CONTAINER_MARKER) .go-pkg-cache
+	rm -rf dist vendor $(DEPLOY_CONTAINER_MARKER) .go-pkg-cache k8s-install/scripts/install_cni.test
 
 release: clean
 ifndef VERSION
@@ -92,7 +92,7 @@ dist/calico-ipam: $(SRCFILES) vendor
 
 .PHONY: test
 ## Run the unit tests.
-test: dist/calico dist/calico-ipam dist/host-local run-etcd run-k8s-apiserver
+test: dist/calico dist/calico-ipam dist/host-local run-etcd run-k8s-apiserver test-install-cni
 	# The tests need to run as root
 	sudo CGO_ENABLED=0 ETCD_IP=127.0.0.1 PLUGIN=calico GOPATH=$(GOPATH) $(shell which ginkgo)
 
@@ -129,6 +129,22 @@ test-containerized: run-etcd run-k8s-apiserver build-containerized dist/host-loc
 			ginkgo'
 	make stop-etcd
 
+# We pre-build the test binary so that we can run it outside a container and allow it
+# to interact with docker.
+k8s-install/scripts/install_cni.test: vendor
+	-mkdir -p .go-pkg-cache
+	docker run --rm \
+	-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
+	-v $(CURDIR):/go/src/github.com/projectcalico/cni-plugin:rw \
+	-v $(CURDIR)/.go-pkg-cache:/go/pkg/:rw \
+		$(CALICO_BUILD) sh -c '\
+			cd /go/src/github.com/projectcalico/cni-plugin && \
+			go test ./k8s-install/scripts -c --tags install_cni_test -o ./k8s-install/scripts/install_cni.test'
+
+.PHONY: test-install-cni
+## Test the install-cni.sh script
+test-install-cni: docker-image k8s-install/scripts/install_cni.test
+	cd k8s-install/scripts && ./install_cni.test
 
 run-test-containerized-without-building: run-etcd run-k8s-apiserver
 	docker run --rm --privileged --net=host \
@@ -163,7 +179,7 @@ build-containerized: vendor
 		$(CALICO_BUILD) sh -c '\
 			cd /go/src/github.com/projectcalico/cni-plugin && \
 			make binary'
-	
+
 ## Etcd is used by the tests
 run-etcd: stop-etcd
 	docker run --detach \
@@ -294,7 +310,7 @@ run-kube-proxy:
 	-docker rm -f calico-kube-proxy
 	docker run --name calico-kube-proxy -d --net=host --privileged gcr.io/google_containers/hyperkube:v$(K8S_VERSION) /hyperkube proxy --master=http://127.0.0.1:8080 --v=2
 
-ci: clean static-checks test-containerized-cni-versions docker-image
+ci: clean static-checks test-containerized-cni-versions docker-image test-install-cni
 
 .PHONY: help
 help: # Some kind of magic from https://gist.github.com/rcmachado/af3db315e31383502660
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -5,8 +5,8 @@
 # - Expects the host CNI network config path to be mounted at /host/etc/cni/net.d.
 # - Expects the desired CNI config in the CNI_NETWORK_CONFIG env variable.
 
-# Ensure all variables are defined.
-set -u
+# Ensure all variables are defined, and that the script fails when an error is hit.
+set -u -e
 
 # The directory on the host where CNI networks are installed. Defaults to
 # /etc/cni/net.d, but can be overridden by setting CNI_NET_DIR.  This is used
--- a/k8s-install/scripts/install_cni_test.go
+++ b/k8s-install/scripts/install_cni_test.go
@@ -1,0 +1,204 @@
+package scripts_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/projectcalico/felix/fv/containers"
+
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+)
+
+var secretFile = "tmp/serviceaccount/token"
+
+// runCniContainer will run install-cni.sh.
+// TODO: We should be returning an error if the command fails, there currently is
+// not a way to get that from the container package.
+func runCniContainer(extraArgs ...string) error {
+	// Get the CWD for mounting directories into container.
+	cwd, err := os.Getwd()
+	if err != nil {
+		Fail("Could not get CWD")
+	}
+
+	// Assemble our arguments.
+	args := []string{
+		"-e", "SLEEP=false",
+		"-e", "KUBERNETES_SERVICE_HOST=127.0.0.1",
+		"-e", "KUBERNETES_SERVICE_PORT=8080",
+		"-v", cwd + "/tmp/bin:/host/opt/cni/bin",
+		"-v", cwd + "/tmp/net.d:/host/etc/cni/net.d",
+		"-v", cwd + "/tmp/serviceaccount:/var/run/secrets/kubernetes.io/serviceaccount",
+	}
+	args = append(args, extraArgs...)
+	args = append(args, "calico/cni:latest", "/install-cni.sh")
+
+	c := containers.Run("cni", args...)
+	Eventually(func() bool { return c.Stopped() }, 5*time.Second, 100*time.Millisecond).Should(BeTrue())
+
+	return nil
+}
+
+// cleanup uses the calico/cni container to cleanup after itself as it creates
+// things as root.
+func cleanup() {
+	cwd, err := os.Getwd()
+	if err != nil {
+		Fail("Could not get CWD")
+	}
+	containers.Run("cni",
+		"-e", "SLEEP=false",
+		"-e", "KUBERNETES_SERVICE_HOST=127.0.0.1",
+		"-e", "KUBERNETES_SERVICE_PORT=8080",
+		"-v", cwd+"/tmp/bin:/host/opt/cni/bin",
+		"-v", cwd+"/tmp/net.d:/host/etc/cni/net.d",
+		"-v", cwd+"/tmp/serviceaccount:/var/run/secrets/kubernetes.io/serviceaccount",
+		"calico/cni:latest",
+		"sh", "-c", "rm -f /host/opt/cni/bin/* /host/etc/cni/net.d/*")
+}
+
+var _ = BeforeSuite(func() {
+	// Make the directories we'll need for storing files.
+	err := os.MkdirAll("tmp/bin", 0755)
+	if err != nil {
+		Fail("Failed to create directory tmp/bin")
+	}
+	err = os.MkdirAll("tmp/net.d", 0755)
+	if err != nil {
+		Fail("Failed to create directory tmp/net.d")
+	}
+	err = os.MkdirAll("tmp/serviceaccount", 0755)
+	if err != nil {
+		Fail("Failed to create directory tmp/serviceaccount")
+	}
+	cleanup()
+
+	// Create a secrets file for parsing.
+	k8sSecret := []byte("my-secret-key")
+	err = ioutil.WriteFile(secretFile, k8sSecret, 0755)
+	if err != nil {
+		Fail(fmt.Sprintf("Failed to write k8s secret file: %v", err))
+	}
+})
+
+var _ = AfterSuite(func() {
+	err := os.RemoveAll("tmp")
+	if err != nil {
+		fmt.Println("Failed to clean up directories")
+	}
+})
+
+var _ = Describe("install-cni.sh tests", func() {
+	AfterEach(func() {
+		cleanup()
+	})
+
+	Describe("Run install-cni", func() {
+		Context("With default values", func() {
+			It("Should install bins and config", func() {
+				err := runCniContainer()
+				Expect(err).NotTo(HaveOccurred())
+
+				// Get a list of files in the defualt CNI bin location.
+				files, err := ioutil.ReadDir("tmp/bin")
+				if err != nil {
+					Fail("Could not list the files in tmp/bin")
+				}
+				names := []string{}
+				for _, file := range files {
+					names = append(names, file.Name())
+				}
+
+				// Get a list of files in the default location for CNI config.
+				files, err = ioutil.ReadDir("tmp/net.d")
+				if err != nil {
+					Fail("Could not list the files in tmp/net.d")
+				}
+				for _, file := range files {
+					names = append(names, file.Name())
+				}
+
+				Expect(names).To(ContainElement("calico"))
+				Expect(names).To(ContainElement("calico-ipam"))
+				Expect(names).To(ContainElement("10-calico.conf"))
+			})
+			It("Should parse and output a templated config", func() {
+				err := runCniContainer()
+				Expect(err).NotTo(HaveOccurred())
+
+				expectedFile := "expected_10-calico.conf"
+
+				var expected = []byte{}
+				if file, err := os.Open(expectedFile); err != nil {
+					Fail("Could not open " + expectedFile + " for reading")
+				} else {
+					_, err := file.Read(expected)
+					if err != nil {
+						Fail("Could not read from expected_10-calico.conf")
+					}
+					err = file.Close()
+					if err != nil {
+						fmt.Println("Failed to close expected_10-calico.conf")
+					}
+				}
+
+				var received = []byte{}
+				if file, err := os.Open("tmp/net.d/10-calico.conf"); err != nil {
+					Fail("Could not open 10-calico.conf for reading")
+				} else {
+					_, err := file.Read(received)
+					if err != nil {
+						Fail("Could not read from 10-calico.conf")
+					}
+					err = file.Close()
+					if err != nil {
+						fmt.Println("Failed to close 10-calico.conf")
+					}
+				}
+
+				Expect(expected).To(Equal(received))
+			})
+		})
+
+		Context("With modified env vars", func() {
+			It("Should rename '10-calico.conf' to '10-calico.conflist'", func() {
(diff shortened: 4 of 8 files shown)


heschlie:
Looks like it's failing some linter checks? Didn't see the `make ci` target, I've fixed them locally so `make ci` runs, and migrated some calls off of `exec.Command` for making and removing dirs.

install_cni_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"io/ioutil"
+)
+
+var script = "k8s-install/scripts/install-cni.sh"
+
+var defaultEnvVars = map[string]string {
+	"KUBERNETES_SERVICE_HOST": "127.0.0.1",
+	"KUBERNETES_SERVICE_PORT": "8080",
+	"KUBERNETES_NODE_NAME": "k8s-node-01",
+	"SERVICEACCOUNT_TOKEN": "my_service_token",
```
tmjd: I think you should remove the SERVICEACCOUNT_TOKEN since that is read from a file instead of environment variable.

install_cni_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"os/exec"
+	"io/ioutil"
+)
+
+var script = "k8s-install/scripts/install-cni.sh"
+
+var defaultEnvVars = map[string]string {
+	"KUBERNETES_SERVICE_HOST": "127.0.0.1",
+	"KUBERNETES_SERVICE_PORT": "8080",
+	"KUBERNETES_NODE_NAME": "k8s-node-01",
```
tmjd: I think  KUBERNETES_NODE_NAME should be removed since it is not used in the default template.

install_cni_test.go
```diff
@@ -0,0 +1,135 @@
...
+				os.Unsetenv("CNI_CONF_NAME")
+				Expect(err).NotTo(HaveOccurred())
+
+				// Check to see if the file was named correctly.
+				if _, err := os.Stat("/host/etc/cni/net.d/10-test.conf"); os.IsNotExist(err) {
+					Fail("Could not locate /host/etc/cni/net.d/10-test.conf")
+				}
+			})
+		})
+		Context("When it cannot write CNI bins", func() {
```
tmjd: How about a failure case when it can't write one of the binaries? (to cover the new "Failed to copy" feature you added)
I wasn't sure how you could do that but it I found that if you create a file, remove the permissions from the file then try to copy a file to it I get a permission denied failure.  To clean up you'd need to add permissions back before you could remove it.

heschlie: Did you try that as root, I did `chmod 000 calico; cp /opt/.../calico /host/../calico` and it didn't complain and happily overwrote the file. I was planning on making the file immutable but `chattr` doesn't exist in our build container.

tmjd: No I didn't :slightly_frowning_face:

heschlie: I think we'll leave this out for now, but I'll check to see if we can get `chattr` installed in the go-build container which would let us add this test.

install_cni_test.go
```diff
@@ -0,0 +1,188 @@
...
+	err := os.Setenv("SLEEP", "false")
+	if err != nil {
+		Fail("Failed to set env var 'SLEEP'")
+	}
+})
+
+var _ = Describe("install-cni.sh tests", func() {
+	BeforeEach(func() {
+		// Create the needed dirs.
+		err := os.MkdirAll("/host/opt/cni/bin", 755)
```
caseydavenport: I understand that we're almost always running this in a container so it's probably OK, but I always feel a bit icky seeing test code creating / deleting directories :)

I don't think anything needs to change here though.

Makefile
```diff
@@ -124,6 +124,7 @@ test-containerized: run-etcd run-k8s-apiserver build-containerized dist/host-loc
 	-e PLUGIN=calico \
 	-e CNI_SPEC_VERSION=$(CNI_SPEC_VERSION) \
 	-v $(CURDIR):/go/src/github.com/projectcalico/cni-plugin:rw \
+	-v $(CURDIR)/dist:/opt/cni/bin \
```
caseydavenport: Hm.

It would be really nice if somehow we could instead just run the `calico/cni` container.

That might be more than we want to bite off right now, but I think it would simplify this a bit.  For example, we wouldn't have to [hack around in the test code](https://github.com/projectcalico/cni-plugin/pull/386/files#diff-2cac8d2716f994aa4216d877a9d2c84aR56) to make it look like the `calico/cni` container filesystem.

That's a more realistic test, and means we don't need to try to keep this test code and the live code in-sync, which is bug prone.

WDYT? Can you see a way to refactor so that we run the real container? It could be its own separate target - `make test-install-cni`.

install_cni_test.go
```diff
@@ -0,0 +1,188 @@
...
+				}
+				for _, file := range files {
+					names = append(names, file.Name())
+				}
+
+				Expect(names).To(ContainElement("calico"))
+				Expect(names).To(ContainElement("calico-ipam"))
+				Expect(names).To(ContainElement("10-calico.conf"))
+			})
+			It("Should parse and output a templated config", func() {
```
caseydavenport: I'd like to see a test for each supported variable and assert that:
- The right values get templated in
- The right default values (if any) are set.

Happy for that to be a follow-on PR once we get the basic structure right here though.

install_cni_test.go
```diff
@@ -0,0 +1,188 @@
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result).To(ContainSubstring("/host/opt/cni/bin is non-writeable, skipping"))
+			})
+			It("Should fail when we cannot write the CNI config", func() {
+				err := os.RemoveAll("/host/etc/cni/net.d")
+				if err != nil {
+					Fail("Failed to remove /host/etc/cni/net.d")
+				}
+				_, err = runCniScript()
+
```
caseydavenport: nit: drop the newline between the err and the handling of the err.

install_cni_test.go
```diff
@@ -0,0 +1,188 @@
...
+					}
+					err = file.Close()
+					if err != nil {
+						fmt.Println("Failed to close 10-calico.conff")
+					}
+				}
+
+				Expect(expected).To(Equal(received))
+			})
+		})
```
caseydavenport: Nit: add newlines between Contexts

Makefile
heschlie: I was mostly trying to avoid a 2nd target for testing, but in reality I don't think it was necessary to do so. I don't see why this wouldn't work, I'll see if I can get that working.

install_cni_test.go
heschlie: I agree, and we also have to run these tests as root.  It might be worth doing something to prevent people from not running the containerized tests if possible?

install_cni_test.go
```diff
@@ -0,0 +1,188 @@
+package main_test
```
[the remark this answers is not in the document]
heschlie: I had tried that and failed to get it to be picked up by the testing package, I'll take another stab at it and sort out why it wasn't working.

Makefile
caseydavenport: Yeah, a second target isn't the end of the world - `test-containerized` can just depend on it and it will be functionally the same.

install_cni_test.go
```diff
@@ -0,0 +1,188 @@
...
+func runCniScript() (string, error) {
+	result, err := exec.Command(script).CombinedOutput()
+	if err != nil {
+		return string(result), err
+	}
+	return string(result), nil
+}
+
+var _ = BeforeSuite(func() {
+	err := os.Setenv("SLEEP", "false")
```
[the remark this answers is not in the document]
tmjd: Nice, I like that.

Makefile
tmjd: With running the install_cni container for each test then it might be possible to create a test that can't write one of the cni binaries.

k8s-install/scripts/install-cni.sh
```diff
@@ -136,7 +140,11 @@ echo "CNI config: $(cat ${TMP_CONF})"
 sed -i s/__SERVICEACCOUNT_TOKEN__/${SERVICEACCOUNT_TOKEN:-}/g $TMP_CONF
 
 # Move the temporary CNI config into place.
-mv $TMP_CONF /host/etc/cni/net.d/${FILENAME}
+if ! mv $TMP_CONF /host/etc/cni/net.d/${FILENAME}
```
caseydavenport: Wonder, can we just `set -e` at the top of the script and then we won't need the special case handling here? It will also ensure we catch all errors in the script.

What's not clear to me is if there are errors we're OK with, if they occur.  In which case we'd need to explicitly add exceptions for them.

heschlie: I was considering the same, but thought we had planned for some potential errors, I'll tinker with it and if we can just add `set -e` I'll do so.

Makefile
heschlie: I tried installing `chattr` on Alpine, while it installs the filesystem is uses does not support the flags you can set with it so it simply errors out when you run it, so I'm still hunting for a way to make the files not writable by root.

heschlie: Turns out this isn't quite so simple, we don't have go installed in the `calico/cni` container, and cannot run Ginkgo, I can add it to the container if we feel it is worth it, WDYGT?

tmjd: Docker in docker?
I don't think it should be added.

caseydavenport: Definitely should not be added.

Maybe we need the test code itself to run on the local machine, but launch the `calico/cni` container via docker rather than running the script directly.

@ozdanborne recently did this for the FV tests in kube-controllers: https://github.com/projectcalico/kube-controllers/blob/master/Makefile#L122

heschlie:
It looks like the semaphore Ubuntu image does not have ginkgo installed on it, should we containerize the `test-install-cni` or run a setup command to install ginkgo on the box before starting?

caseydavenport:
@heschlie what do we do elsewhere? I suspect we just install it on the build machine, but we should check what the other repos do so we're consistent.

heschlie:
Looking around most of our tests seem to be containerized now, I did see that in the kube-contorllers repo we are compiling the tests using the go-build container then using that to run the tests. I've replicated that here, seems to work well.

.gitignore
```diff
@@ -10,3 +10,5 @@ delstats
 *.pyc
 *.created
 tmp-cni
+k8s-install/scripts/tmp/
```
caseydavenport: I was imagining using something like `mktemp`: https://www.mktemp.org/manual.html

That way we don't need to check in a `tmp` directory in the repo. WDYT?

Makefile
caseydavenport: @heschlie do we still need this mount now?

Makefile
```diff
@@ -124,11 +124,29 @@ test-containerized: run-etcd run-k8s-apiserver build-containerized dist/host-loc
...
 	$(CALICO_BUILD) sh -c '\
 			cd  /go/src/github.com/projectcalico/cni-plugin && \
 			ginkgo'
 	make stop-etcd
 
+.PHONY: k8s-install/scripts/install_cni.test
+## We pre-build the test binary so that we can run it outside a container and allow it
```
caseydavenport: I don't think these should be `##` - that indicates the text should be emitted when printing `make help`

Makefile
```diff
@@ -124,11 +124,29 @@ test-containerized: run-etcd run-k8s-apiserver build-containerized dist/host-loc
...
+	docker run --rm \
+	-e LOCAL_USER_ID=$(LOCAL_USER_ID) \
+	-v $(CURDIR):/go/src/github.com/projectcalico/cni-plugin:rw \
+	-v $(CURDIR)/.go-pkg-cache:/go/pkg/:rw \
+		$(CALICO_BUILD) sh -c '\
+			cd /go/src/github.com/projectcalico/cni-plugin && \
+			go test ./k8s-install/scripts -c --tags install_cni_test -o ./k8s-install/scripts/install_cni.test'
+
+.PHONY: test-install-cni
+## Test the install-cni.sh script
```
caseydavenport: These ones are fine to have as `##` since it's a "user-facing" make target we'd want to show up in `make help`

Makefile
```diff
@@ -137,6 +155,7 @@ run-test-containerized-without-building: run-etcd run-k8s-apiserver
 	-e PLUGIN=calico \
 	-e CNI_SPEC_VERSION=$(CNI_SPEC_VERSION) \
 	-v $(CURDIR):/go/src/github.com/projectcalico/cni-plugin:rw \
+	-v $(CURDIR)/dist:/opt/cni/bin \
```
caseydavenport: Can we remove this mount?

k8s-install/scripts/install_cni_test.go
```diff
@@ -0,0 +1,208 @@
...
+		"-e", "KUBERNETES_SERVICE_PORT=8080",
+		"-v", cwd + "/tmp/bin:/host/opt/cni/bin",
+		"-v", cwd + "/tmp/net.d:/host/etc/cni/net.d",
+		"-v", cwd + "/tmp/serviceaccount:/var/run/secrets/kubernetes.io/serviceaccount",
+	}
+	args = append(args, extraArgs...)
+	args = append(args, "calico/cni:latest", "/install-cni.sh")
+
+	c := containers.Run("cni", args...)
+	for !c.Stopped() {
```
caseydavenport: It's scary to loop forever here - could we instead use a gomega `Eventually` statement?

https://onsi.github.io/gomega/#eventually

k8s-install/scripts/install_cni_test.go
```diff
@@ -0,0 +1,208 @@
...
+						fmt.Println("Failed to close 10-calico.conff")
+					}
+				}
+
+				Expect(expected).To(Equal(received))
+			})
+		})
+
+		Context("With modified env vars", func() {
+			It("Should rename '10-calico.conf' to '10-test.conf'", func() {
```
caseydavenport: nit: let's make the expected name more realistic.

e.g. `10-calico.conflist` is something we actually set this to in a bunch of places.  I doubt there will ever be something that affects that case but not others, but I like the tests to be more realistic where possible.

k8s-install/scripts/install_cni_test.go
```diff
@@ -0,0 +1,208 @@
...
+		fmt.Println("Failed to clean up directories")
+	}
+})
+
+var _ = Describe("install-cni.sh tests", func() {
+	AfterEach(func() {
+		cleanup()
+	})
+
+	Describe("Run install-cni", func() {
```
caseydavenport: I think we should have one test case which expects the install-cni.sh command to fail.

Let's run a test where we set:
```
KUBERNETES_SERVICE_HOST=""
KUBERENTES_SERVICE_PORT=""
```

We've hit bugs in the past where those values aren't set and we added special code so that we fail in that case.

https://github.com/projectcalico/cni-plugin/blob/master/k8s-install/scripts/install-cni.sh#L117

k8s-install/scripts/install-cni.sh
```diff
@@ -6,7 +6,7 @@
 # - Expects the desired CNI config in the CNI_NETWORK_CONFIG env variable.
 
 # Ensure all variables are defined.
```
caseydavenport: comment should be updated to say "and that the script fails when an error is hit"

k8s-install/scripts/install_cni_test.go
```diff
@@ -0,0 +1,208 @@
...
+func runCniContainer(extraArgs ...string) error {
+	// Get the CWD for mounting directories into container.
+	cwd, err := os.Getwd()
+	if err != nil {
+		Fail("Could not get CWD")
+	}
+
+	// Assemble our arguments.
+	args := []string{
+		"--privileged",
```
tmjd: Is it necessary to use privileged here?

k8s-install/scripts/install_cni_test.go
```diff
@@ -0,0 +1,208 @@
...
+						Fail("Could not read from 10-calico.conf")
+					}
+					err = file.Close()
+					if err != nil {
+						fmt.Println("Failed to close 10-calico.conff")
+					}
+				}
+
+				Expect(expected).To(Equal(received))
+			})
```
tmjd: Would it be possible to mount a file now that couldn't be overwritten in the container to test the failure of writing a bin file? Either chattr the file before or possibly using a `-v` option with RO.

heschlie: I think so, but at the moment the container package doesn't support a way to get errors back from the container, so I had to drop the tests which check for it. But mounting it as RO into the container should do the trick.

k8s-install/scripts/install_cni_test.go
heschlie: I believe not, will remove and verify.

k8s-install/scripts/install_cni_test.go
heschlie: I can add the test, but at the moment there is no way to get the error codes back from the container, so I had to drop a couple of the tests which check for it.

Makefile
heschlie: Ah, didn't know that, will fix.

.gitignore
heschlie: Well this is explicitly ignoring it so the directory is not checked in. But are you thinking we run `mktmp` in the make target and had it off to the script?

k8s-install/scripts/install_cni_test.go
caseydavenport: there's no way to tell if a container failed or succeeded?

hmm...

heschlie: So, I was trying the `c.DockerInspect("{{.State.ExitCode}}")` command, but it was after verifying the container was closed, which is after the container has been destroyed and errors out with `Error: No such object`. If you run the inspect before it is closed then it grabs the exit status before the script has finished, so it always returns 0, I verified this by forcing the script to `exit 1` which is reported by the logger, but the docker inspect shows 0:

```shell
# docker inspect output with no format
[
    {
        ...
        "State": {
            ...
            "ExitCode": 0,
            ...
            }
      }
]

# log message
INFO[0013] Container stopped                             error="exit status 1" name=cni-838-6-felixfv
```

I might be missing something obvious, but AFAICT there isn't a way to check it using the current API for the Container.

k8s-install/scripts/install_cni_test.go
```diff
@@ -0,0 +1,204 @@
...
+				if file, err := os.Open("tmp/net.d/10-calico.conflist"); err != nil {
+					Fail("Could not open 10-calico.conflist for reading")
+				} else {
+					_, err := file.Read(received)
+					if err != nil {
+						Fail("Could not read from 10-calico.conflist.conf")
+					}
+					err = file.Close()
+					if err != nil {
+						fmt.Println("Failed to close 10-calico.conflist.conff")
```
caseydavenport: typo: 10-calico-conflist.conff should just be 10-calico.conflist (A few other places in this PR with the same issue)

heschlie: Gah, how did I manage to do that.

caseydavenport:
@heschlie could you squash you commits?

Makefile
```diff
@@ -129,6 +129,23 @@ test-containerized: run-etcd run-k8s-apiserver build-containerized dist/host-loc
 			ginkgo'
 	make stop-etcd
 
+.PHONY: k8s-install/scripts/install_cni.test
```
caseydavenport: @heschlie just noticed this - I don't think should be marked as .PHONY

Also, we need to add this binary to `make clean`

heschlie: Ah you're right, it spits out a binary, standby.

heschlie: OK, fixed and kept squashed.

PR #387, opened 2017-10-04
Adding support for ppc64le.

djlwilder:
Signed-off-by: David Wilder <wilder@us.ibm.com>

## Description
This add multi-arch support and ppc64le support to the build.

Summery of the changes:
- Build with ARCH=ppc64le make <...>
- If ARCH is undefined it defaults to amd64.
- The dist directory is now architecture specific, dist-$(ARCH).
- New environment variable DIST is passed to the test code.
- Various images are renamed to include -($ARCH) in their names.

I verified "make ci" runs on both amd64 and ppc64le.

For arm64 builds it should be trivial to merge the Makefile changes from pull request #371 on top of my changes. Adding something like this at the top of the Makefile should do.

ifeq ($(ARCH),arm64)
        ARCHTAG:=-arm64
	CALICO_BUILD=linaro/ci-aarch64-leg-go-build-alpine
	ETCD_CONTAINER=linaro/ci-aarch64-leg-etcd
endif

We still need a Dockerfile-arm64 to be included.  Hopefully @TrevorTaoARM can submit that.


TrevorTaoARM:
My comments:
1. I think there is NO need to add a dist-ppc64le or dist-arm64, since for the user of each platform, he or she  would obviously build/set his/her own binaries in dist/ directory. 
So there is NO need to modify the "dist" in the Makefile to a variable "DIST" and NO need to add a new Dockerfile.  Everything is very natural in this way.
2.  If you want to add ppc64le support in the Makefile, please follow my PR#371:
https://github.com/projectcalico/cni-plugin/pull/371
and add the corresponding parts for your ppc64le support.

fasaxc:
@djlwilder Did you and @TrevorTaoARM agree on an approach?

djlwilder:
@TrevorTaoARM
The idea of DIST is to prevent picking up the wrong binary when build for multiple archs using the same source tree.  This is what I did for multi arch support in kube-controllers, except in that case we tag the arch for each created binary.

Can you explain how you build for a different arch with out a Dockerfile for that arch?  At a minimum the base image must match the arch.  Are you planing on updating the Dockerfile only locale?  If that is the case why would my changes prevent that?

caseydavenport:
@TrevorTaoARM I like having the process build distinct binaries for each platform - this is better for the cross-compilation case where I might be generating release artifacts for multiple platforms on the same build machine and like @djlwilder mentions above it removes the worry about accidentally packaging the binaries for the wrong platform.

I would have thought some directory structure like this would be a good way to do it here:

```
./dist
  ./dist/amd64
    calico
    calico-ipam
    . . .
  ./dist/ppc64le
    calico
    calico-ipam
    . . .
  ./dist/arm64
    calico
    calico-ipam
    . . .
```

djlwilder:
My changes are currently doing this:
```
./dist-amd64
	calico
	calico-ipam
        ....
./dist-ppc64le
	calico
	calico-ipam
        ....
./dist-arm64
	calico
	calico-ipam
        ....
```
@caseydavenport suggestion is fine with me as well,  what approach do you prefer?

TrevorTaoARM:
@caseydavenport 
I understand your cross-compilation requirement to get all the releases for each platform.  If you think it is necessary, I would agree with it.

My concern here is the conciseness of the code.  For example, for the Dockerfile, I just changed the base image 'FROM busybox' to 'FROM aarch64/busybox' in Dockerfile and just 
named it as Dockerfile.arm64, all other things are kept as is.  I got the docker image after a normal 'docker build'. I think most users are just concern about their own platform.

caseydavenport:
@djlwilder yep that works for me - let's keep it as you have it to minimize the changes required. @TrevorTaoARM is that OK with you?

We'll need to maintain multiple Dockerfiles anyway, at which point having them pull in different binaries isn't such a big deal IMO.

TrevorTaoARM:
@caseydavenport 
It's fine with me. 
I think your way of having a single 'dist' dir but with 3 subdir 'dist-xxx' is better. 
I could provide the corresponding changes for those of arm64, including Dockerfile, Makefile.

djlwilder:
Thanks $TrevorTaoARM 
I am confused,  were you agreeing to dist-$(ARCH)  or dist/$(ARCH)?

TrevorTaoARM:
@djlwilder 
I prefer dist/$(ARCH) personally.  Thanks.

djlwilder:
Ok, I changing dist-$ARCH to dist/$ARCH and updated the Dockerfiles and .dockerignore.  BTW:  don't forget to update the .dockerignore when adding a new arch :)

djlwilder:
@fasaxc,  @TrevorTaoARM we should be in agreement now with the multi-arch changes.
I am not sure what is up with Ci, I don't think it was run after the the last commit.

djlwilder:
@fasaxc I believe we are in agreement with all these changes.
I don't know what happened with the Ci run.  It runs ok for me.  Maybe this is a transient failure?  Can you re-submit it?

PR #388, opened 2017-10-04
 Update CNI to use the V2 API

gunjan5:
## Description

Update CNI to use the V2 API
I've also taken this opportunity to refactor the code a little bit.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
None Required
```

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -14,27 +14,28 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
-	goerrors "errors"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
 	"runtime"
 
-	"net"
-
 	"github.com/containernetworking/cni/pkg/ipam"
 	"github.com/containernetworking/cni/pkg/skel"
-	"github.com/containernetworking/cni/pkg/types"
+	cnitypes "github.com/containernetworking/cni/pkg/types"
 	"github.com/containernetworking/cni/pkg/types/current"
 	cniSpecVersion "github.com/containernetworking/cni/pkg/version"
 	"github.com/projectcalico/cni-plugin/k8s"
-	. "github.com/projectcalico/cni-plugin/utils"
-	"github.com/projectcalico/libcalico-go/lib/api"
-	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/cni-plugin/types"
+	"github.com/projectcalico/cni-plugin/utils"
+	api "github.com/projectcalico/libcalico-go/lib/apiv2"
+	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/logutils"
-	cnet "github.com/projectcalico/libcalico-go/lib/net"
-	log "github.com/sirupsen/logrus"
+	"github.com/projectcalico/libcalico-go/lib/options"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var nodename string
@@ -44,71 +45,125 @@ func init() {
 	// since namespace ops (unshare, setns) are done for a single thread, we
 	// must ensure that the goroutine does not jump from OS thread to thread
 	runtime.LockOSThread()
-
-	nodename, _ = os.Hostname()
-}
-
-func updateNodename(conf NetConf, logger *log.Entry) {
-	if conf.Hostname != "" {
-		nodename = conf.Hostname
-		logger.Warn("Configuration option 'hostname' is deprecated, use 'nodename' instead.")
-	}
-	if conf.Nodename != "" {
-		nodename = conf.Nodename
-	}
 }
 
 func cmdAdd(args *skel.CmdArgs) error {
 	// Unmarshal the network config, and perform validation
-	conf := NetConf{}
+	conf := types.NetConf{}
 	if err := json.Unmarshal(args.StdinData, &conf); err != nil {
 		return fmt.Errorf("failed to load netconf: %v", err)
 	}
 
-	cniVersion := conf.CNIVersion
-
-	ConfigureLogging(conf.LogLevel)
-
-	workload, orchestrator, err := GetIdentifiers(args)
-	if err != nil {
-		return err
-	}
-
-	logger := CreateContextLogger(workload)
+	utils.ConfigureLogging(conf.LogLevel)
 
 	// Allow the nodename to be overridden by the network config
-	updateNodename(conf, logger)
+	nodename = utils.DetermineNodename(conf)
 
-	logger.WithFields(log.Fields{
-		"Orchestrator": orchestrator,
-		"Node":         nodename,
-		"Workload":     workload,
-		"ContainerID":  args.ContainerID,
-	}).Info("Extracted identifiers")
-
-	logger.WithFields(log.Fields{"NetConfg": conf}).Info("Loaded CNI NetConf")
-	calicoClient, err := CreateClient(conf)
+	// Extract WEP identifiers such as pod name, pod namespace (for k8s), containerID, IfName.
+	wepIDs, err := utils.GetIdentifiers(args, nodename)
 	if err != nil {
 		return err
 	}
 
-	// Always check if there's an existing endpoint.
-	endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{
-		Node:         nodename,
-		Orchestrator: orchestrator,
-		Workload:     workload})
+	logrus.WithField("EndpointIDs", wepIDs).Info("Extracted identifiers")
+	logrus.WithField("NetConfg", conf).Info("Loaded CNI NetConf")
+
+	calicoClient, err := utils.CreateClient(conf)
 	if err != nil {
 		return err
 	}
 
-	logger.Debugf("Retrieved endpoints: %v", endpoints)
+	// Remove the endpoint field (IfName) from the wepIDs so we can get a WEP name prefix.
+	// We use the WEP name prefix (e.g. prefix: "node1-k8s-mypod--1-", full name: "node1-k8s-mypod--1-eth0"
+	// to list all the WEPs so if we have a WEP with a different IfName (e.g. "node1-k8s-mypod--1-eth1")
+	// we could still get that.
+	wepIDs.Endpoint = ""
+
+	// Calculate the workload name prefix from the WEP specific identifiers
+	// for the given orchestrator.
+	wepPrefix, err := wepIDs.CalculateWorkloadEndpointName(true)
+	if err != nil {
+		return fmt.Errorf("error constructing WorkloadEndpoint prefix: %s", err)
+	}
+
+	ctx := context.Background()
+
+	// Check if there's an existing endpoint by listing the existing endpoints based on the WEP name prefix.
+	endpoints, err := calicoClient.WorkloadEndpoints().List(ctx, options.ListOptions{Name: wepPrefix, Namespace: wepIDs.Namespace, Prefix: true})
+	if err != nil {
+		return err
+	}
+
+	var logger *logrus.Entry
+	if wepIDs.Orchestrator == "k8s" {
+		logger = logrus.WithFields(logrus.Fields{
+			"WorkloadEndpoint": fmt.Sprintf("%s%s", wepPrefix, wepIDs.Endpoint),
+			"ContainerID":      wepIDs.ContainerID,
+			"Pod":              wepIDs.Pod,
+			"Namespace":        wepIDs.Namespace,
+		})
+	} else {
+		logger = logrus.WithFields(logrus.Fields{
+			"ContainerID": wepIDs.ContainerID,
+		})
+	}
+
+	logger.Debugf("Retrieved list of endpoints: %v", endpoints)
 
 	var endpoint *api.WorkloadEndpoint
-	if len(endpoints.Items) == 1 {
-		endpoint = &endpoints.Items[0]
+
+	// If the prefix list returns 1 or more items, we go through the items and try to see if the name matches the WEP
+	// identifiers we have. The identifiers we use for this match at this point are:
+	// 1. Node name
+	// 2. Orchestrator ID ('cni' or 'k8s')
+	// 3. ContainerID
+	// 4. Pod name (only for k8s)
+	// Note we don't use the interface name (endpoint) for this match.
+	// If we find a match from the returned list then we've found the workload endpoint,
+	// and we reuse that even if it has a different interface name, because
+	// we only support one interface per pod right now.
+	// For example, you have a WEP for a k8s pod "mypod-1", and IfName "eth0" on node "node1", that will result in
+	// a WEP name "node1-k8s-mypod--1-eth0" in the datastore, now you're trying to schedule another pod "mypod",
+	// IfName "eth0" and node "node1", so we do a prefix list to get all the endpoints for that workload, with
+	// the prefix "node1-k8s-mypod-". Now this search would return any existing endpoints for "mypod", but it will also
+	// list "node1-k8s-mypod--1-eth0" which is not the same WorkloadEndpoint, so to avoid that, we go through the
+	// list of returned WEPs from the prefix list and call NameMatches() based on all the
+	// identifiers (pod name, containerID, node name, orchestrator), but omit the IfName (Endpoint field) since we can
+	// only have one interface per pod right now, and NameMatches() will return true if the WEP matches the identifiers.
+	// It is possible that none of the WEPs in the list match the identifiers, which means we don't already have an
+	// existing WEP to reuse. See `names.WorkloadEndpointIdentifiers` GoDoc comments for more details.
+	if len(endpoints.Items) > 0 {
+		logger.Debugf("List of WorkloadEndpoints %v", endpoints.Items)
+		for _, ep := range endpoints.Items {
+			match, err := wepIDs.WorkloadEndpointIdentifiers.NameMatches(ep.Name)
+			if err != nil {
+				// We should never hit this error, because it should have already been
+				// caught by CalculateWorkloadEndpointName.
+				return fmt.Errorf("invalid WorkloadEndpoint identifiers: %v", wepIDs.WorkloadEndpointIdentifiers)
+			}
+
+			if match {
+				logger.Debugf("Found a match for WorkloadEndpoint: %v", ep)
+				endpoint = &ep
+				// Assign the WEP name to wepIDs' WEPName field.
+				wepIDs.WEPName = endpoint.Name
+				// Put the endpoint name from the matched WEP in the identifiers.
+				wepIDs.Endpoint = ep.Spec.Endpoint
+				logger.Infof("Calico CNI found existing endpoint: %v", endpoint)
+				break
+			}
+		}
 	}
 
-	fmt.Fprintf(os.Stderr, "Calico CNI checking for existing endpoint: %v\n", endpoint)
+	// If we don't find a match from the existing WorkloadEndpoints then we calculate
+	// the WEP name with the IfName passed in so we can create the WorkloadEndpoint later in the process.
+	if endpoint == nil {
+		wepIDs.Endpoint = args.IfName
+		wepIDs.WEPName, err = wepIDs.CalculateWorkloadEndpointName(false)
+		if err != nil {
+			return fmt.Errorf("error constructing WorkloadEndpoint name: %s", err)
+		}
+	}
 
 	// Collect the result in this variable - this is ultimately what gets "returned" by this function by printing
 	// it to stdout.
@@ -116,8 +171,8 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 	// If running under Kubernetes then branch off into the kubernetes code, otherwise handle everything in this
 	// function.
-	if orchestrator == "k8s" {
-		if result, err = k8s.CmdAddK8s(args, conf, nodename, calicoClient, endpoint); err != nil {
+	if wepIDs.Orchestrator == "k8s" {
+		if result, err = k8s.CmdAddK8s(ctx, args, conf, *wepIDs, calicoClient, endpoint); err != nil {
 			return err
 		}
 	} else {
@@ -131,9 +186,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 			// Don't create the veth or do any networking.
 			// Just update the profile on the endpoint. The profile will be created if needed during the
 			// profile processing step.
-			fmt.Fprintf(os.Stderr, "Calico CNI appending profile: %s\n", profileID)
+			logger.Infof("Calico CNI appending profile: %s\n", profileID)
 			endpoint.Spec.Profiles = append(endpoint.Spec.Profiles, profileID)
-			result, err = CreateResultFromEndpoint(endpoint)
+			result, err = utils.CreateResultFromEndpoint(endpoint)
 			logger.WithField("result", result).Debug("Created result from endpoint")
 			if err != nil {
 				return err
@@ -145,7 +200,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 			// 3) Create the veth, configuring it on both the host and container namespace.
 
 			// 1) Run the IPAM plugin and make sure there's an IP address returned.
-			logger.WithFields(log.Fields{"paths": os.Getenv("CNI_PATH"),
+			logger.WithFields(logrus.Fields{"paths": os.Getenv("CNI_PATH"),
 				"type": conf.IPAM.Type}).Debug("Looking for IPAM plugin in paths")
 			ipamResult, err := ipam.ExecAdd(conf.IPAM.Type, args.StdinData)
 			logger.WithField("IPAM result", ipamResult).Info("Got result from IPAM plugin")
@@ -159,13 +214,13 @@ func cmdAdd(args *skel.CmdArgs) error {
 			// See CNI Spec doc for more details.
 			result, err = current.NewResultFromResult(ipamResult)
 			if err != nil {
-				ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
+				utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
 				return err
 			}
 
 			if len(result.IPs) == 0 {
-				ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
-				return goerrors.New("IPAM plugin returned missing IP config")
+				utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
+				return errors.New("IPAM plugin returned missing IP config")
 			}
 
 			// Parse endpoint labels passed in by Mesos, and store in a map.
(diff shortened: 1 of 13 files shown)


k8s/k8s.go
```diff
@@ -282,13 +281,12 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
 		utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
 		return nil, err
 	}
-	endpoint.Spec.MAC = &cnet.MAC{HardwareAddr: mac}
+	endpoint.Spec.MAC = mac.String()
 	endpoint.Spec.InterfaceName = hostVethName
-	endpoint.Metadata.ActiveInstanceID = args.ContainerID
```
caseydavenport: We still need to update the the ContainerID here!

utils/utils.go
```diff
@@ -44,6 +50,48 @@ func Min(a, b int) int {
...
+	}
+	if conf.Nodename != "" {
+		nodename = conf.Nodename
+	}
+	return nodename
+}
+
+// CreateOrUpdate tries to create the WEP first, if it already exists then it will try to Update it.
+// This is similar to what was previously called Apply operation.
+func CreateOrUpdate(ctx context.Context, client client.Interface, wep *api.WorkloadEndpoint) (*api.WorkloadEndpoint, error) {
```
caseydavenport: Shouldn't this logic be:

```
if wep.ResourceVersion:
  Update(wep)
else:
  Create(wep)
```

utils/utils.go
```diff
@@ -44,6 +50,48 @@ func Min(a, b int) int {
 	return b
 }
 
+// DetermineNodename gets the node name, in order of priority:
+// 1. Hostname field in NetConf.
```
caseydavenport: Comment is wrong - it's "Hostname" that is deprecated.

calico.go
```diff
@@ -44,80 +45,93 @@ func init() {
...
-		"Orchestrator": orchestrator,
-		"Node":         nodename,
-		"Workload":     workload,
-		"ContainerID":  args.ContainerID,
-	}).Info("Extracted identifiers")
-
+	logger := &log.Entry{}
+	if epIDs.OrchestratorID == "k8s" {
+		logger = log.WithFields(log.Fields{
+			"Workload":    epIDs.WEPName,
```
robbrockbank: I don't think we need workload anymore here.

calico.go
```diff
@@ -44,80 +45,93 @@ func init() {
...
+			"Namespace":   epIDs.Namespace,
+		})
+	} else {
+		logger = log.WithFields(log.Fields{
+			"Workload":    epIDs.WEPName,
+			"ContainerID": epIDs.ContainerID,
+		})
+	}
+	logger.Info("Extracted identifiers")
 	logger.WithFields(log.Fields{"NetConfg": conf}).Info("Loaded CNI NetConf")
```
robbrockbank: You can just use logger.WithField() singular rather than plural here.

calico.go
```diff
@@ -44,80 +45,93 @@ func init() {
...
+
+	// Check if there's an existing endpoint.
+	endpoints, err := calicoClient.WorkloadEndpoints().List(ctx, options.ListOptions{Name: epIDs.WEPName, Namespace: epIDs.Namespace, Prefix: true})
 	if err != nil {
 		return err
 	}
 
-	logger.Debugf("Retrieved endpoints: %v", endpoints)
+	endpoint := &api.WorkloadEndpoint{}
+	ids := names.WorkloadEndpointIdentifiers{
```
robbrockbank: I think this logic might be a little off.

IIUC, you want to zero-out the endpoint before you do the List and the NameMatches check - that way you're doing a prefix list without the endpoint name - so basically enumerating all endpoints under the workload (in the old hierarchical sense).

So rather than copying this struct I think just remove the endpoint and store off the old value for later.

robbrockbank: So, I don't think you need to copy this struct like this (and beside it feels like a bad thing to have to maintain)

utils/utils.go
```diff
@@ -153,42 +207,83 @@ func CreateResultFromEndpoint(ep *api.WorkloadEndpoint) (*current.Result, error)
...
+			ipNet.Address.Mask = net.CIDRMask(128, 128)
+		}
+
+		wep.Spec.IPNetworks = append(wep.Spec.IPNetworks, ipNet.Address.String())
+	}
+
+	return nil
+}
+
+type EndpointIdentifiers struct {
```
robbrockbank: Do you really gain anything here by defining your own type?  What's wrong with using the one from libcalico-go?

If this didn't have the WEP name in it then it would basically be the same, and it's easy to calculate the name once you have the WorkloadEndpointIdentifiers struct in your hand.

calico.go
robbrockbank: Oh - this is a completely different WorkloadEndpointsIdentifiers struct.  Hmmm - I think that's not necessary (see comment below).

utils/utils.go
```diff
@@ -153,42 +207,83 @@ func CreateResultFromEndpoint(ep *api.WorkloadEndpoint) (*current.Result, error)
...
+// GetIdentifiers takes CNI command arguments, and extracts identifiers i.e. pod name, pod namespace,
+// container ID, and orchestratorID based on the orchestrator.
+func GetIdentifiers(args *skel.CmdArgs, nodename string) (*EndpointIdentifiers, error) {
 	// Determine if running under k8s by checking the CNI args
-	k8sArgs := K8sArgs{}
-	if err = types.LoadArgs(args.Args, &k8sArgs); err != nil {
-		return workloadID, orchestratorID, err
+	k8sArgs := types.K8sArgs{}
+	if err := cniTypes.LoadArgs(args.Args, &k8sArgs); err != nil {
+		return &EndpointIdentifiers{"", "", "", "", "", "", ""}, err
```
robbrockbank: You don't need to include the values like that.  Just use &EndpointIdentifiers{}.
In fact, it's a pointer, why can't you just return nil?

calico.go
robbrockbank: And if we aren't re-using the endpoint, why did we need to enumerate it?

calico.go
caseydavenport: We still want something that identifies the name of the thing, right? It could be called "PodName" or something instead, or just "WorkloadEndpoint".

k8s/k8s.go
gunjan5: dunno what I was thinking, I moved it up in the `if` block, but I realize this would bring back the double ADD bug you recently fixed. I'll move it back to here

utils/utils.go
gunjan5: The current logic ultimately does the same thing, so I think the logic is correct but your logic is a lot easier to read. I'll update it

calico.go
gunjan5: I'm changing it to `WorkloadEndpoint` since that'll be the same for k8s and non-k8s

utils/utils.go
caseydavenport: Well, the logic is actually different - the way I've suggested you don't need to perform the `Create` operation, whereas as it is now you always execute `Create` and only do an `Update` if `Create` fails.

utils/utils.go
gunjan5: needed one extra field, but it makes sense. I'll just embed the libcalico one

calico.go
```diff
@@ -44,80 +44,91 @@ func init() {
...
 		return err
 	}
 
-	logger.Debugf("Retrieved endpoints: %v", endpoints)
-
 	var endpoint *api.WorkloadEndpoint
-	if len(endpoints.Items) == 1 {
-		endpoint = &endpoints.Items[0]
+
+	if len(endpoints.Items) >= 1 {
```
robbrockbank: >= 1 ;  i think >0 is easier to grok at a glance.

robbrockbank: also, this check is redundant.  if there are 0 entries in the Items, then the for loop won't do anything.

calico.go
```diff
@@ -44,80 +44,91 @@ func init() {
...
+		}
+
+		if endpoint == nil {
+			return fmt.Errorf("WorklodaEndpoint name does not match the provided identifiers")
+		}
 	}
 
-	fmt.Fprintf(os.Stderr, "Calico CNI checking for existing endpoint: %v\n", endpoint)
+	logger.Debugf("Retrieved endpoints: %v", endpoint)
+	logger.Infof("Calico CNI checking for existing endpoint: %v", endpoint)
```
robbrockbank: I don't get this log - i thought we'd just checked for existing endpoint?

calico.go
```diff
@@ -44,80 +44,91 @@ func init() {
...
-	// Allow the nodename to be overridden by the network config
-	updateNodename(conf, logger)
-
-	logger.WithFields(log.Fields{
-		"Orchestrator": orchestrator,
-		"Node":         nodename,
-		"Workload":     workload,
-		"ContainerID":  args.ContainerID,
-	}).Info("Extracted identifiers")
+	logger := &log.Entry{}
```
robbrockbank: since you are never using that value, it's more common I think to just define the var, so:

`var logger *log.Entry`

calico.go
```diff
@@ -44,80 +44,91 @@ func init() {
...
 
-	// Always check if there's an existing endpoint.
-	endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{
-		Node:         nodename,
-		Orchestrator: orchestrator,
-		Workload:     workload})
+	ctx := context.Background()
+
+	// Check if there's an existing endpoint.
+	endpoints, err := calicoClient.WorkloadEndpoints().List(ctx, options.ListOptions{Name: epIDs.Workload, Namespace: epIDs.Namespace, Prefix: true})
```
robbrockbank: Unless I'm missing something, the name is NOT the workload.  You need to construct the prefix using the set of IDs that you have, and removing the endpoint.

calico.go
```diff
@@ -44,80 +44,91 @@ func init() {
...
-	logger.WithFields(log.Fields{
-		"Orchestrator": orchestrator,
-		"Node":         nodename,
-		"Workload":     workload,
-		"ContainerID":  args.ContainerID,
-	}).Info("Extracted identifiers")
+	logger := &log.Entry{}
+	if epIDs.Orchestrator == "k8s" {
+		logger = log.WithFields(log.Fields{
+			"WorkloadEndpoint": epIDs.Workload,
```
robbrockbank: What is the Workload for a k8s orchestrator, I think I asked that before but got shouted down.  Shouldn't it be blank?

We'll have a Pod, Namespace, Endpoint, Node that uniquely identifier the WEP in k8s no?  The Workload is not required ... unless I'm misunderstanding.  Let's chat.

robbrockbank: and ContainerID as an additional identifier too of course.

calico.go
```diff
@@ -251,28 +257,23 @@ func cmdAdd(args *skel.CmdArgs) error {
...
-				Metadata: api.ProfileMetadata{
-					Name: conf.Name,
-					Tags: []string{conf.Name},
-				},
-				Spec: api.ProfileSpec{
-					EgressRules:  []api.Rule{{Action: "allow"}},
-					IngressRules: inboundRules,
-				},
-			}
+			profile := api.NewProfile()
```
robbrockbank: not sure why this changed.  The old code was more readable.  

I think it's because you thought we needed the TypeMetadata, but we shouldn't.

testutils/utils.go
```diff
@@ -388,5 +422,7 @@ func CheckSysctlValue(sysctlPath, value string) error {
 		return fmt.Errorf("error asserting sysctl value: expected: %s, got: %s for sysctl path: %s", value, string(buf), sysctlPath)
 	}
 
+	_ = fh.Close()
```
robbrockbank: what was wrong with using defer?

testutils/utils.go
```diff
@@ -340,9 +374,10 @@ func DeleteContainerWithId(netconf, netnspath, name, containerId string) (exitCo
 	if err != nil {
 		return
 	}
-	io.WriteString(stdin, netconf)
-	io.WriteString(stdin, "\n")
-	stdin.Close()
+
+	_, _ = io.WriteString(stdin, netconf)
```
robbrockbank: why?

testutils/utils.go
```diff
@@ -244,9 +278,9 @@ func RunCNIPluginWithId(netconf, k8sName, ip, containerId string, targetNs ns.Ne
 		panic("some error found")
 	}
 
-	io.WriteString(stdin, netconf)
-	io.WriteString(stdin, "\n")
-	stdin.Close()
+	_, _ = io.WriteString(stdin, netconf)
```
robbrockbank: not sure why you changed this.

robbrockbank: unless you plan to do some Expect or err checking then best left as it.

testutils/utils.go
```diff
@@ -134,9 +168,9 @@ func RunIPAMPlugin(netconf, command, args, cniVersion string) (*current.Result,
...
```
robbrockbank: if you ignore all return parms then no need to capture them at all.

ipam/calico-ipam.go
```diff
@@ -14,24 +14,24 @@
...
 	"fmt"
 	"net"
-
 	"os"
 
-	log "github.com/sirupsen/logrus"
-
 	"github.com/containernetworking/cni/pkg/skel"
-	"github.com/containernetworking/cni/pkg/types"
+	cniTypes "github.com/containernetworking/cni/pkg/types"
```
robbrockbank: it's unusual to use camelcase for imports isn't it, or am i making that?

ipam/calico-ipam.go
```diff
@@ -14,24 +14,24 @@
...
 	"github.com/containernetworking/cni/pkg/types/current"
 	cniSpecVersion "github.com/containernetworking/cni/pkg/version"
+	"github.com/projectcalico/cni-plugin/types"
```
robbrockbank: maybe have this as calicotypes too, to avoid confusion?

robbrockbank: actually, it looks like this change is because you moved a file from utils to types.

IMHO both utils and types are rather meaningless package names so I don't think this rename has actually done anything other than cause a conflict over two different type packages.  I think this would have been best left unchanged.

testutils/utils.go
gunjan5: this is not refactoring. We have static error checking as part of the CI so it fails unless all the errors are handled or ignored, and it only checks the files that changed, and it was added to the go-build container recently that's why no one hit it until now

testutils/utils.go
gunjan5: see the other comment

testutils/utils.go
robbrockbank: ok, so if we're being forced to change due to CI - should we not be asserting something about the return values?

testutils/utils.go
robbrockbank: yup yup

ipam/calico-ipam.go
gunjan5: we had camelcase imports at a bunch of places, but you're right, it feels weird. I'll make it all lowercase

ipam/calico-ipam.go
gunjan5: I prefer not to unless you have a strong preference

testutils/utils.go
gunjan5: can't do it because Close returns an error which needs to be checked or ignored

calico.go
gunjan5: ok if it works without metadata then I'll use the old style of code

calico.go
caseydavenport: @robbrockbank we might be talking in circles about the same thing here.

Basically, we just need logs to be emitted which include the following information:

- WorkloadEndpoint Name
- The ContainerID of the container being networked.
- The Kubernetes pod Name and Namespace pair.

calico.go
caseydavenport: Yea the wording of this log has always been weird.
It should say something like "Calico CNI found existing endpoint: %v"

ipam/calico-ipam.go
robbrockbank: yeah I'm not overly fussed since it's done.

ipam/calico-ipam.go
robbrockbank: ditto.

robbrockbank: yeah i'm not overly fussed.

testutils/utils.go
robbrockbank: can we prevent validation in the testutils module since it is just test code?

calico.go
gunjan5: agree with `>0` part, but I'd still like to keep that if statement because it's more explicit

testutils/utils.go
gunjan5: I think there's no easy fix since this is from the go-build container which affects all the Calico repos

calico.go
```diff
@@ -44,80 +45,113 @@ func init() {
...
+	}
+
+	logger.Debugf("Retrieved list of endpoints: %v", endpoints)
 
 	var endpoint *api.WorkloadEndpoint
-	if len(endpoints.Items) == 1 {
-		endpoint = &endpoints.Items[0]
+	// Put the endpoint name back in the identifiers
+	// so we can get the exact WEP match from the WEP list.
+	epIDs.Endpoint = args.IfName
```
robbrockbank: Logic here is still not quite right.

You are setting the endpoint back to the original value too early - which means you are checking to see if the enumerated endpoint also has the same endpoint ID.  In which case, you might just as well have done an exact get rather than a prefix list.

Your logic is:
e.g.
WEP ids are `node1, cni, container1, eth0`

You do a prefix query by removing the `eth0` - so you get get a list of the endpoints for container1.
You get returned an endpoint for eth1.
You drop it on the floor because you are checking again for eth0.

robbrockbank: You need to check it's a valid endpoint *before* putting the endpoint back in.

You then need to extract the endpoint ID from the workload Spec.

calico.go
```diff
@@ -131,9 +165,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 			// Don't create the veth or do any networking.
 			// Just update the profile on the endpoint. The profile will be created if needed during the
 			// profile processing step.
-			fmt.Fprintf(os.Stderr, "Calico CNI appending profile: %s\n", profileID)
+			log.Infof("Calico CNI appending profile: %s\n", profileID)
```
robbrockbank: shouldn't we be using logger here?

if there is a risk of accidentally using log rather than some predefined logger - IMHO I think it's a good idea to rename the log import to something other than log -- makes it less likely to make that mistake.

calico.go
robbrockbank: A test for this case would have probably highlighted this early on and saved some cycles.

gunjan5:
@robbrockbank ok I've moved that line to after NameMatch() and added a test for it. I've also renamed logrus imports to `logrus`. PTAL

calico.go
```diff
@@ -44,80 +45,111 @@ func init() {
...
+			"ContainerID": epIDs.ContainerID,
+		})
+	}
+
+	logger.Debugf("Retrieved list of endpoints: %v", endpoints)
 
 	var endpoint *api.WorkloadEndpoint
-	if len(endpoints.Items) == 1 {
-		endpoint = &endpoints.Items[0]
-	}
```
robbrockbank: I think you need to put a detailed comment here explaining exactly what you are doing here.

calico.go
```diff
@@ -44,80 +45,111 @@ func init() {
...
+				// Assign the WEP name to epIDs' WEPName field.
+				epIDs.WEPName = endpoint.Name
+				// Put the endpoint name from the matched WEP in the identifiers.
+				epIDs.Endpoint = ep.Spec.Endpoint
+				logger.Infof("Calico CNI found existing endpoint: %v", endpoint)
+				break
+			}
+		}
+
+		if endpoint == nil {
```
robbrockbank: Is this correct?

calico.go
```diff
@@ -109,6 +109,17 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	// try to see if the name matches the WEP identifiers we have.
+	// The identifiers we use for this match at this point are:
+	// 1. Node name
+	// 2. Orchestrator ID ('cni' or 'k8s')
+	// 3. ContainerID
+	// 4. Pod name
+	// Note we don't use the interface name (endpoint) for this match.
+	// If we find a match from the returned list then we've found the workload endpoint,
+	// and we reuse that even if it has a different interface name, because
+	// we only support one interface per pod right now.
```
robbrockbank: Hmmm, the question is - with this comment, would you have understood the subtleties of whats going on?

We can't hold this up for this comment, but I was expecting to see in this comment, all those nitty gritty details that I went through with you today.

calico_cni_k8s_test.go
```diff
@@ -1245,7 +1253,85 @@ var _ = Describe("CalicoCni", func() {
 					Expect(endpoints.Items[0].Spec.Endpoint).Should(Equal("eth0"))
 					Expect(endpoints.Items[0].Spec.ContainerID).Should(Equal(containerID))
 
-					_, err = testutils.DeleteContainer(netconf, contNs.Path(), name, testutils.K8S_TEST_NS)
+					// Now we create another pod with a very similar name.
```
robbrockbank: I think you should be calling out what you mean by "very similar".  Again - if you were reading this code for the first time would you get the subtleties of what's being tested here?

robbrockbank:
Does Casey want to give a final nod?

calico.go
gunjan5: Ok I've beefed up the comment with an example, PTAL

caseydavenport:
@gunjan5 I don't think this should get a release note - it's not a user-facing impact.

The ultimate release note will cover all the components (i.e. "Calico now supports etcdv3 blah blah")

k8s/k8s.go
```diff
@@ -23,48 +24,44 @@ import (
...
+	calicoclient "github.com/projectcalico/libcalico-go/lib/clientv2"
+	"github.com/projectcalico/libcalico-go/lib/options"
+	logrus "github.com/sirupsen/logrus"
 )
 
 // CmdAddK8s performs the "ADD" operation on a kubernetes pod
 // Having kubernetes code in its own file avoids polluting the mainline code. It's expected that the kubernetes case will
 // more special casing than the mainline code.
-func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoClient *calicoclient.Client, endpoint *api.WorkloadEndpoint) (*current.Result, error) {
+func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epIDs utils.EndpointIdentifiers, calicoClient calicoclient.Interface, endpoint *api.WorkloadEndpoint) (*current.Result, error) {
```
caseydavenport: wepIDs (and utils.EndpointIdentifiers should be renamed to utils.WEPIdentifiers)

k8s/k8s.go
```diff
@@ -240,17 +237,19 @@ func CmdAddK8s(args *skel.CmdArgs, conf utils.NetConf, nodename string, calicoCl
 
 		// Create the endpoint object and configure it.
 		endpoint = api.NewWorkloadEndpoint()
-		endpoint.Metadata.Name = args.IfName
-		endpoint.Metadata.Node = nodename
-		endpoint.Metadata.Orchestrator = orchestrator
-		endpoint.Metadata.Workload = workload
-		endpoint.Metadata.Labels = labels
+		endpoint.Name = epIDs.WEPName
```
caseydavenport: minor nit, but we're mixing up the Spec and Meta fields here - would be nice to put the Name and Namespace and Labels at the top and spec fields after.

k8s/k8s.go
```diff
@@ -321,25 +320,25 @@ func CmdDelK8s(c *calicoclient.Client, ep api.WorkloadEndpointMetadata, args *sk
 		// Check if ActiveInstanceID is populated (it will be an empty string "" if it was populated
 		// before this field was added to the API), and if it is there then compare it with ContainerID
 		// passed by the orchestrator to make sure they are the same, return without deleting if they aren't.
-	} else if wep.Metadata.ActiveInstanceID != "" && args.ContainerID != wep.Metadata.ActiveInstanceID {
+	} else if wep.Spec.ContainerID != "" && args.ContainerID != wep.Spec.ContainerID {
 		logger.WithField("WorkloadEndpoint", wep).Warning("CNI_ContainerID does not match WorkloadEndpoint ActiveInstanceID so ignoring the DELETE cmd.")
```
caseydavenport: nit (and not code you touched in this PR) but `CNI_ContainerID` is pretty bad - should be `ContainerID` or `CNI_CONTAINERID`

k8s/k8s.go
```diff
@@ -321,25 +320,25 @@ func CmdDelK8s(c *calicoclient.Client, ep api.WorkloadEndpointMetadata, args *sk
...
-	} else if err = c.WorkloadEndpoints().Delete(wep.Metadata); err != nil {
+	} else if _, err = c.WorkloadEndpoints().Delete(ctx, wep.Namespace, wep.Name, options.DeleteOptions{}); err != nil {
 		switch err := err.(type) {
 		case cerrors.ErrorResourceDoesNotExist:
 			// Log and proceed with the clean up if WEP doesn't exist.
 			logger.WithField("endpoint", wep).Info("Endpoint object does not exist, no need to clean up.")
 		case cerrors.ErrorResourceUpdateConflict:
 			// This case means the WEP object was modified between the time we did the Get and now,
-			// so it's not a safe Compare-and-Delete operation, so log and abort with the error.
+			// so it's not a safe Compare-and-Delete operation, so logrus and abort with the error.
```
caseydavenport: "so logrus and abort" - looks like a find and replace error :p

caseydavenport: Again, not sure why we decided to refactor that at this point.

calico.go
```diff
@@ -14,27 +14,28 @@
...
-	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/cni-plugin/types"
+	"github.com/projectcalico/cni-plugin/utils"
+	api "github.com/projectcalico/libcalico-go/lib/apiv2"
 	"github.com/projectcalico/libcalico-go/lib/errors"
 	"github.com/projectcalico/libcalico-go/lib/logutils"
-	cnet "github.com/projectcalico/libcalico-go/lib/net"
-	log "github.com/sirupsen/logrus"
+	"github.com/projectcalico/libcalico-go/lib/options"
+	logrus "github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
gunjan5: @robbrockbank suggested we call it `logrus`, I'm ok with either

calico.go
```diff
@@ -131,9 +183,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 			// Don't create the veth or do any networking.
 			// Just update the profile on the endpoint. The profile will be created if needed during the
 			// profile processing step.
-			fmt.Fprintf(os.Stderr, "Calico CNI appending profile: %s\n", profileID)
```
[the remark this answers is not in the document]
gunjan5: but as part of the `ConfigureLogging` func, we're setting the output to `os.Stderr` (`logrus.SetOutput(os.Stderr)`) already now, so does having a `fmt.Fprintf(os.Stderr, ...` really make any difference?

calico.go
```diff
@@ -44,80 +45,131 @@ func init() {
...
+		return err
+	}
 
-	logger.WithFields(log.Fields{
-		"Orchestrator": orchestrator,
-		"Node":         nodename,
-		"Workload":     workload,
-		"ContainerID":  args.ContainerID,
-	}).Info("Extracted identifiers")
+	// Remove the endpoint from the epIDs so we can get a WEP name prefix.
```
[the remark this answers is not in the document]
gunjan5: I've beefed up the comment and renamed epIDs to wepIDs

calico.go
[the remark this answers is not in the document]
caseydavenport: I'm not that bothered by calling it `logrus` - it's inconsistent with other places but not crazy.

We shouldn't alias it as itself though!

calico.go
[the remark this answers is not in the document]
caseydavenport: ̄\\\_( ͡° ͜ʖ ͡°)\_/ ̄

calico.go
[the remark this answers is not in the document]
gunjan5: ok aliasing it to itself was my bad, I'll fix it

PR #391, opened 2017-10-10
Retrieve named ports and add them to WEP

tmjd:
When using K8s policy, retrieve named ports from a K8s pod and attach
them to the WEP in the backend when doing an ADD.

## Description
With libcalico-go now having support for Named Ports they can be enabled in the CNI plugin

## Todos
- [X] Tests
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -28,6 +28,7 @@ import (
 	"github.com/projectcalico/cni-plugin/utils"
 	"github.com/projectcalico/libcalico-go/lib/api"
 	k8sbackend "github.com/projectcalico/libcalico-go/lib/backend/k8s"
+	backendconverter "github.com/projectcalico/libcalico-go/lib/converter"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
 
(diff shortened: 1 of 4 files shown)


k8s/k8s.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/projectcalico/cni-plugin/utils"
 	"github.com/projectcalico/libcalico-go/lib/api"
 	k8sbackend "github.com/projectcalico/libcalico-go/lib/backend/k8s"
+	backendConverter "github.com/projectcalico/libcalico-go/lib/converter"
```
caseydavenport: I think in general camelcase for imports is weird (you probably copied this from somewhere else, so we should fix that too!)

this should just be `backendconverter`

k8s/k8s.go
```diff
@@ -644,6 +653,28 @@ func getK8sLabelsAnnotations(client *kubernetes.Clientset, k8sargs utils.K8sArgs
 	return labels, pod.Annotations, nil
 }
 
+func getK8sPodPorts(client *kubernetes.Clientset, k8sargs utils.K8sArgs) ([]api.EndpointPort, error) {
+	var c k8sbackend.Converter
+	var bc backendConverter.WorkloadEndpointConverter
+
+	pod, err := client.Pods(string(k8sargs.K8S_POD_NAMESPACE)).Get(fmt.Sprintf("%s", k8sargs.K8S_POD_NAME), metav1.GetOptions{})
```
caseydavenport: Hm, this means we're now doing two queries for the same information in the k8s API.

I think we should combine this with the get labels + annotations function so that we only need to query for the pod once.

k8s/k8s.go
tmjd: yeah that's in kube-controllers/pkg/converter/pod_converter.go where I got it from.

PR #392, opened 2017-10-11
Update install script to allow conf name changes

weikinhuang:
## Description
This will allow users to upgrade the calico cni config to a different name while at the same time will clean up the old config files. This change is useful in the case where we want to move to a `.conflist` extension to enable additional plugins such as `portmap` during a upgrade.

This mirrors changes in https://github.com/coreos/flannel-cni/pull/5.

## Todos
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Optional environment variable `CNI_OLD_NAME` for `install-cni.sh` to clean up old configs when migrating to a different `CNI_CONF_NAME`.
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -128,7 +128,8 @@ sed -i s~__ETCD_CA_CERT_FILE__~${CNI_CONF_ETCD_CA:-}~g $TMP_CONF
 sed -i s~__ETCD_ENDPOINTS__~${ETCD_ENDPOINTS:-}~g $TMP_CONF
 sed -i s~__LOG_LEVEL__~${LOG_LEVEL:-warn}~g $TMP_CONF
 
-FILENAME=${CNI_CONF_NAME:-10-calico.conf}
+CNI_CONF_NAME=${CNI_CONF_NAME:-10-calico.conf}
+CNI_OLD_CONF_NAME=${CNI_OLD_CONF_NAME:-10-calico.conf}
 
 # Log the config file before inserting service account token.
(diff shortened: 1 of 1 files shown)


k8s-install/scripts/install-cni.sh
```diff
@@ -128,18 +128,23 @@ sed -i s~__ETCD_CA_CERT_FILE__~${CNI_CONF_ETCD_CA:-}~g $TMP_CONF
 sed -i s~__ETCD_ENDPOINTS__~${ETCD_ENDPOINTS:-}~g $TMP_CONF
 sed -i s~__LOG_LEVEL__~${LOG_LEVEL:-warn}~g $TMP_CONF
 
-FILENAME=${CNI_CONF_NAME:-10-calico.conf}
+CNI_CONF_NAME=${CNI_CONF_NAME:-10-calico.conf}
+CNI_OLD_NAME=${CNI_OLD_NAME:-10-calico.conf}
```
caseydavenport: One minor thing - I think it would be good to name the config option `CNI_OLD_CONF_NAME` to make it more clear it's referring to the old CNI configuration file name, not something else.

weikinhuang: sure thing!

weikinhuang:
Thanks!

weikinhuang:
will a release be cut for this? I see that this has a v2.6.2 milestone label.

caseydavenport:
@weikinhuang sorry for the long delay.

This exists now in Calico v3.0.0-beta1.

We also need to port this to v2.6.x, which is now looking like v2.6.4.

PR #397, opened 2017-10-23
Update for new label format and new profile naming format

robbrockbank:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -247,7 +247,7 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 		// If it's not, then just use the network name (which is the normal behavior)
 		// otherwise use one based on the Kubernetes pod's Namespace.
 		if conf.Policy.PolicyType == "k8s" {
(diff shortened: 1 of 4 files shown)


k8s/k8s.go
```diff
@@ -637,7 +637,7 @@ func getK8sPodInfo(client *kubernetes.Clientset, podName, podNamespace string) (
 		labels = make(map[string]string)
 	}
 
-	labels["calico/k8s_ns"] = podNamespace
+	labels[apiv2.LabelNamespace] = podNamespace
```
caseydavenport: This will get automatically added by libcalico, right?

I don't think we need this line here any more, though I'm happy to leave it since it won't hurt anything.

gunjan5:
Moving this PR over to https://github.com/projectcalico/cni-plugin/pull/398

k8s/k8s.go
robbrockbank: oh yes it will

PR #398, opened 2017-10-23
Update for new label format and new profile naming format

gunjan5:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


k8s/k8s.go
```diff
@@ -637,7 +637,7 @@ func getK8sPodInfo(client *kubernetes.Clientset, podName, podNamespace string) (
 		labels = make(map[string]string)
 	}
 
-	labels["calico/k8s_ns"] = podNamespace
+	labels[api.LabelNamespace] = podNamespace
```
robbrockbank: As Casey pointed out in the other review - this should not be required.

I think we should remove this line to ensure everything is working as expected.

robbrockbank: probably means the labels we return from this function should be pulled from the workload endpoint that we created.

gunjan5:
CI is being naughty, closing reopening this PR

PR #399, opened 2017-10-25
Set LabelsToApply on the CNI-created Profile

neiljerram:
## Description
I believe the intention with the CNI plugin is that all workloads with the same conf.Name should have connectivity to each other, in the absence of any subsequent Policy.

In order to achieve that, the Profile needs to have LabelsToApply: {conf.Name: ""}, so that each WorkloadEndpoint inherits that label.

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -313,8 +313,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 					Labels: map[string]string{conf.Name: ""},
 				},
 				Spec: api.ProfileSpec{
-					EgressRules:  []api.Rule{{Action: "allow"}},
-					IngressRules: inboundRules,
+					EgressRules:   []api.Rule{{Action: "allow"}},
+					IngressRules:  inboundRules,
+					LabelsToApply: map[string]string{conf.Name: ""},
 				},
 			}
 


gunjan5:
LGTM, thanks @neiljerram!

calico.go
```diff
@@ -313,8 +313,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 					Labels: map[string]string{conf.Name: ""},
```
caseydavenport: I think we don't need these labels any more.  They don't help or hurt, so maybe we can just leave them as-is, but might be good to clean up.

calico.go
```diff
@@ -313,8 +313,9 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 				},
 				Spec: api.ProfileSpec{
-					EgressRules:  []api.Rule{{Action: "allow"}},
-					IngressRules: inboundRules,
+					EgressRules:   []api.Rule{{Action: "allow"}},
```
caseydavenport: I think this should be switched to `api.Allow`

neiljerram:
@caseydavenport I'd be happy to make those two cleanups after 3.0 - or at least once all our CI is running on gas again - but inclined to leave well alone for now.  I'll make myself a JIRA task for this.

PR #402, opened 2017-10-26
[WiP] Skip write to datastore when using KDD.

fasaxc:
## Description

KDD doesn't support writes, but the write is not needed since the kubelet writes the IP allocation for us.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
@fasaxc see here: https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/resources/workloadendpoint.go#L46-L77

We do actually write the IP to work around an issue where the IP address doesn't make it into the API server fast enough.

fasaxc:
Sorry, meant to close this after you said the issue had already been fixed.

PR #408, opened 2017-11-02
backport bugfix: CNI plugin throws file exists while programming route

gunjan5:
Backport of the fix for 1253
Fixes https://github.com/projectcalico/calico/issues/1253

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
bugfix: CNI plugin throws file exists while programming route
```

### DIFF
--- a/utils/network.go
+++ b/utils/network.go
@@ -6,7 +6,7 @@ import (
 	"net"
 	"os"
 
-	"reflect"
+	"syscall"
 
 	"github.com/containernetworking/cni/pkg/ip"
 	"github.com/containernetworking/cni/pkg/ns"
@@ -188,7 +188,7 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
 	}
 
 	// Now that the host side of the veth is moved, state set to UP, and configured with sysctls, we can add the routes to it in the host namespace.
-	err = setupRoutes(hostVeth, result)
+	err = SetupRoutes(hostVeth, result)
 	if err != nil {
 		return "", "", fmt.Errorf("error adding host side routes for interface: %s, error: %s", hostVeth.Attrs().Name, err)
 	}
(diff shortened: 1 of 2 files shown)


utils/network.go
```diff
@@ -225,16 +223,18 @@ func setupRoutes(hostVeth netlink.Link, result *current.Result) error {
 				// exactly what we are intending to program.
 				// If the route we want is already there then most likely it's programmed by Felix, so we ignore it,
 				// and we return an error if none of the routes match the route we're trying to program.
+				log.Debugf("Constructed route: %v , scope: %v \n", route, route.Scope)
```
fasaxc: These debugs all have `\n` at the end; is that intentional?  Also, should we be using `log.WithField()` and friends?

utils/network.go
```diff
@@ -225,22 +223,24 @@ func setupRoutes(hostVeth netlink.Link, result *current.Result) error {
...
+					if r.LinkIndex == route.LinkIndex && r.Dst.IP.Equal(route.Dst.IP) && r.Scope == route.Scope {
 						// Route was already present on the host.
-						log.Infof("CNI skipping add route. Route already exists for %s\n", hostVeth.Attrs().Name)
+						log.WithFields(log.Fields{"interface": hostVeth.Attrs().Name}).Infof("CNI skipping add route. Route already exists")
 						return nil
 					}
 				}
-				return fmt.Errorf("route (Dst: %s, Scope: %v) already exists for an interface other than '%s'",
-					route.Dst.String(), route.Scope, hostVeth.Attrs().Name)
+				return fmt.Errorf("route (Ifindex: %d, Dst: %s, Scope: %s) already exists for an interface other than '%s'",
```
fasaxc: Why the change from `%v` to `%s` here and below; I think scope isn't a string so `%s` will fail.  `%v` always works so it's a sensible default unless you know you want something different.

gunjan5: good catch, fixed it for both PRs

PR #409, opened 2017-11-03
Handle libcalico-go returning non-IP'd WEPs

robbrockbank:
## Description
The libcalico-go behavior has been modified to return Pod-based WEPs (for KDD) even if they do not have any IPNetworks assigned.  The new requirement for it being a "Calico WEP" is not host-networked and a Node is assigned.

Minor tweaks to handle the fact that an endpoint will be returned in the list for the KDD scenario.

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -58,7 +58,7 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 
 	logger.Info("Extracted identifiers for CmdAddK8s")
 
-	if endpoint != nil {
+	if endpoint != nil && len(endpoint.Spec.IPNetworks) != 0 {
 		// This happens when Docker or the node restarts. K8s calls CNI with the same parameters as before.
 		// Do the networking (since the network namespace was destroyed and recreated).
 		// There's an existing endpoint - no need to create another. Find the IP address from the endpoint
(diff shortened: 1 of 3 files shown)


k8s/k8s.go
```diff
@@ -233,15 +235,17 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 			logger.Debugf("IPAM result set to: %+v", result)
 		}
 
-		// Create the endpoint object and configure it.
-		endpoint = api.NewWorkloadEndpoint()
-		endpoint.Name = epIDs.WEPName
-		endpoint.Namespace = epIDs.Namespace
+		// Configure the endpoint (creating if required).
+		if endpoint == nil {
```
caseydavenport: I think we only want to wrap the creation of the wep in this branch, but still set the other spec / metadata fields on the already existing wep.

All-in-all it doesn't really matter since we should only hit this with KDD, and KDD doesn't care about any of this stuff anyway, but I think that would be more correct should we eventually hit this code path with etcd in the future for some reason.

robbrockbank: not sure I understand that - unless you are saying the existing code is wrong too (irrespective of my libcalico-go change).

Change we chat?

caseydavenport: I guess my question is "why do we need the if statement here?"

Am I right in thinking that the WEP this code generates will be the same as the wep that already exists, if it does?

And, if the weps don't match, we want to update the existing wep to have the right fields?

robbrockbank: Ah gotcha.  Could any of those fields have changed?  I'm guessing the ContainerID possibly?

But yeah perhaps I should just have the creation of the new WEP in the if statement and move everything else outside.

caseydavenport:
@robbrockbank could you squash?

PR #413, opened 2017-11-09
Added CNI test args to allow cni to set WEP namespace from command line.

song-jiang:
Added CNI test args to allow cni to set WEP namespace from command line. This is useful to have multiple namespaces for non k8s cluster. Currently it is used for test purpose only and subject to change without prior notice.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -240,6 +240,15 @@ func GetIdentifiers(args *skel.CmdArgs, nodename string) (*WEPIdentifiers, error
 		epIDs.Pod = ""
 		// For any non-k8s orchestrator we set the namespace to default.
 		epIDs.Namespace = "default"
+
+		// Warning: CNITestArgs is used for test purpose only and subject to change without prior notice.
+		CNITestArgs := types.CNITestArgs{}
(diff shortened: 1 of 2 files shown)


utils/utils.go
```diff
@@ -208,6 +208,11 @@ func PopulateEndpointNets(wep *api.WorkloadEndpoint, result *current.Result) err
 	return nil
 }
 
+type CNITestArgs struct {
```
caseydavenport: I think this belongs in types.go alongside the k8s args - https://github.com/projectcalico/cni-plugin/blob/master/types/types.go#L89

utils/utils.go
```diff
@@ -208,6 +208,11 @@ func PopulateEndpointNets(wep *api.WorkloadEndpoint, result *current.Result) err
...
+	cnitypes.CommonArgs
```
caseydavenport: It probably also don't need the `CommonArgs` bit since we already unmarshal those elsewhere.

song-jiang: https://github.com/containernetworking/cni/blob/master/pkg/types/args.go#L54

It seems we have to embed this CommonArgs into the structure.

PR #416, opened 2017-11-09
Revving libcalico

heschlie:
## Description

Revving libcalico-go and fixing the tests for `install-cni.sh` as there were some changes to the containers package that are causing them to fail, so I've resorted to just calling the docker daemon directly from the tests as they are pretty simple anyway.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/k8s-install/scripts/install_cni_test.go
+++ b/k8s-install/scripts/install_cni_test.go
@@ -3,13 +3,12 @@ package scripts_test
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"github.com/projectcalico/felix/fv/containers"
 
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


k8s-install/scripts/install_cni_test.go
```diff
@@ -18,14 +17,25 @@ var secretFile = "tmp/serviceaccount/token"
...
 func runCniContainer(extraArgs ...string) error {
+	name := "cni"
+
 	// Get the CWD for mounting directories into container.
 	cwd, err := os.Getwd()
 	if err != nil {
 		Fail("Could not get CWD")
 	}
 
+	// Ensure the was not left over
```
robbrockbank: ensure the what?

PR #420, opened 2017-11-14
Copy File Error, Please check whether disabled selinux on your host

jiulongzaitian:
Signed-off-by: zhangjie <zhangjie0619@yeah.net>

## Description

we can not write file in container source volume when  host selinux is not disable .
so if we did not disable selinux, when cni_install container started, i will occurred following errors:

```

CNI plugin version: 
cp: can't create '/host/opt/cni/bin/calico': Permission denied
cp: can't create '/host/opt/cni/bin/calico-ipam': Permission denied
cp: can't create '/host/opt/cni/bin/flannel': File exists
cp: can't create '/host/opt/cni/bin/host-local': File exists
cp: can't create '/host/opt/cni/bin/loopback': File exists
cp: can't create '/host/opt/cni/bin/portmap': Permission denied
/install-cni.sh: line 83: /host/opt/cni/bin/calico: not found
         "k8s_api_root": "https://10.96.0.1:__KUBERNETES_SERVICE_PORT__",
CNI config: {
    "name": "k8s-pod-network",
    "cniVersion": "0.1.0",
    "type": "calico",
    "etcd_endpoints": "http://10.*.*.136:6666",
    "log_level": "info",
    "mtu": 1500,
    "ipam": {
        "type": "calico-ipam"
    },
    "policy": {
        "type": "k8s",
         "k8s_api_root": "https://10.*.0.1:443",
         "k8s_auth_token": "__SERVICEACCOUNT_TOKEN__"
    },
    "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
    }
}
Created CNI config 10-calico.conf
/install-cni.sh: line 98: can't create /host/etc/cni/net.d/calico-kubeconfig: Permission denied
mv: can't create '/host/etc/cni/net.d/10-calico.conf': Permission denied

```

this pr will give some tips when copy file error.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note

```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -77,6 +77,11 @@ do
 			continue
 		fi
 		cp $path $dir/
+		if [ "$?" != "0" ];
+		then
+			echo "Failed to copy $path to $dir. This may be caused by selinux configuration on the host, or something else."
+			exit 1
+		fi
 	done
 
 	echo "Wrote Calico CNI binaries to $dir"
@@ -143,6 +148,11 @@ if [ "${CNI_CONF_NAME}" != "${CNI_OLD_CONF_NAME}" ]; then
 fi
 # Move the temporary CNI config into place.
 mv $TMP_CONF /host/etc/cni/net.d/${CNI_CONF_NAME}
+if [ "$?" != "0" ];
+then
+	echo "Failed to mv files. This may be caused by selinux configuration on the host, or something else."
+	exit 1
+fi
 
 echo "Created CNI config ${CNI_CONF_NAME}"
 


k8s-install/scripts/install-cni.sh
```diff
@@ -77,6 +77,10 @@ do
 			continue
 		fi
 		cp $path $dir/
+		if [ "$?" != "0" ];
+		then
+			echo "Copy File Error, Please check whether disabled selinux on your host"
```
caseydavenport: I think we should tweak the message so that it doesn't mention selinux, since there may be copy errors for other reasons.

For example:

```
echo "Failed to copy $path to $dir. This may be caused by selinux configuration on the host, or something else."
```

We probably also want to `exit 1` right after to indicate a failure.

jiulongzaitian: if use `exit 1`, kubernetes will constantly restart the pod , Is it ok ?

caseydavenport: Eventually it will hit `CrashLoopBackoff`.

My thinking was it's better to get notified of this than for it to fail silently - WDYT?

jiulongzaitian: OK， it`s better.

k8s-install/scripts/install-cni.sh
```diff
@@ -77,6 +77,11 @@ do
 			continue
 		fi
 		cp $path $dir/
+		if [ "$?" != "0" ];
+		then
+            echo "Failed to copy $path to $dir. This may be caused by selinux configuration on the host, or something else."
```
caseydavenport: Last thing - intendation looks wrong here. Could you indent this to match?

thanks!

jiulongzaitian: sorry for  such mistake

PR #427, opened 2017-11-16
Rev libcalico go to master

fasaxc:
## Description

Continuation of #421 

Rev libcalico-go to master and fix up tests.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -95,7 +95,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	}
 
 	var logger *logrus.Entry
-	if wepIDs.Orchestrator == "k8s" {
+	if wepIDs.Orchestrator == api.OrchestratorKubernetes {
 		logger = logrus.WithFields(logrus.Fields{
 			"WorkloadEndpoint": fmt.Sprintf("%s%s", wepPrefix, wepIDs.Endpoint),
 			"ContainerID":      wepIDs.ContainerID,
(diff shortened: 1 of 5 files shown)


calico.go
```diff
@@ -316,9 +316,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 			fmt.Fprintf(os.Stderr, "Calico CNI creating profile: %s\n", conf.Name)
 			var inboundRules []api.Rule
 			if wepIDs.Orchestrator == "k8s" {
```
robbrockbank: nit (and out of scope):  would be nice to update this to use apis.v3.OrchestratorKubernetes

but wouldn't let that stop us merging.

calico_cni_k8s_test.go
```diff
@@ -1055,8 +1055,8 @@ var _ = Describe("CalicoCni", func() {
 					// Now create a K8s pod.
 					name := fmt.Sprintf("run%d", rand.Uint32())
 
-					cniContainerIDX := "container-id-00X"
```
caseydavenport: @robbrockbank is there a reason we need to limit container IDs to be lowercase?

Probably fine, but might bite us some day.

robbrockbank: container ID is used in the WEP name construction.  We could certainly make the fields more permissive and just have the name construction lowercase it.

Issue #340, opened 2017-06-02
Is it possible to use a certain Hardware address(MAC) when generating veths in cni?


randmonkey:
Since there is an issue in udev https://github.com/systemd/systemd/issues/3374, it is possible that cni plugin fails to create veth pairs, that causes failures in configuring network of containers. In this udev issue we see that we can set a MAC when creating veths to avoid this bug. So, I would like to try to set MAC for veths in CNI, as a possible solution of https://github.com/projectcalico/calico/issues/763. I would like to get some suggestions on feasiblity and implemetation of these things.  
## Expected Behavior

Hardware addresses(MAC) of vethes in containers are set by cni 
## Current Behavior

CNI does not set MAC when creating veths, leaving it to systemd/udevd.
## Possible Solution

generate a hardware address from IP allocation result, and use the MAC in LinkAttr when calling netlink.LinkAdd to add veths in func DoNetworking in utils/network.go.

## Context

My team deployed a k8s + calico cluster, and we saw that a few containers has no network. We checked network interfaces, and did not find the veths for those containers. Then, we checked system logs and confirmed the failure is caused by https://github.com/systemd/systemd/issues/3374. So, I got an idea to set a MAC in CNI to avoid this udev issue.
## Your Environment

* CNI plugin version (e.g. `/opt/cni/bin/calico --version`): 1.6.1 (behavior still the same in the newest release 1.8.3)
* Orchestrator version (e.g. kubernetes, mesos, rkt): kunernetes
* CNI config:
```
{
    "name": "k8s-pod-network",
    "type": "calico",
    "etcd_endpoints": "http://x.x.x.x:2379",
    "etcd_key_file": "",
    "etcd_cert_file": "",
    "etcd_ca_cert_file": "",
    "log_level": "info",
    "ipam": {
        "type": "calico-ipam"
    },
    "policy": {
        "type": "k8s",
        "k8s_api_root": "https://x.x.x.x:443",
        "k8s_auth_token": "================================="
    },
    "kubernetes": {
        "kubeconfig": "/etc/cni/net.d/calico-kubeconfig"
    }
}
```
* Operating System and version: Ubuntu 16.04.2 LTS (GNU/Linux 4.4.0-62-generic x86_64)
* Link to your project (optional):

## Possible concerns 
What if the CNI generated a MAC duplicating to other virtual interfaces? can we avoid that?

tmjd:
[...]

PR #432, opened 2017-11-21
Generate MAC addr from container ID

tmjd:

### DIFF
--- a/utils/network.go
+++ b/utils/network.go
@@ -59,6 +59,16 @@ func DoNetworking(args *skel.CmdArgs, conf types.NetConf, result *current.Result
 			return err
 		}
 
+		if mac, err := net.ParseMAC("EE:EE:EE:EE:EE:EE"); err != nil {
+			logger.Infof("failed to parse MAC Address: %v. Using kernel generated MAC.", err)
+		} else {
+			// Set the MAC address on the host side interface so the kernel does not
+			// have to generate a persistent address which fails some times.
+			if err = netlink.LinkSetHardwareAddr(hostVeth, mac); err != nil {
+				logger.Warnf("failed to Set MAC of %q: %v. Using kernel generated MAC.", hostVethName, err)
+			}
+		}
+
 		// Explicitly set the veth to UP state, because netlink doesn't always do that on all the platforms with net.FlagUp.
 		// veth won't get a link local address unless it's set to UP state.
 		if err = netlink.LinkSetUp(hostVeth); err != nil {
--- a/calico_cni_test.go
+++ b/calico_cni_test.go
@@ -62,8 +62,6 @@ var _ = Describe("CalicoCni", func() {
 					log.Fatalf("Error getting result from the session: %v\n", err)
 				}
 
-				mac := contVeth.Attrs().HardwareAddr
-
 				Expect(len(result.IPs)).Should(Equal(1))
 				ip := result.IPs[0].Address.IP.String()
 				result.IPs[0].Address.IP = result.IPs[0].Address.IP.To4() // Make sure the IP is respresented as 4 bytes
@@ -96,6 +94,8 @@ var _ = Describe("CalicoCni", func() {
 				Expect(endpoints.Items[0].Name).Should(Equal(wrkload))
 				Expect(endpoints.Items[0].Namespace).Should(Equal(testutils.TEST_DEFAULT_NS))
 
+				mac := contVeth.Attrs().HardwareAddr
+
 				Expect(endpoints.Items[0].Spec).Should(Equal(api.WorkloadEndpointSpec{
 					InterfaceName: fmt.Sprintf("cali%s", containerID),
 					IPNetworks:    []string{result.IPs[0].Address.String()},
@@ -116,6 +116,7 @@ var _ = Describe("CalicoCni", func() {
 				Expect(err).ToNot(HaveOccurred())
 				Expect(hostVeth.Attrs().Flags.String()).Should(ContainSubstring("up"))
 				Expect(hostVeth.Attrs().MTU).Should(Equal(1500))
+				Expect(hostVeth.Attrs().HardwareAddr.String()).Should(Equal("ee:ee:ee:ee:ee:ee"))
 
 				// Assert hostVeth sysctl values are set to what we expect for IPv4.
 				err = testutils.CheckSysctlValue(fmt.Sprintf("/proc/sys/net/ipv4/conf/%s/proxy_arp", hostVethName), "1")


utils/network.go
```diff
@@ -28,7 +30,9 @@ func DoNetworking(args *skel.CmdArgs, conf types.NetConf, result *current.Result
 		hostVethName = desiredVethName
 	}
 
-	logger.Infof("Setting the host side veth name to %s", hostVethName)
+	mac := genLocalMacAddr(args.ContainerID)
+
+	logger.Errorf("Setting the host side veth name to %s with MAC %s from containerID %s", hostVethName, mac, args.ContainerID)
```
bcreane: do you want `Infof` here?

utils/network.go
```diff
@@ -340,3 +345,41 @@ func writeProcSys(path, value string) error {
...
+
+		var m string
+		m = ""
+		// Go through the next bytes until a hex one is found
+		for ; j < srcLen && m == ""; j++ {
+			// Only use the byte if the decode succeeds
+			if _, err := hex.DecodeString("0" + srcBytes[j]); err == nil {
+				m = srcBytes[j]
+			} else {
+				logrus.Infof("Convert %s failed with %v", srcBytes[j], err)
```
bcreane: looks like the import is `log`, not `logrus`.

utils/network.go
```diff
@@ -340,3 +345,41 @@ func writeProcSys(path, value string) error {
...
+			}
+		}
+		if m == "" {
+			m = "0"
+		}
+		macString += m
```
bcreane: Since you're appending rather than prepending, the "02" above is going to be at the beginning of the mac address which doesn't work for locally administered address since it'll be the MSB rather than the LSB.

utils/network.go
```diff
@@ -340,3 +345,41 @@ func writeProcSys(path, value string) error {
 	}
 	return err
 }
+
+// Generate a mac address from the passed in string
+func genLocalMacAddr(src string) net.HardwareAddr {
```
bcreane: would be nice to see some UTs to ensure correct behavior (and resilience in the face of unexpected input) of `genLocalMacAddr`. Some nice test cases:
* length == 0
* invalid characters
* fewer characters (or non-zero characters) than are present in a mac address

utils/network.go
```diff
@@ -340,3 +345,41 @@ func writeProcSys(path, value string) error {
...
+	// The 2nd bit set of the first octet makes this MAC address a "locally administered address"
+	// https://en.wikipedia.org/wiki/MAC_address
+	macString := "02"
```
bcreane: The 2 is just one of 6 remaining bits (the first bit is overloaded for uni/multicast). I'm not sure how big the namespace needs to be for these generated mac addresses, but potentially we'll need those extra 6 bits.

utils/network.go
tmjd: Yep, good catch

utils/network.go
tmjd: The rest of the octets leaves a space of 2^40 so I don't think it would be a big deal.  Especially since these don't really need to be unique.  We could just set them all the same and still be fine.  So I don't think the additional complication of using those 6 bits is worth it.

utils/network.go
tmjd: The way I read the wiki article I linked I think setting that bit is correct. It says `Universally administered and locally administered addresses are distinguished by setting the second-least-significant bit of the first octet of the address`  and even gives the example of `06-00-00-00-00-00` as being a "locally administered address".

utils/network.go
tmjd: I don't think it is worth while to log that out, that was a debug thing that I forgot to remove.  Any problem with just dropping that?

utils/network.go
bcreane: You're right, my mistake. Its the 2nd bit of the first (not the 6th) octet which is what I (mis)read.

utils/network.go
bcreane: nope, that's fine.

utils/network.go
bcreane: okay. but I was looking forward to seeing bitwise operators in golang. maybe next time.

utils/network.go
caseydavenport: @tmjd I'm sorry to backtrack here, but I think we should do what @fasaxc suggested here: https://github.com/projectcalico/cni-plugin/issues/340#issuecomment-339349037

Using a fixed mac will reduce a lot of complexity here and should behave just the same, without the scariness of using the container ID to generate the address.

tmjd: Should I track down what is used in OpenStack or just use "EE:EE:EE:EE:EE:EE" like libnetwork uses?

tmjd:
Ok swapped to use a fixed MAC address and setting it on the host side of the Veth instead of the Container side.

I wasn't sure if setting the MAC would help with the problem this is trying to address because the interface is created and then the MAC is set on it.  Looking at journalctl on my test Vagrant cluster, the logs showed that it did clear up the `Could not generate persistent MAC address for cali756af06e782: No such file or directory` errors that were happening previously.

PTAL @caseydavenport and @bcreane

tmjd:
Addresses #340

k8s/k8s.go
```diff
@@ -300,7 +300,10 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 	endpoint.Spec.MAC = mac.String()
 	endpoint.Spec.InterfaceName = hostVethName
 	endpoint.Spec.ContainerID = epIDs.ContainerID
-	logger.WithField("endpoint", endpoint).Info("Added Mac, interface name, and active container ID to endpoint")
+	logger.WithFields(logrus.Fields{
+		"endpoint": endpoint,
+		"mac":      mac.String(),
```
caseydavenport: I don't think we need this change - the endpoint already has the mac set, so it gets logged out already.

caseydavenport: See line 300

utils/network.go
```diff
@@ -28,7 +28,7 @@ func DoNetworking(args *skel.CmdArgs, conf types.NetConf, result *current.Result
 		hostVethName = desiredVethName
 	}
 
-	logger.Infof("Setting the host side veth name to %s", hostVethName)
+	logger.Infof("Setting the host side veth name to %s from containerID %s", hostVethName, args.ContainerID)
```
caseydavenport: Is this useful? I'd rather leave this log as is, unless there's a good reason.

utils/network.go
```diff
@@ -59,6 +59,16 @@ func DoNetworking(args *skel.CmdArgs, conf types.NetConf, result *current.Result
 			return err
 		}
 
+		mac, err := net.ParseMAC("EE:EE:EE:EE:EE:EE")
+		if err != nil {
+			return fmt.Errorf("failed to parse MAC Address: %v", err)
+		}
+		err = netlink.LinkSetHardwareAddr(hostVeth, mac)
+		if err != nil {
+			err = fmt.Errorf("failed to Set MAC of host%q: %v", hostVethName, err)
```
caseydavenport: I think you're missing a space - `host%q`

utils/network.go
```diff
@@ -59,6 +59,16 @@ func DoNetworking(args *skel.CmdArgs, conf types.NetConf, result *current.Result
 			return err
 		}
 
+		if mac, err := net.ParseMAC("EE:EE:EE:EE:EE:EE"); err != nil {
+			logger.Infof("failed to parse MAC Address: %v. Using kernel generated MAC.", err)
+		} else {
+			// Set the MAC address on the host side interface so the kernel does not
+			// have to generate a persistent address which fails some times.
+			if err = netlink.LinkSetHardwareAddr(hostVeth, mac); err != nil {
+				logger.Infof("failed to Set MAC of %q: %v. Using kernel generated MAC.", hostVethName, err)
```
caseydavenport: I think this one should probably be a Warning log

bcreane: nit: should probably be `set` rather than `Set`

PR #436, opened 2017-11-27
Set MAC on host side of veth pair

tmjd:
- Applied changes made on v2.0-beta (Calico v3.0-beta)
- Setting the MAC address avoids issues when kernel fails to generate a
  persistent MAC address

### DIFF
--- a/utils/network.go
+++ b/utils/network.go
@@ -57,6 +57,16 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
 			return err
 		}
 
+		if mac, err := net.ParseMAC("EE:EE:EE:EE:EE:EE"); err != nil {
+			logger.Infof("failed to parse MAC Address: %v. Using kernel generated MAC.", err)
+		} else {
+			// Set the MAC address on the host side interface so the kernel does not
+			// have to generate a persistent address which fails some times.
+			if err = netlink.LinkSetHardwareAddr(hostVeth, mac); err != nil {
+				logger.Warnf("failed to Set MAC of %q: %v. Using kernel generated MAC.", hostVethName, err)
+			}
+		}
+
 		// Explicitly set the veth to UP state, because netlink doesn't always do that on all the platforms with net.FlagUp.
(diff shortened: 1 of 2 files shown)


tmjd:
I put `hold-merge` on since this I am assuming we don't want to add this to the v2.6.3 release which is on the verge of being cut.

tmjd:
Need to make sure the FAQ entry in https://github.com/projectcalico/calico/pull/1394 is added to the v2.6 docs when this PR is ready for to go into a v2.6 release

utils/network.go
```diff
@@ -57,6 +57,18 @@ func DoNetworking(args *skel.CmdArgs, conf NetConf, result *current.Result, logg
 			return err
 		}
 
+		mac, err := net.ParseMAC("EE:EE:EE:EE:EE:EE")
+		if err != nil {
+			return fmt.Errorf("failed to parse MAC Address: %v", err)
```
caseydavenport: Hm, wondering if we should return an error in these cases, or just _log_ an error.

thought process is - we don't do this today and it works on most systems, should we just fall back to today's behavior if for some reason we fail to set the mac on a given system?

caseydavenport: @tmjd @fasaxc WDYT?

tmjd: Makes sense to me.  Just log an error and only do LinkSetHardwareAddr if it is successful.

caseydavenport: Yeah, and if `LinkSetHardwareAddr` fails then we carry on as normal (but log an error).

fasaxc: Sounds fine to me.  (TBH, I'd be tempted to panic if parsing a fixed MAC fails.  That should work 100% of the time so the most likely scenario for that to fail is if we accidentally introduced a typo into the MAC during maintenance.)

tmjd:
I have cherry-picked the change from 0038728e5e0055c85037bbf7064fb21a507c6064 that was made against the master branch so I think this is ready to go.
I also removed the `hold-merge` label since v2.6.3 has been released so I think this can go into the v1.11.x-series branch now.

PR #438, opened 2017-11-29
Fix detection of hostname to use lowercase

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -41,3 +41,5 @@ import:
   - fv.containers
 - package: github.com/coreos/etcd
   version: v3.2.7
+- package: golang.org/x/crypto
+  version: 1351f936d976c60a0a48d728281922cf63eafb8d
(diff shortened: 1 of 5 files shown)


gunjan5:
@caseydavenport seems like the linter is linting the vendor directory?

caseydavenport:
hmmm

caseydavenport:
@gunjan5 not sure what was going on... I've pinned the offending package for now, but we're passing `--vendor` so not sure why it was even checking it...

glide.yaml
```diff
@@ -41,3 +41,5 @@ import:
   - fv.containers
 - package: github.com/coreos/etcd
   version: v3.2.7
+- package: golang.org/x/crypto
+  version: 1351f936d976c60a0a48d728281922cf63eafb8d
```
gunjan5: I think this version causes issues with calicoctl (especially /x/sys windows package) but should be fine in CNI. This pin doesn't seem relevant to this PR, was it conflicting with glide up?

gunjan5: Oh I see your comment about --vendor now.

PR #440, opened 2017-11-30
error out if Ready flag is not set (v1.11.x)

tmjd:
Here is a quick shot at checking the Ready flag before doing any operations.

Todo:
- Add some tests
- Maybe rework if we want to expose the ready flag through the Client interface instead of using a Get to the Backend

```release-note
None required
```

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -92,6 +92,15 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	ready, err := IsReady(calicoClient)
+	if err != nil {
+		return err
+	}
+	if !ready {
+		logger.Warn("Upgrade may be in progress, ready flag is not set")
+		return fmt.Errorf("Calico is currently not ready to process requests")
+	}
+
 	// Always check if there's an existing endpoint.
 	endpoints, err := calicoClient.WorkloadEndpoints().List(api.WorkloadEndpointMetadata{
 		Node:         nodename,
@@ -331,6 +340,15 @@ func cmdDel(args *skel.CmdArgs) error {
 	calicoClient, err := CreateClient(conf)
 	if err != nil {
 		return err
+	}
+
+	ready, err := IsReady(calicoClient)
+	if err != nil {
+		return err
+	}
+	if !ready {
+		logger.Warn("Upgrade may be in progress, ready flag is not set")
+		return fmt.Errorf("Calico is currently not ready to process requests")
 	}
 
 	wep := api.WorkloadEndpointMetadata{
--- a/calico_cni_test.go
+++ b/calico_cni_test.go
@@ -17,6 +17,7 @@ import (
 	. "github.com/projectcalico/cni-plugin/test_utils"
 	"github.com/projectcalico/cni-plugin/utils"
 	"github.com/projectcalico/libcalico-go/lib/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/client"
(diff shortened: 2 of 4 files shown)


utils/utils.go
```diff
@@ -263,6 +264,15 @@ func CreateClient(conf NetConf) (*client.Client, error) {
 	return calicoClient, nil
 }
 
+func IsReady(client *client.Client) bool {
```
gunjan5: Can you add a comment explaining the purpose of this, so people don't wonder what this is for in future

utils/utils.go
```diff
@@ -263,6 +264,15 @@ func CreateClient(conf NetConf) (*client.Client, error) {
...
+	kvPair, err := client.Backend.Get(model.ReadyFlagKey{})
+	if err != nil {
+		return true
+	}
+
+	return kvPair.Value.(bool)
```
gunjan5: I think we should check the interface assertion, because someone can manually set the value to something and all the CNI plugins will start panicing.

utils/utils.go
```diff
@@ -263,6 +264,15 @@ func CreateClient(conf NetConf) (*client.Client, error) {
 	return calicoClient, nil
 }
 
+func IsReady(client *client.Client) bool {
+	kvPair, err := client.Backend.Get(model.ReadyFlagKey{})
+	if err != nil {
+		return true
```
gunjan5: Maybe we should log a line here maybe a warning level saying can't get to the datastore or something. WDYT?

calico_cni_test.go
```diff
@@ -200,6 +201,36 @@ var _ = Describe("CalicoCni", func() {
...
+				It("should return error on Add", func() {
+					kv, err := calicoClient.Backend.Apply(
+						&model.KVPair{
+							Key:   model.ReadyFlagKey{},
+							Value: false,
+						})
+					Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("%v", kv))
+					_, session, _, _, _, _, err := CreateContainer(netconf, "", "")
+					Expect(err).Should(HaveOccurred())
+					Eventually(session).ShouldNot(gexec.Exit(0))
```
gunjan5: This works, but it looks a bit weird using `ShouldNot` with `Eventually`, because `Eventually` only checks for the condition for a certain amount of time (I think it's 1 second by default) and if it's set to `ShouldNot` then it's possible the plugin exits with exit code 0 but we still don't catch it. I suggest asserting the expected error code (code `1`?) with `Eventually`

calico.go
```diff
@@ -92,6 +92,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	if ready := IsReady(calicoClient); !ready {
+		return fmt.Errorf("Calico is currently not ready to process requests")
```
gunjan5: What do you think about adding a line saying, upgrade might be in progress, so users don't wonder why it's not ready

gunjan5: OR maybe a warning log before the return line saying "upgrade might be in progress" or something

utils/utils.go
caseydavenport: Yeah, there are sort of three cases here:
- datastore is reachable, not ready
- datastore is reachable, ready
- datastore is not reachabe

We want to make sure the correct logging and errors returned for each. We might want to change the signature of this function to return an error, so we can properly bubble up the connection error.

tmjd:
Should this include a release note?

gunjan5:
@tmjd not sure. @caseydavenport ?

PR #443, opened 2017-12-01
If ready flag is not true then do not proccess ADD or DEL

tmjd:
```release-note
None required
```

### DIFF
--- a/calico_cni_test.go
+++ b/calico_cni_test.go
@@ -197,6 +197,59 @@ var _ = Describe("CalicoCni", func() {
 					Expect(err).ShouldNot(HaveOccurred())
 				})
 			})
+
+			Context("when ready flag is false", func() {
+				It("errors when ADD is done", func() {
+					ci, err := calicoClient.ClusterInformation().Get(ctx, "default", options.GetOptions{})
+					Expect(err).ShouldNot(HaveOccurred())
+					r := false
+					ci.Spec.DatastoreReady = &r
+					ci, err = calicoClient.ClusterInformation().Update(ctx, ci, options.SetOptions{})
+					Expect(err).ShouldNot(HaveOccurred())
+					_, session, _, _, _, _, err := testutils.CreateContainer(netconf, "", testutils.TEST_DEFAULT_NS, "")
+					Expect(err).Should(HaveOccurred())
+					Eventually(session).Should(gexec.Exit(1))
+				})
+
+				It("errors when DEL is done", func() {
+					_, session, _, _, _, contNs, err := testutils.CreateContainer(netconf, "", testutils.TEST_DEFAULT_NS, "")
+					Expect(err).ShouldNot(HaveOccurred())
+					Eventually(session).Should(gexec.Exit(0))
+
+					ci, err := calicoClient.ClusterInformation().Get(ctx, "default", options.GetOptions{})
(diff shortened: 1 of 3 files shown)


testutils/utils.go
```diff
@@ -59,6 +61,22 @@ func WipeEtcd() {
...
+	calicoClient, _ := client.NewFromEnv()
+	ctx := context.Background()
+	newClusterInfo := api.NewClusterInformation()
+	newClusterInfo.Name = "default"
+	newClusterInfo.Spec.CalicoVersion = "v0.0.0"
+	newClusterInfo.Spec.ClusterType = "test"
+	newClusterInfo.Spec.ClusterGUID = fmt.Sprintf("%s", hex.EncodeToString(uuid.NewV4().Bytes()))
+	datastoreReady := true
+	newClusterInfo.Spec.DatastoreReady = &datastoreReady
+	ci, err := calicoClient.ClusterInformation().Create(ctx, newClusterInfo, options.SetOptions{})
```
gunjan5: you can just put `context.Background()` here since the `ctx` variable is not used anywhere else

testutils/utils.go
```diff
@@ -59,6 +61,22 @@ func WipeEtcd() {
...
 	_ = be.Clean()
+
+	// Set the ready flag so calls to the CNI plugin can proceed
+	calicoClient, _ := client.NewFromEnv()
+	ctx := context.Background()
+	newClusterInfo := api.NewClusterInformation()
+	newClusterInfo.Name = "default"
+	newClusterInfo.Spec.CalicoVersion = "v0.0.0"
+	newClusterInfo.Spec.ClusterType = "test"
+	newClusterInfo.Spec.ClusterGUID = fmt.Sprintf("%s", hex.EncodeToString(uuid.NewV4().Bytes()))
```
gunjan5: I don't think we need to populate any other `Spec` fields since they're not used in CNI. Especially this one, it'll save us an extra import

calico_cni_test.go
```diff
@@ -197,6 +197,59 @@ var _ = Describe("CalicoCni", func() {
...
+					ci, err := calicoClient.ClusterInformation().Get(ctx, "default", options.GetOptions{})
+					Expect(err).ShouldNot(HaveOccurred())
+					r := false
+					ci.Spec.DatastoreReady = &r
+					ci, err = calicoClient.ClusterInformation().Update(ctx, ci, options.SetOptions{})
+					Expect(err).ShouldNot(HaveOccurred())
+
+					exitCode, err := testutils.DeleteContainer(netconf, contNs.Path(), "", testutils.TEST_DEFAULT_NS)
+					Expect(err).ShouldNot(HaveOccurred())
+					Expect(exitCode).ShouldNot(Equal(0))
```
gunjan5: Nit: I think you forgot to change this line to be should exit 1

tmjd: Since this test is using Expect (not Eventually) I went with what I really wanted to test. It is testing that the result is not success so I just wanted to make sure the exist code is not 0.

tmjd: In other words I don't care about the value except that it is not zero.

gunjan5: Oh sorry, didn't see it si not eventually

PR #445, opened 2017-12-07
Changing the handle we use for IPAM

heschlie:
## Description

Instead of just using the workload, we use the container network
namespace and the container ID combined to generate a unique handle to
track allocations.

Some questions before the WIP is removed:
- Should we be trimming off the path portion of the container network ns? It seems like so but I haven't done that yet.
- There was mention of separate code paths for k8s vs others, and it seems like they both lead back to `calico-ipam.go` so I feel like I've probably overlooked something in the code path for CNI
- If we trim out the path portion of the container net ns (and maybe even if not) I don't think there is any escaping that needs to be done
- Should we bother to check in `utils.GetHandleId()` for the container ID and network namespace as they are supposed to be a required part of the CNI spec (provided the orchestrators are all following it properly)?

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -969,7 +969,8 @@ var _ = Describe("CalicoCni", func() {
 
 				checkIPAMReservation := func() {
 					// IPAM reservation should still be in place.
-					ipamIPs, err := calicoClient.IPAM().IPsByHandle(workloadName)
+					handleID, _ := utils.GetHandleID("net8", containerID, containerID)
+					ipamIPs, err := calicoClient.IPAM().IPsByHandle(handleID)
 					ExpectWithOffset(0, err).NotTo(HaveOccurred())
 					ExpectWithOffset(0, ipamIPs).To(HaveLen(1),
 						"There should be an IPAM handle for endpoint")
--- a/ipam/calico-ipam.go
+++ b/ipam/calico-ipam.go
@@ -101,7 +101,15 @@ func cmdAdd(args *skel.CmdArgs) error {
 	if err != nil {
 		return err
 	}
-	logger := utils.CreateContextLogger(workloadID)
+
+	handleID, err := utils.GetHandleID(conf.Name, args.ContainerID, workloadID)
+	if err != nil {
+		return err
+	}
+	logger := log.WithFields(log.Fields{
+		"workloadID": workloadID,
+		"handleID":   handleID,
+	})
 
 	ipamArgs := ipamArgs{}
 	if err = types.LoadArgs(args.Args, &ipamArgs); err != nil {
@@ -113,7 +121,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 		fmt.Fprintf(os.Stderr, "Calico CNI IPAM request IP: %v\n", ipamArgs.IP)
 
 		// The hostname will be defaulted to the actual hostname if conf.Nodename is empty
-		assignArgs := client.AssignIPArgs{IP: cnet.IP{ipamArgs.IP}, HandleID: &workloadID, Hostname: nodename}
+		assignArgs := client.AssignIPArgs{IP: cnet.IP{ipamArgs.IP}, HandleID: &handleID, Hostname: nodename}
 		logger.WithField("assignArgs", assignArgs).Info("Assigning provided IP")
 		err := calicoClient.IPAM().AssignIP(assignArgs)
 		if err != nil {
@@ -169,7 +177,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 		assignArgs := client.AutoAssignArgs{
 			Num4:      num4,
 			Num6:      num6,
-			HandleID:  &workloadID,
+			HandleID:  &handleID,
 			Hostname:  nodename,
 			IPv4Pools: v4pools,
 			IPv6Pools: v6pools,
@@ -222,24 +230,44 @@ func cmdDel(args *skel.CmdArgs) error {
 		return err
 	}
 
-	// Release the IP address by using the handle - which is workloadID.
 	workloadID, _, err := utils.GetIdentifiers(args)
 	if err != nil {
 		return err
 	}
 
-	logger := utils.CreateContextLogger(workloadID)
-
-	logger.Info("Releasing address using workloadID")
-	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
-		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
-			logger.WithField("workloadId", workloadID).Warn("Asked to release address but it doesn't exist. Ignoring")
-			return nil
-		}
+	// Release the IP address by using the handleID.
+	handleID, err := utils.GetHandleID(conf.Name, args.ContainerID, workloadID)
+	if err != nil {
 		return err
 	}
 
-	logger.Info("Released address using workloadID")
-	return nil
+	logger := log.WithFields(log.Fields{
+		"workloadID": workloadID,
+		"handleID":   handleID,
+	})
 
+	logger.Info("Releasing address using handleID")
+	if err := calicoClient.IPAM().ReleaseByHandle(handleID); err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			logger.WithError(err).Error("Failed to release address")
+			return err
+		}
+		logger.Warn("Asked to release address but it doesn't exist. Ignoring")
+	} else {
+		logger.Info("Released address using handleID")
+	}
+
+	// We also need to handle the old way by workloadID, to not leak IPs in existing installs.
+	logger.Info("Releasing address using workloadID")
+	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			logger.WithError(err).Error("Failed to release address")
+			return err
+		}
+		logger.Debug("Asked to release address but it doesn't exist. Ignoring")
+	} else {
+		logger.Info("Released address using workloadID")
+	}
+
+	return nil
 }
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -20,22 +20,22 @@ import (
 	"net"
 	"os"
 	"regexp"
+	"strings"
+
+	"github.com/containernetworking/cni/pkg/ns"
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/vishvananda/netlink"
 
 	log "github.com/sirupsen/logrus"
 
-	"strings"
-
 	"github.com/containernetworking/cni/pkg/ip"
 	"github.com/containernetworking/cni/pkg/ipam"
-	"github.com/containernetworking/cni/pkg/ns"
-	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
-	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/projectcalico/libcalico-go/lib/api"
 	"github.com/projectcalico/libcalico-go/lib/backend/model"
 	"github.com/projectcalico/libcalico-go/lib/client"
 	cnet "github.com/projectcalico/libcalico-go/lib/net"
-	"github.com/vishvananda/netlink"
 )
 
 func Min(a, b int) int {
@@ -169,6 +169,18 @@ func GetIdentifiers(args *skel.CmdArgs) (workloadID string, orchestratorID strin
 		orchestratorID = "cni"
 	}
 	return workloadID, orchestratorID, nil
+}
+
+func GetHandleID(netName string, containerID string, workload string) (string, error) {
+	handleID := fmt.Sprintf("%s.%s", netName, containerID)
+	log.WithFields(log.Fields{
+		"Network":     netName,
+		"ContainerID": containerID,
+		"Workload":    workload,
+		"HandleID":    handleID,
+	}).Debug("Generated IPAM handle")
+
+	return handleID, nil
 }
 
 func PopulateEndpointNets(endpoint *api.WorkloadEndpoint, result *current.Result) error {
--- a/calico_cni_ipam_test.go
+++ b/calico_cni_ipam_test.go
@@ -8,6 +8,8 @@ import (
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	. "github.com/projectcalico/cni-plugin/test_utils"
+	"github.com/projectcalico/libcalico-go/lib/client"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/testutils"
 )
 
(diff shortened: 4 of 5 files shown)


calico_cni_k8s_test.go
```diff
@@ -969,7 +969,7 @@ var _ = Describe("CalicoCni", func() {
 
 				checkIPAMReservation := func() {
 					// IPAM reservation should still be in place.
-					ipamIPs, err := calicoClient.IPAM().IPsByHandle(workloadName)
+					ipamIPs, err := calicoClient.IPAM().IPsByHandle(contNs.Path() + "." + containerID)
```
caseydavenport: `contNs.path()` isn't what we want to use here.

The key information should be some form of `(cni_network_name, containerID)`

utils/utils.go
```diff
@@ -171,6 +171,10 @@ func GetIdentifiers(args *skel.CmdArgs) (workloadID string, orchestratorID strin
 	return workloadID, orchestratorID, nil
 }
 
+func GetHandleId(args *skel.CmdArgs) (handleId string, err error) {
+	return fmt.Sprintf("%s.%s", args.Netns, args.ContainerID), nil
```
caseydavenport: again, the netns isn't the thing we want to use - it's the network name.

ipam/calico-ipam.go
```diff
@@ -97,11 +97,11 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
-	workloadID, _, err := utils.GetIdentifiers(args)
+	handleId, err := utils.GetHandleId(args)
 	if err != nil {
 		return err
 	}
-	logger := utils.CreateContextLogger(workloadID)
+	logger := utils.CreateContextLogger(handleId)
```
fasaxc: Please can we have a logger that has the workload ID *and* the handle ID in its context?  The container ID is just a random string so the really useful thing is the combination of the two.

ipam/calico-ipam.go
```diff
@@ -222,18 +222,34 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		return err
+	}
 
+	// We also need to handle the old way by workloadID, to not leak IPs in existing installs.
 	logger.Info("Releasing address using workloadID")
-	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
+	if err := calicoClient.IPAM().ReleaseByHandle(workloadId); err != nil {
 		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
-			logger.WithField("workloadId", workloadID).Warn("Asked to release address but it doesn't exist. Ignoring")
+			logger.WithField("workloadId", workloadId).Warn("Asked to release address but it doesn't exist. Ignoring")
```
fasaxc: If we succeeded in deleting the key above, then we should suppress this warning.  Going forward, we only expect the above handle ID delete to succeed.

ipam/calico-ipam.go
fasaxc: I'm also not a fan of putting _everything_ in a utils package, what's wrong with `logger := log.WithField("handle", handleID)`  It's just a short and it saves a trip to the other file to see what's going on.

utils/utils.go
```diff
@@ -171,6 +171,10 @@ func GetIdentifiers(args *skel.CmdArgs) (workloadID string, orchestratorID strin
 	return workloadID, orchestratorID, nil
 }
 
+func GetHandleId(args *skel.CmdArgs) (handleId string, err error) {
```
fasaxc: Nit: go naming convention is `GetHandleID`

ipam/calico-ipam.go
```diff
@@ -222,18 +222,34 @@ func cmdDel(args *skel.CmdArgs) error {
...
 	}
 
-	logger := utils.CreateContextLogger(workloadID)
+	// Release the IP address by using the workloadID, to not leak IPs in existing installs.
+	workloadId, _, err := utils.GetIdentifiers(args)
+	if err != nil {
+		return err
+	}
+
+	logger := utils.CreateContextLogger(handleId)
```
fasaxc: Let's have a logger with both workloadID and handle here.  It's super useful to be able to grep the kubelet log for the workload ID and see all the handles or vice versa.

heschlie:
In the tests I hardcoded the network name, as we're not returning it anywhere at the moment. I did notice that in `RunCNIPluginWithId` in the test_utils that we have a TODO to use `NetConf` instead of calling the CNI plugin, should we go ahead and switch now since we could use some info from it, and add the `NetConf` struct to the returned items (maybe trimming out some of them that are included in the struct as well)?

ipam/calico-ipam.go
```diff
@@ -222,24 +219,43 @@ func cmdDel(args *skel.CmdArgs) error {
 		return err
 	}
 
-	// Release the IP address by using the handle - which is workloadID.
+	// Release the IP address by using the handleID.
+	handleID := fmt.Sprintf("%s.%s", conf.Name, args.ContainerID)
```
caseydavenport: Are we worried about `conf.Name` containing invalid characters?

The CNI spec says this field is a "string" but doesn't give any more detail than that. Do we actually need to hash this or do some escaping to ensure we don't try to write a bad value?

heschlie: I think we do, if we don't really care how pretty it is hashing it seems like the better option to me, but if we want it to look sensible we should escape it, do we already have a function for this in libcalico somewhere?

heschlie: I went with b64 encoding the network name, once the tests pass locally I'll push up the code for the next review.

heschlie:
@caseydavenport @fasaxc Another push for your review.

ipam/calico-ipam.go
```diff
@@ -222,24 +222,46 @@ func cmdDel(args *skel.CmdArgs) error {
 		return err
 	}
 
-	// Release the IP address by using the handle - which is workloadID.
+	// Release the IP address by using the handleID.
```
caseydavenport: @heschlie I find the ordering of the current logic to be pretty confusing, and a bit hard to follow. Roughly, we've got:

- calculate the handle ID
- calculate the workloadID
- attempt release using the handle ID
  - unset a bool if IP not assigned
- attempt release using the workload ID, check boolean, emit error based on value.
  - only log if does not exist, and the previous bool was unset. 

I'm not a huge fan of using a boolean here for the sole purpose of hiding a log.

What about a flow like this?

- calculate handle ID
- attempt release using handle ID
  - if does not exist error:
    - calculate workloadID
    - attempt release using the workload ID
      - if does not exist error, log warning and return nil
  - if another type of error:
    - log error and return it

We can encapsulate the legacy workloadID logic into a helper function to make this code a bit cleaner.

heschlie: Originally I was thinking that there might be a situation where both the `handleID` and `workloadId` would exist at the same time, so we would need to call both (though admittedly could have been written better) but I'm less certain of that now, so I think encapsulating the `workloadId` call  is the way to go.

caseydavenport: As discussed offline, let's still release both, but I think we can clean up the logic a bit.

ipam/calico-ipam.go
```diff
@@ -222,24 +222,41 @@ func cmdDel(args *skel.CmdArgs) error {
...
-	logger.Info("Released address using workloadID")
-	return nil
+	// We also need to handle the old way by workloadID, to not leak IPs in existing installs.
+	workloadLogger.Info("Releasing address using workloadID")
+	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			return err
+		}
+	} else {
+		workloadLogger.Info("Released address using workloadId")
```
caseydavenport: nit: `workloadID`

ipam/calico-ipam.go
```diff
@@ -222,24 +222,41 @@ func cmdDel(args *skel.CmdArgs) error {
...
 	}
+	workloadLogger := utils.CreateContextLogger(workloadID)
 
-	logger.Info("Released address using workloadID")
-	return nil
+	// We also need to handle the old way by workloadID, to not leak IPs in existing installs.
+	workloadLogger.Info("Releasing address using workloadID")
+	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			return err
```
caseydavenport: Might be worth an error log in this branch.

ipam/calico-ipam.go
```diff
@@ -222,24 +222,41 @@ func cmdDel(args *skel.CmdArgs) error {
...
+		}
```
caseydavenport: And a debug log for the case where we get a `ResourceDoesNotExist` error.

Info is too high because this will be the mainline case, but debug should be fine and useful for debugging any potential issues with this code.

utils/utils.go
```diff
@@ -171,6 +175,16 @@ func GetIdentifiers(args *skel.CmdArgs) (workloadID string, orchestratorID strin
 	return workloadID, orchestratorID, nil
 }
 
+func GetHandleID(netName string, containerID string) (string, error) {
```
caseydavenport: It would be really useful to include a log in this function which correlates the netname, containerID, and the generated handle ID.

Do we get that anywhere else? Need to make sure that shows up both in ADD and DEL code paths.

heschlie: Added a context logger here, not sure if it is exactly what you had in mind but it should get the idea across

heschlie:
@caseydavenport @fasaxc ready for another look.

ipam/calico-ipam.go
```diff
@@ -222,24 +222,45 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	logger := utils.CreateContextLogger(handleID)
 
-	logger.Info("Releasing address using workloadID")
-	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
+	logger.Info("Releasing address using handleID")
+	if err := calicoClient.IPAM().ReleaseByHandle(handleID); err != nil {
 		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
-			logger.WithField("workloadId", workloadID).Warn("Asked to release address but it doesn't exist. Ignoring")
-			return nil
+			logger.Warn("Asked to release address but it doesn't exist. Ignoring")
```
caseydavenport: nit: you could/should reduce this down to a single branch by using `!ok` in the check above - then we don't need the "else" branch, because we'd just return in the error case.

ipam/calico-ipam.go
```diff
@@ -222,24 +222,45 @@ func cmdDel(args *skel.CmdArgs) error {
...
 		return err
 	}
+	workloadLogger := utils.CreateContextLogger(workloadID)
 
-	logger.Info("Released address using workloadID")
-	return nil
+	// We also need to handle the old way by workloadID, to not leak IPs in existing installs.
+	workloadLogger.Info("Releasing address using workloadID")
+	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
```
caseydavenport: Same here - if you put the error case first, then you can get rid of the "else" branch.

utils/utils.go
```diff
@@ -171,6 +175,23 @@ func GetIdentifiers(args *skel.CmdArgs) (workloadID string, orchestratorID strin
...
+		"ContainerID": containerID,
+	})
+	hasher := md5.New()
+	_, err := hasher.Write([]byte(netName))
+	if err != nil {
+		return "", err
+	}
+	hashedName := hex.EncodeToString(hasher.Sum(nil))
+	handleID := fmt.Sprintf("%s.%s", hashedName, containerID)
+	logger.Info("Generated handleID(%s) from hashed network name and containerID", handleID)
```
caseydavenport: Just a nit here - let's just use `WithFields` for all of them to make it consistent. 

Then, the log text should just be "Generated IPAM handle"

Also, it doesn't include the workload - we should pass in the main logger in order to get that context.

And finally, I think it should be a debug log.

ipam/calico-ipam.go
```diff
@@ -222,24 +227,42 @@ func cmdDel(args *skel.CmdArgs) error {
...
+
+	// We also need to handle the old way by workloadID, to not leak IPs in existing installs.
+	workloadLogger := utils.CreateContextLogger(workloadID)
+	workloadLogger.Info("Releasing address using workloadID")
 	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
-		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
-			logger.WithField("workloadId", workloadID).Warn("Asked to release address but it doesn't exist. Ignoring")
-			return nil
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			workloadLogger.Errorf("Failed to release address: %v", err)
```
fasaxc: Please use `workloadLogger.WithError(err).Error("Failed to release address.")`

utils/utils.go
```diff
@@ -171,6 +175,23 @@ func GetIdentifiers(args *skel.CmdArgs) (workloadID string, orchestratorID strin
 	return workloadID, orchestratorID, nil
 }
 
+func GetHandleID(netName string, containerID string, workload string) (string, error) {
+	hasher := md5.New()
```
fasaxc: Don't use MD5, it's broken.  If you need a crypto hash, reach for a SHA-2 family function.

fasaxc: Do we actually need to hash the network name?  It'd be really nice to avoid that if we can since it makes it a real pain to correlate later.

heschlie: @fasaxc this wasn't for security, it was just to ensure we ended up with only valid characters in the network name portion, as the CNI spec only dictates that it is a string. My thought was instead of trying to escape all invalid characters we just hash it as this value isn't presented to users (correct me if I'm wrong there).

If we really feel it would be helpful to store this value in a human readable format we can do that, we just need to escape the string.

heschlie: @fasaxc @caseydavenport thinking about this a bit more, are we worried about invalid values in `conf.Name`? Is the `HandleID` used anywhere but the IPAM code?

ipam/calico-ipam.go
```diff
@@ -222,24 +227,42 @@ func cmdDel(args *skel.CmdArgs) error {
...
+
+	// We also need to handle the old way by workloadID, to not leak IPs in existing installs.
+	workloadLogger := utils.CreateContextLogger(workloadID)
+	workloadLogger.Info("Releasing address using workloadID")
 	if err := calicoClient.IPAM().ReleaseByHandle(workloadID); err != nil {
-		if _, ok := err.(errors.ErrorResourceDoesNotExist); ok {
-			logger.WithField("workloadId", workloadID).Warn("Asked to release address but it doesn't exist. Ignoring")
-			return nil
+		if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			logger.WithError(err).Error("Failed to release address")
```
caseydavenport: Shouldn't this be `workloadLogger`?

utils/utils.go
```diff
@@ -171,6 +171,17 @@ func GetIdentifiers(args *skel.CmdArgs) (workloadID string, orchestratorID strin
 	return workloadID, orchestratorID, nil
 }
 
+func GetHandleID(netName string, containerID string, workload string) (string, error) {
+	handleID := fmt.Sprintf("%s.%s", netName, containerID)
+	log.WithFields(log.Fields{
+		"Network":     netName,
+		"ContainerID": containerID,
+		"Workload":    workload,
```
caseydavenport: Log is missing the handle itself :)

ipam/calico-ipam.go
heschlie: argh, copy paste problems

utils/utils.go
heschlie: Should that be part of the with fields or in the log message itself?

caseydavenport: fields

PR #447, opened 2017-12-14
rev libcalico-go

bcreane:
## Description
rev libcalico-go to pick up empty host endpoint fix in libcalico-go

also moved to latest felix hash, and also pinned client-go to the same version that felix pins it to.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -22,7 +22,7 @@ import:
   subpackages:
   - gexec
 - package: github.com/projectcalico/libcalico-go
-  version: c07bf34a9a362e92e5e51853fd5fe03c27ecf333
+  version: f58e21ead99b6da5b01074873e099b97ef017ca7
   subpackages:
   - lib/apiconfig
   - lib/apis/v2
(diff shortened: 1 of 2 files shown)


glide.yaml
```diff
@@ -32,12 +32,13 @@ import:
...
 - package: github.com/vishvananda/netlink
 - package: k8s.io/client-go
+  version: 82aa063804cf055e16e8911250f888bc216e8b61
   subpackages:
   - kubernetes
   - tools/clientcmd
 - package: github.com/mcuadros/go-version
 - package: github.com/projectcalico/felix
-  version: 518dbc352494a962e46d50f8bfe72a70c86951bf
+  version: 357cdd9e7fc181f56c6c197201aa39a6b0ba83c2
```
gunjan5: curious, any reason we had to update this?

bcreane: I pulled the client-go version from felix, so this change was more for completeness rather than any particular feature/fix that I wanted to pull in.

PR #448, opened 2017-12-15
refine logic in case ls take too much time

pikeszfish:
Signed-off-by: Pike <pikeszfish@gmail.com>

## Description
I think previous code logic will fail when `if [ "$(ls -A ${SECRETS_MOUNT_DIR} 2>/dev/null)" ];` take too much time and `stat_output ` will always be equal to the command `$(stat -c%y ${SECRETS_MOUNT_DIR}/etcd-cert 2>/dev/null)` executed 10 second later.
So I just refine them.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note
I will add anything is needed.

## Release Note
```
refine logic code in case `ls` take too much time
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -166,7 +166,7 @@ while [ "$should_sleep" == "true"  ]; do
 	# version to the host. Just check the timestamp on the certificate to see if it
 	# has been updated.  A bit hokey, but likely good enough.
 	current_stat_output=$(stat -c%y ${SECRETS_MOUNT_DIR}/etcd-cert 2>/dev/null)
-	if [ "$old_stat_output" != "$current_output" ]; then
+	if [ "$old_stat_output" != "$current_stat_output" ]; then
 		old_stat_output=$current_stat_output
 		echo "Updating installed secrets at: $(date)"
 		cp ${SECRETS_MOUNT_DIR}/* /host/etc/cni/net.d/calico-tls/
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -168,8 +168,8 @@ while [ "$should_sleep" == "true"  ]; do
 	current_stat_output=$(stat -c%y ${SECRETS_MOUNT_DIR}/etcd-cert 2>/dev/null)
 	if [ "$old_stat_output" != "$current_output" ]; then
 		old_stat_output=$current_stat_output
-        echo "Updating installed secrets at: $(date)"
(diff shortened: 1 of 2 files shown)


caseydavenport:
@pikeszfish thanks for the PR!

I hope to review this in the new year. I've targeted v3.1.0 for now, but we might be able to get it in v3.0.x.

k8s-install/scripts/install-cni.sh
```diff
@@ -160,19 +160,16 @@ echo "Created CNI config ${CNI_CONF_NAME}"
...
-        if [ "$stat_output" != "$(stat -c%y ${SECRETS_MOUNT_DIR}/etcd-cert 2>/dev/null)" ]; then
-            echo "Updating installed secrets at: $(date)"
-            cp ${SECRETS_MOUNT_DIR}/* /host/etc/cni/net.d/calico-tls/
-        fi
-    else
-        sleep 10
+	current_stat_output=$(stat -c%y ${SECRETS_MOUNT_DIR}/etcd-cert 2>/dev/null)
+	if [ "$old_stat_output" != "$current_output" ]; then
+		old_stat_output=$current_stat_output
+        echo "Updating installed secrets at: $(date)"
```
caseydavenport: Could you fix up the indentation here?

caseydavenport:
@pikeszfish looks like CI is failing: https://semaphoreci.com/calico/cni-plugin/branches/pull-request-448/builds/1

Is this an optimization, or did you actually encounter this on a live system?

k8s-install/scripts/install-cni.sh
pikeszfish: updated. Using tab now.

k8s-install/scripts/install_cni_test.go
```diff
@@ -42,6 +42,7 @@ func runCniContainer(extraArgs ...string) error {
 		"-v", cwd + "/tmp/bin:/host/opt/cni/bin",
 		"-v", cwd + "/tmp/net.d:/host/etc/cni/net.d",
 		"-v", cwd + "/tmp/serviceaccount:/var/run/secrets/kubernetes.io/serviceaccount",
+		"-v", cwd + "/tmp/etcd-cert:/calico-secrets/etcd-cert",
```
caseydavenport: Looks like we now require the `calico-secrets/etcd-cert` file to exist? I think we need to function even when that file is not present.

caseydavenport:
I'm going to close this for now due to inactivity. We can re-open but I think we need to understand what this is trying to solve a bit better first. Please raise an issue to discuss if you're interested :)

Issue #428, opened 2017-11-16
Pended test leaks a veth and route


fasaxc:
`Create a container then send another ADD for the same container but with a different interface` test leaks a veth and route, which causes flakes if other tests try to use the same IP.

I've pended the test while I diagnose and fix.

## Expected Behavior

No leak.
## Current Behavior

veth and route are left over.  It's not clear if that's a test bug or a code bug.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* CNI plugin version (e.g. `/opt/cni/bin/calico --version`):
* Orchestrator version (e.g. kubernetes, mesos, rkt):
* CNI config:
* Operating System and version:
* Link to your project (optional):

bcreane:
@fasaxc , @caseydavenport - is this a v3.0.0 blocker or can we retarget to v3.0.1?

fasaxc:
I think it can be v3.0.1

PR #452, opened 2018-01-02
Un-pend multi-container test.

fasaxc:
## Description

Fix and un-pend formerly flaky test.

The test was assuming that the name of the veth inside the container was always "eth0" but the CNI plugin now honours the requested interface name for the container veth, _even though it always stores the WEP with name "eth0"_.

Tearing it down with "eth0" was silently ignored (the CNI plugin couldn't find the interface so it assumed the tear down had already been done).

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -1406,22 +1406,22 @@ var _ = Describe("CalicoCni", func() {
 				})
 			})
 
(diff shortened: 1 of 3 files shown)


fasaxc:
Fixes #428

PR #453, opened 2018-01-03
Clean up superfluous Profile metadata labels

neiljerram:
The meaning that used to be attached to these labels is now with
Spec.LabelsToApply.

This is a cleanup that was requested during the review of
https://github.com/projectcalico/cni-plugin/pull/399.


neiljerram:
Thanks @gunjan5 !

PR #457, opened 2018-01-18
[WIP] Adds support to use Namespace annotation for ippools

rikatz:
## Description

Type of fix: new feature

This PR adds support to global Namespace IP Pool allocations.

The main idea here is to allow the cluster administrator to define and/or force the usage of a specific ippool in some Namespace

The following workflow happens:

* If the POD is the only object annotated with ipv4/ipv6 pools, this annotation is used
* If the Namespace is the only object annotated with ipv4/ipv6 pools, namespace annotation is used
* If both the POD and Namespace are annotated with ipv4/ipv6 pools, POD annotation is used
* If both are annotated, but the additional annotation ``cni.projectcalico.org/forceNSPools`` is found as 'true', then the ipv4/ipv6 pools annotation from namespace is used
* If forceNSpool is found as 'true' but no ipv4/ipv6 annotation are found in the Namespace, no annotation is used and the default pools configured in CNI are used (probably this should be improved with a log message)

This is still a WIP but needs review and also a lot of tests.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -89,6 +89,15 @@ var _ = Describe("CalicoCni", func() {
 					panic(err)
 				}
 
+				// Create the Namespace before the tests
+				_, err = clientset.CoreV1().Namespaces().Create(&v1.Namespace{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:        "test",
+						Annotations: map[string]string{},
+					},
+				})
+				Expect(err).NotTo(HaveOccurred())
+
 				name := fmt.Sprintf("run%d", rand.Uint32())
 
 				// Create a K8s pod w/o any special params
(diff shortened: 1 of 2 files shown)


rikatz:
Hum, Don't know why the CI failed, but as this is a WIP/PoC, if someone can guide me through the right path I can finish this.

Also I'll make some tests (haven't tested yet, just wrote the idea), improve the test files and so.

Thanks!!

rikatz:
Just figured out a crucial error here, always setting pod annotation when forcens is not true. Will improve this tomorrow

Ricardo P. Katz

Em 17 de jan de 2018, à(s) 23:34, CLAassistant <notifications@github.com> escreveu:

caseydavenport:
@rikatz thanks for the PR - I think this is a really cool idea. 

The CI linter seems to be failing due to a few unchecked error returns:

```
```

rikatz:
Tks @caseydavenport 

I'm facing some other errors right now, but will debug it later :)

rikatz:
@caseydavenport It's working (at least, tests say so).

I'm going to test also in my lab environment, overwriting calico and calico-ipam programs.

Can you please also take a look to see if I didn't messed up something? :)

Thanks!!

rikatz:
Okay, everything worked here.

There's also an improvement needed in calico-cni-plugin ClusterRole (as now we're reading also the namespace resource), made in https://github.com/projectcalico/calico/pull/1629

As this PR is approved, I'll also change the CNI Docs with the new annotations.

caseydavenport:
@rikatz sorry for the delay, I've got a backlog of reviews to get through before I get to this one but I haven't forgotten it!

rikatz:
Sure! Just to know if this PR is still alive :) Take your time!! :)

caseydavenport:
@rikatz sorry for the long delay on this! I'd like to merge this for Calico v3.2.

My main feedback is this - I'd like to remove the `cni.projectcalico.org/forceNSPools` option for now (we can add it or something similar later if needed).

Instead, what about this behavior:

- If the annotation is present on the Namespace, use it. Otherwise, use what's on the Pod.
- If the annotation is present on both the Namespace _and_ the Pod, return an appropriate error to the user.

I think this is a simpler and more intuitive behavior. If we want to support both in the future then we can extend that behavior. WDYT? Would that meet your requirements?

rikatz:
@caseydavenport I'll make the changes as suggested :) Sorry about the delay to answer also, I've been pretty busy here. I'll take a look at this in this week

caseydavenport:
CC @matthewdupre who I believe is interested in this

rikatz:
Hi @caseydavenport I'm really sorry about taking so long to take a look into this. I've returned from my vacations today, and a lot of unread emails to take care.

Will try to re-write this ASAP, but if someone else can also take care of this it would be great!!

Tks

caseydavenport:
> I'm really sorry about taking so long to take a look into this. I've returned from my vacations today, and a lot of unread emails to take care.

No worries and no rush, I hope you had a great vacation!

I believe @matthewdupre is potentially looking to do this "soon", but not sure. So whoever gets to it first can finish it off :)

rikatz:
@caseydavenport I'm reviewing here.

I was thinking about not returning an error to the user (if both annotations are set) as this could lead to some strange behaviours.

What if we respect the most granular information, explaining:

1) Namespace annotation exists? Use annotation info
2) POD annotatation exists? Replace the old annotation (set by namespace) to POD annontation
3) No annotation exists? Use the default Pools.

This could make easier to develop, in a future release the 'force' option.

Anyway I'm finishing here as suggested, waiting for your feedback :)

Tks

rikatz:
OK, finished here rebasing, squashing, etc etc.

@caseydavenport I did as I firstly suggested instead of giving user an error:

* If there's only a NS annotation, use this NS annotation
* If there's a NS annotation and a POD annotation, use POD annotation
* If there's only a POD annotation, use the POD annotation (original behaviour)
* Otherwise uses the default IP Pool (original Behaviour)

I think this way, if a cluster admin puts a NS annotation in a project that still haves a lot of deployments with Pod annotations it will not generate any impact. In a near future we can improve the 'Force NS annotation', but right now these changes have no impact in working environments :)

Thanks!!

caseydavenport:
@rikatz thanks much!

I think I'm OK with that behavior. I'd like @matthewdupre to review as well since he's interested in this feature.

PR #467, opened 2018-02-07
Sanitize mesos labels

ozdanborne:
Slashes (`/`) are a valid character for Mesos labels.

This PR sanitizes slashes in Mesos labels by converting them to periods (`.`).

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -18,10 +18,9 @@ import (
 	goerrors "errors"
 	"flag"
 	"fmt"
+	"net"
 	"os"
 	"runtime"
-
-	"net"
 
 	"github.com/containernetworking/cni/pkg/ipam"
 	"github.com/containernetworking/cni/pkg/skel"
@@ -191,7 +190,12 @@ func cmdAdd(args *skel.CmdArgs) error {
 			// Parse endpoint labels passed in by Mesos, and store in a map.
 			labels := map[string]string{}
 			for _, label := range conf.Args.Mesos.NetworkInfo.Labels.Labels {
-				labels[label.Key] = label.Value
+				// Sanitize mesos labels so that they pass the k8s label validation,
+				// as mesos labels accept any unicode value.
+				k := SanitizeMesosLabel(label.Key)
+				v := SanitizeMesosLabel(label.Value)
+
+				labels[k] = v
 			}
 
 			// 2) Create the endpoint object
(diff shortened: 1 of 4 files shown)


calico.go
```diff
@@ -191,7 +192,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 			// Parse endpoint labels passed in by Mesos, and store in a map.
 			labels := map[string]string{}
 			for _, label := range conf.Args.Mesos.NetworkInfo.Labels.Labels {
-				labels[label.Key] = label.Value
+				k := strings.Replace(label.Key, "/", ".", -1)
```
caseydavenport: Probably worth a quick comment along the lines of "Slashes are allowed in Mesos labels but not in Calico ones"

calico_cni_test.go
```diff
@@ -281,6 +281,132 @@ var _ = Describe("CalicoCni", func() {
...
+						"k": "v",
+					},
+				}))
+
+				_, err = DeleteContainer(netconf, contNs.Path(), "")
+				Expect(err).ShouldNot(HaveOccurred())
+			})
+		})
+
+		Context("Mesos Labels", func() {
```
caseydavenport: Shouldn't these all be in the same `Context`? You've got multiple Contexts with the same description as it is.

calico.go
```diff
@@ -312,6 +318,33 @@ func cmdAdd(args *skel.CmdArgs) error {
 	return types.PrintResult(result, cniVersion)
 }
 
+// sanitizeMesosLabel converts a string from a valid mesos label to a valid Calico label.
+// Mesos labels have no restriction outside of being unicode.
+func sanitizeMesosLabel(s string) string {
+	// Inspired by:
+	// https://github.com/projectcalico/libcalico-go/blob/2ff29bed865c4b364d4fcf1ad214b2bd8d9b4afa/lib/upgrade/converters/names.go#L39-L58
+	var (
```
caseydavenport: A bit weird to have a `var` section inside a function - could you just use `:=` here instead?

You might want to move this into one of the utility files as well.

utils/utils.go
```diff
@@ -154,6 +154,32 @@ func CreateResultFromEndpoint(ep *api.WorkloadEndpoint) (*current.Result, error)
 	return result, nil
 }
 
+// SanitizeMesosLabel converts a string from a valid mesos label to a valid Calico label.
+// Mesos labels have no restriction outside of being unicode.
+func SanitizeMesosLabel(s string) string {
```
caseydavenport: I think we need some more unit-level testing of this function, e.g. for labels like the following:

- `-some-label`
- `$mylabel..`
- `//something/here/`
- `some_label-with.lots*of^weird#characters`

PR #472, opened 2018-02-14
Check for non-hidden cert files

tmjd:
## Description
With Kubernetes v1.8.8 and v1.9.3 this change https://github.com/kubernetes/kubernetes/pull/57422 was introduced which creates hidden folders in any mounted in Secret.  This was problematic because now the check for files returned true but then attempting to copy (non-hidden) files failed.  This change has the check only return true if there are non-hidden files.

## Todos
- [ ] Test

## Release Note

```release-note
Ignore hidden files when checking for etcd certificates to copy over when installing CNI.
```


tmjd:
How about that? :smiley:

PR #480, opened 2018-03-02
Use the /var/lib/calico/nodename file if it exists

caseydavenport:
## Description

In combination with https://github.com/projectcalico/calico/pull/1722, will make it much harder to accidentally end up with mismatched node names between the two components.

Fixes https://github.com/projectcalico/calico/issues/1093

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
The Calico CNI plugin by default expects the /var/lib/calico/nodename file to be created by calico/node. To disable this feature, set `nodename_file_optional: true` in your CNI network configuration.
```

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -38,8 +38,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var nodename string
-
 func init() {
 	// This ensures that main runs only on main thread (thread group leader).
 	// since namespace ops (unshare, setns) are done for a single thread, we
@@ -56,8 +54,17 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 	utils.ConfigureLogging(conf.LogLevel)
 
-	// Allow the nodename to be overridden by the network config
-	nodename = utils.DetermineNodename(conf)
+	if !conf.NodenameFileOptional {
+		// Configured to wait for the nodename file - don't start until it exists.
+		if _, err := os.Stat("/var/lib/calico/nodename"); err != nil {
+			s := "%s: check that the calico/node container is running and has mounted /var/lib/calico/"
+			return fmt.Errorf(s, err)
+		}
+		logrus.Debug("/var/lib/calico/nodename exists")
+	}
+
+	// Determine which node name to use.
+	nodename := utils.DetermineNodename(conf)
 
 	// Extract WEP identifiers such as pod name, pod namespace (for k8s), containerID, IfName.
 	wepIDs, err := utils.GetIdentifiers(args, nodename)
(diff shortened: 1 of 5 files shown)


calico.go
```diff
@@ -56,8 +54,16 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 
+	if conf.RequireNodenameFile {
+		// Configured to use the nodename file - don't start until it exists.
+		if _, err := os.Stat("/var/lib/calico/nodename"); err != nil {
+			return fmt.Errorf("error reading /var/lib/calico/nodename: %s: is calico/node running?", err)
+		}
+		logrus.Debug("/var/lib/calico/nodename exists")
+	}
+
 	// Allow the nodename to be overridden by the network config
```
ozdanborne: Now that nodename is being instantiated here, I don't think this comment is right

utils/utils.go
```diff
@@ -57,12 +58,34 @@ func DetermineNodename(conf types.NetConf) string {
...
+		logrus.Debugf("Read node name from CNI conf: %s", conf.Nodename)
 		nodename = conf.Nodename
 	}
+	logrus.Debugf("Using node name %s", nodename)
 	return nodename
 }
 
+// nodenameFromFile reads the /var/lib/calico/nodename file if it exists and
+// returns the nodename within.
+func nodenameFromFile() string {
```
ozdanborne: A nodeNameFile that exists but is empty takes an interesting path through this code, but I suppose it is ok...

utils/utils.go
```diff
@@ -57,12 +58,34 @@ func DetermineNodename(conf types.NetConf) string {
 		nodename = conf.Hostname
 		logrus.Warn("Configuration option 'hostname' is deprecated, use 'nodename' instead.")
 	}
+	if nff := nodenameFromFile(); nff != "" {
+		logrus.Debugf("Read node name from file: %s", nff)
+		nodename = nff
+	}
 	if conf.Nodename != "" {
```
ozdanborne: Should this be an `else if`? If multiple methods are provided (albeit an unusual edge case), the lowest priority would take precedence.

calico.go
caseydavenport: I don't think the comment was ever right.

utils/utils.go
caseydavenport: The way the code is structured now is acting as designed - the `conf.Nodename` takes priority over the `/var/lib/calico/nodename` file.

utils/utils.go
caseydavenport: Hm, yeah. I suspect we probably want to error in that case.

caseydavenport: Actually, let's leave this as it is for now. I think the code will do the right thing and I'm not sure how we'd end up with an empty file anyway.

Issue #459, opened 2018-01-23
TLS secrets readable by all users after copy to host


hafe:
Provided etcd TLS secrets are readable by all users on the host after the CNI plugin has executed. This has security implications, depending if the etcd cluster is shared and/or RBAC is enabled the credentials could means indirect K8S cluster admin access.

## Expected Behavior
TLS secrets should have "-r--r-----" or "-r-----" mode.

## Current Behavior
File after copy has "-rw-r--r--" mode.

## Possible Solution
Instead of using cp in the install script, install can be used where mode can be specified as an option.

## Steps to Reproduce (for bugs)

1. deploy kubernetes and calico as CNI
2.
3.
4.

## Context
Unintentional sec audit after k8s+calico installation

## Your Environment

* CNI plugin version (e.g. `/opt/cni/bin/calico --version`): 2.6.5 (problem exist in master)
* Orchestrator version (e.g. kubernetes, mesos, rkt): kubernetes
* CNI config:
* Operating System and version:
* Link to your project (optional):

caseydavenport:
@hafe yeah. would be great to do this, and should be an easy fix somewhere around here: https://github.com/projectcalico/cni-plugin/blob/master/k8s-install/scripts/install-cni.sh#L26-L33

Anyone interested in submitting a PR?

caseydavenport:
Looks like we should be setting permissions, since k8s defaults to 0644: https://kubernetes.io/docs/concepts/configuration/secret/

Then, we just need to use `cp -p` to make sure permissions are transferred as well. I've done that as part of this PR: https://github.com/projectcalico/cni-plugin/pull/481

PR #481, opened 2018-03-02
Improve secret copy + kubeconfig permissions 

caseydavenport:
## Description

A few improvements to our install CNI script:

- Copy secrets with `-p` to retain permissions (See https://github.com/projectcalico/cni-plugin/issues/459)
- Set `calico-kubeconfig` mode to `0600` by default
- Move token + API location into calico-kubeconfig file.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
The install-cni container now maintains the original mode on certificates copied from Kubernetes secrets.
```

```release-note
The install-cni container now writes the calico-kubeconfig file with mode 600 by default. It can be configured by setting the KUBECONFIG_MODE option.
```

```release-note
The install-cni container now only writes the calico-kubeconfig file when running as a Kubernetes pod.
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -29,7 +29,7 @@ if [ "$(ls ${SECRETS_MOUNT_DIR} 3>/dev/null)" ];
 then
 	echo "Installing any TLS assets from ${SECRETS_MOUNT_DIR}"
 	mkdir -p /host/etc/cni/net.d/calico-tls
-	cp ${SECRETS_MOUNT_DIR}/* /host/etc/cni/net.d/calico-tls/
+	cp -p ${SECRETS_MOUNT_DIR}/* /host/etc/cni/net.d/calico-tls/
 fi
 
 # If the TLS assets actually exist, update the variables to populate into the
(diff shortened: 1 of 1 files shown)


k8s-install/scripts/install-cni.sh
```diff
@@ -119,7 +127,6 @@ current-context: calico-context
 EOF
 
 # Insert any of the supported "auto" parameters.
-SERVICEACCOUNT_TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)
```
tmjd: The replacement below of SERVICEACCOUNT_TOKEN could be removed once it is removed from the CNI config, are we waiting until that is done to allow an easier transition. I think that makes sense, just making sure it has been considered.

caseydavenport: Yeah, I think we should wait a while. We should probably leave it there for compat purposes indefinitely. We can document it as deprecated if we want though.

k8s-install/scripts/install-cni.sh
```diff
@@ -96,20 +96,28 @@ ${CNI_NETWORK_CONFIG:-}
...
+touch /host/etc/cni/net.d/calico-kubeconfig
+chmod ${KUBECONFIG_MODE:-600} /host/etc/cni/net.d/calico-kubeconfig
 cat > /host/etc/cni/net.d/calico-kubeconfig <<EOF
 # Kubeconfig file for Calico CNI plugin.
 apiVersion: v1
 kind: Config
 clusters:
 - name: local
   cluster:
+    server: https://${KUBERNETES_SERVICE_HOST:-}:${KUBERNETES_SERVICE_PORT:-}
```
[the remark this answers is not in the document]
caseydavenport: I don't know if we can put a reasonable default here. It's all a bit awkward, but those envs really should be set whenever we're running as a k8s pod (but not say, mesos).

+1 on adding http/https configuration.

caseydavenport:
@tmjd thanks for the review, I've pushed some changes, LMK what you think.

PR #484, opened 2018-03-08
Dont allocate from disabled poools

ozdanborne:
## Description

Bump version of libcalico-go to include patch where IP's aren't allocated from disabled pools.

Additionally, add a test (thanks @tmjd ) which tests IPAM to ensure this change works.

See https://github.com/projectcalico/libcalico-go/issues/689

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Don't allocate addresses from disabled pools
```

### DIFF
--- a/calico_cni_ipam_test.go
+++ b/calico_cni_ipam_test.go
@@ -1,10 +1,9 @@
 package main_test
 
 import (
+	"context"
 	"fmt"
 	"os"
-
-	"context"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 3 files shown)


ozdanborne:
I'm rebased and not seeing any issues....

Do you mean a newer version of libcalico-go? Or perhaps since this is testing calico-ipam it's unaffected?

caseydavenport:
Ohh, yes. You're right, it's only running the IPAM code which doesn't have that check in it.

PR #491, opened 2018-03-18
Unifying MTU settings in ConfigMap

ketkulka:
- Have a single point of MTU definition
- CNI_MTU variable will set mtu for cni as well as
  FELIX_IPINIPMTU
- For more control, both var can still be set individually

## Description
Fixes Issue #1482 from calico


k8s-install/scripts/install-cni.sh
```diff
@@ -144,11 +144,13 @@ EOF
 fi
 
 
+CNI_MTU=${CNI_MTU:-1500}
 # Insert any of the supported "auto" parameters.
 grep "__KUBERNETES_SERVICE_HOST__" $TMP_CONF && sed -i s/__KUBERNETES_SERVICE_HOST__/${KUBERNETES_SERVICE_HOST}/g $TMP_CONF
 grep "__KUBERNETES_SERVICE_PORT__" $TMP_CONF && sed -i s/__KUBERNETES_SERVICE_PORT__/${KUBERNETES_SERVICE_PORT}/g $TMP_CONF
 sed -i s/__KUBERNETES_NODE_NAME__/${KUBERNETES_NODE_NAME:-$(hostname)}/g $TMP_CONF
 sed -i s/__KUBECONFIG_FILENAME__/calico-kubeconfig/g $TMP_CONF
+grep "__CNI_MTU__" $TMP_CONF && sed -i s/__CNI_MTU__/${CNI_MTU}/g $TMP_CONF
```
caseydavenport: You could do this in one line (without the defaulting above):

```
grep "__CNI_MTU__" $TMP_CONF && sed -i s/__CNI_MTU__/${CNI_MTU:-1500}/g $TMP_CONF
```

You can also drop the `grep` part from the front - that was required for the `KUBERNETES_` variables because on non-k8s systems they don't exist.

So just:
```
sed -i s/__CNI_MTU__/${CNI_MTU:-1500}/g $TMP_CONF
```

ketkulka: Oh I thought someone can just have "mtu" : 1100 directly in cni_config in which case __CNI_MTU__ won't be there. So included grep part to take care of that case. Won't that happen?

caseydavenport:
@ketkulka it's fine, but the `sed` command handles that for you - if it's not in the cni config then it simply won't be replaced - note that most of the other sed commands don't have the grep in front, because it's not needed.

It's only required for the KUBERNETES_SERVICE_* variables because there is no reasonable default value.

ketkulka:
thanks @caseydavenport i have pushed the change removing grep

ketkulka:
@caseydavenport appreciate your patience with this PR. I did squash using `git reset --soft HEAD~4; git commit; git push -f` Result is what I expected but not sure if that was the right set of commands i used :)

PR #501, opened 2018-04-09
Allow sysctl configuration

caseydavenport:
## Description

https://github.com/projectcalico/cni-plugin/pull/381 broke some users who were relying on this behavior, and we probably shouldn't have changed the default.

This PR makes IP forwarding configurable in the CNI network config.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Add support for configuring container IP forwarding via the CNI configuration file.
```

### DIFF
--- a/calico_cni_test.go
+++ b/calico_cni_test.go
@@ -127,6 +127,13 @@ var _ = Describe("CalicoCni", func() {
 				err = testutils.CheckSysctlValue(fmt.Sprintf("/proc/sys/net/ipv4/conf/%s/forwarding", hostVethName), "1")
 				Expect(err).ShouldNot(HaveOccurred())
 
+				// Assert the container sysctl values are set to what we expect for IPv4.
+				targetNs, _ := ns.GetNS(contNs.Path())
+				err = targetNs.Do(func(_ ns.NetNS) error {
+					return testutils.CheckSysctlValue("/proc/sys/net/ipv4/ip_forward", "0")
+				})
+				Expect(err).ShouldNot(HaveOccurred())
+
 				// Assert if the host side route is programmed correctly.
 				hostRoutes, err := netlink.RouteList(hostVeth, syscall.AF_INET)
 				Expect(err).ShouldNot(HaveOccurred())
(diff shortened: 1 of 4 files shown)


ozdanborne:
It seems like a bit of a security risk to allow arbitrary setting of sysctl through pod annotations. I understand that RBAC should prevent this, but maybe we should be more rigorous by whitelisting which sysctl's we allow for now.

types/types.go
```diff
@@ -69,21 +69,30 @@ type NetConf struct {
...
 	EtcdEndpoints        string     `json:"etcd_endpoints"`
 	LogLevel             string     `json:"log_level"`
 	Policy               Policy     `json:"policy"`
 	Kubernetes           Kubernetes `json:"kubernetes"`
-	Args                 Args       `json:"args"`
 	EtcdScheme           string     `json:"etcd_scheme"`
 	EtcdKeyFile          string     `json:"etcd_key_file"`
 	EtcdCertFile         string     `json:"etcd_cert_file"`
 	EtcdCaCertFile       string     `json:"etcd_ca_cert_file"`
+	Sysctls              Sysctls    `json:"sysctls,omitempty"`
```
fasaxc: `Sysctls` is a bit linux-specific.  How about `ContainerPermissions|Settings|Tweaks`   or similar?

caseydavenport:
Noting here - I'd like to cherry-pick this into v2.6, v3.0, v3.1.

@fasaxc I think this is ready for another review!

calico_cni_test.go
```diff
@@ -254,6 +260,49 @@ var _ = Describe("CalicoCni", func() {
...
+			By("asserting IPv4 forwarding is enabled", func() {
+				targetNs, _ := ns.GetNS(contNs.Path())
+				err = targetNs.Do(func(_ ns.NetNS) error {
+					return testutils.CheckSysctlValue("/proc/sys/net/ipv4/ip_forward", "1")
+				})
+				Expect(err).ShouldNot(HaveOccurred())
+			})
+
+			By("tearing down the container", func() {
+				_, err = testutils.DeleteContainerWithId(netconf, contNs.Path(), "", testutils.TEST_DEFAULT_NS, containerID)
```
ozdanborne: should this be in a `defer`?

caseydavenport: Probably not a defer, but perhaps it belongs in an `AfterEach` clause.

PR #505, opened 2018-04-11
Disable ipAddrsNoIpam by default

ozdanborne:
## Description

Add a flag to enable/disable ipAddrsNoIpam. Disable it by default.

The flag is added to the toplevel section of the cni config, in a boolean field called `enable_ip_addrs_no_ipam`.  The CNI spec explicitly allows special fields at this level. I also considered adding a new object field called `calico_features` but wasn't sure there was a strong enough use case for that yet.

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Move ipAddrsNoIpam behind a feature flag in the CNI configuration, and disable it by default. Clusters currently using this feature will need to update their CNI configuration to continue using it when upgrading to this release.
```

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -191,7 +191,13 @@ func cmdAdd(args *skel.CmdArgs) error {
 			return err
 		}
 	} else {
-		// Default CNI behavior - use the CNI network name as the Calico profile.
+		// Default CNI behavior
+		// Validate enabled features
+		if conf.FeatureControl.IPAddrsNoIpam {
+			return errors.New("requested feature is not supported for this runtime: ip_addrs_no_ipam")
+		}
+
+		// use the CNI network name as the Calico profile.
 		profileID := conf.Name
 
 		endpointAlreadyExisted := endpoint != nil
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/containernetworking/plugins/pkg/ns"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
 	"github.com/projectcalico/cni-plugin/testutils"
 	"github.com/projectcalico/cni-plugin/utils"
@@ -460,6 +461,9 @@ var _ = Describe("CalicoCni", func() {
 			      "cniVersion": "%s",
 				  "name": "net3",
 				  "type": "calico",
+				  "feature_control": {
+					"ip_addrs_no_ipam": true
+				  },
 				  "etcd_endpoints": "http://%s:2379",
 				  "datastore_type": "%s",
 			          "nodename_file_optional": true,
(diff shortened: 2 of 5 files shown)


calico_cni_k8s_test.go
```diff
@@ -465,6 +466,7 @@ var _ = Describe("CalicoCni", func() {
 			          "nodename_file_optional": true,
 			 	  "ipam": {},
 					"kubernetes": {
+					  "enable_ip_addrs_no_ipam": true,
```
caseydavenport: I think my main bit of feedback is whether this belongs in the "kubernetes" section.

Imagining a world where we support this feature in other integrations, or where we want to enable / disable other features that aren't k8s specific.

So, we might want something more generic. WDYT about something like this?

```json
{
    "kubernetes": {},
    "feature_control": {
        "ip_addrs_no_ipam": false,
        "ip_addrs": true,
        "namespace_ipam": true,
    }
}
```

k8s/k8s.go
```diff
@@ -208,6 +208,12 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 			logger.Error(e)
 			return nil, e
 		case ipAddrsNoIpam != "":
+			if !conf.FeatureControl.IPAddrsNoIpam {
+				e := fmt.Errorf("can't honor 'ipAddrsNoIpam' request: ipAddrsNoIpam is not enabled")
```
caseydavenport: I'd recommend this for an error message:

`"requested feature is not enabled: cni.projectcalico.org/ipAddrsNoIpam"`

ozdanborne: What about `requested feature is not enabled: ip_addrs_no_ipam`, since that's what the name of it is in the CNI config?

caseydavenport: Yeah, I did consider that as well.

I went with the annotation since this is in the Kubernetes code and that's the way that Kubernetes users will interact with that feature. I'm OK with using `ip_addrs_no_ipam` but I do prefer the full annotation.

Makefile
```diff
@@ -235,7 +235,7 @@ build-containerized: vendor
 ## Run the tests locally, must have local etcd running
 run-tests:
 	# Run tests recursively (-r).
-	ginkgo -cover -r -skipPackage vendor -skipPackage k8s-install
+	ginkgo -cover -r -skipPackage vendor -skipPackage k8s-install -focus "fail if ipAddrsNoIpam is not enabled"
```
caseydavenport: Remember to take this out (... or, add support for a FOCUS argument!)

calico.go
```diff
@@ -191,7 +191,13 @@ func cmdAdd(args *skel.CmdArgs) error {
 			return err
 		}
 	} else {
-		// Default CNI behavior - use the CNI network name as the Calico profile.
+		// Default CNI behavior
+		// Validate enabled features
+		if conf.FeatureControl.IPAddrsNoIpam {
+			return errors.New("requested feature is not supported for this orchestrator: ip_addrs_no_ipam")
```
caseydavenport: I think we should swap out the word "orchestrator" for "runtime" or "environment" or something. It's possible to have a Kubernetes deployment using a different runtime that doesn't support this feature, for example.

calico_cni_k8s_test.go
```diff
@@ -553,7 +557,60 @@ var _ = Describe("CalicoCni", func() {
...
+			          "nodename_file_optional": true,
+			 	  "ipam": {},
+					"kubernetes": {
+					  "k8s_api_root": "http://127.0.0.1:8080"
+					 },
+					"policy": {"type": "k8s"},
+					"log_level":"info"
+				}`, cniVersion, os.Getenv("ETCD_IP"), os.Getenv("DATASTORE_TYPE"))
+
+					// assignIP := net.IPv4(10, 0, 0, 1).To4()
```
caseydavenport: stray comment?

Makefile
ozdanborne: will do in my next pr

PR #506, opened 2018-04-12
Remove CreateContainerNamespaceWithCid

ozdanborne:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/testutils/utils.go
+++ b/testutils/utils.go
@@ -222,10 +222,6 @@ func RunIPAMPlugin(netconf, command, args, cniVersion string) (*current.Result, 
 }
 
 func CreateContainerNamespace() (containerNs ns.NetNS, containerId string, err error) {
-	return CreateContainerNamespaceWithCid("")
-}
-
-func CreateContainerNamespaceWithCid(container_id string) (containerNs ns.NetNS, containerId string, err error) {
 	containerNs, err = ns.NewNS()
 	if err != nil {
 		return nil, "", err
@@ -233,10 +229,6 @@ func CreateContainerNamespaceWithCid(container_id string) (containerNs ns.NetNS,
 
 	netnsname := path.Base(containerNs.Path())
 	containerId = netnsname[:10]
-
-	if container_id != "" {
-		containerId = container_id
-	}
 
 	err = containerNs.Do(func(_ ns.NetNS) error {
 		lo, err := netlink.LinkByName("lo")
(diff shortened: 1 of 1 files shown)


calico_cni_test.go
```diff
@@ -190,7 +190,7 @@ var _ = Describe("CalicoCni", func() {
 					if err := testutils.CreateHostVeth(containerID, "", "", hostname); err != nil {
 						panic(err)
 					}
-					_, session, _, _, _, contNs, err := testutils.CreateContainerWithId(netconf, "", testutils.TEST_DEFAULT_NS, "", containerID)
+					_, session, _, _, _, contNs, err := testutils.CreateContainer(netconf, "", testutils.TEST_DEFAULT_NS, "")
```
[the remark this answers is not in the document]
ozdanborne: This is where the beauty of this test code truly shines. 

`CreateContainerWithId` just called `CreateContainerNamespaceWithCid` with that ID, and all the latter did was return it: https://github.com/projectcalico/cni-plugin/blob/fb6ab1f8750a380ffd49734c5c2fc0ac08de284b/testutils/utils.go#L237-L239

The Container ID doesn't _have_ to match the veth. `CreateContainer` is returning it as a sample value you can use when you call `RunCNIPluginWithId`, since it what we know Kubernetes will do. But other orchestrators may use other values. So instead of passing a containerID to CreateContainer that will just get passed back to you, I've switched it so that it gives you a containerID back that *you don't have to use if you don't want to*.

I've updated the function descriptions to better explain this. And am considering if we don't even want to return any ContainerID at all and have the caller use some helper if it wants to get the ContainerID that kubernetes would usually assign.

ozdanborne:
I simplified this a bit. Removed `CreateContainerNamespaceWithId` and pushed the ContainerID overriding aspect of that function into the (only) caller function.

testutils/utils.go
```diff
@@ -250,17 +242,30 @@ func CreateContainerNamespaceWithCid(container_id string) (containerNs ns.NetNS,
 }
 
 func CreateContainer(netconf, podName, podNamespace string, ip string) (container_id string, session *gexec.Session, contVeth netlink.Link, contAddr []netlink.Addr, contRoutes []netlink.Route, targetNs ns.NetNS, err error) {
-	return CreateContainerWithId(netconf, podName, podNamespace, ip, "")
+	targetNs, container_id, err = CreateContainerNamespace()
+
```
caseydavenport: extra newline here

testutils/utils.go
```diff
@@ -250,17 +242,30 @@ func CreateContainerNamespaceWithCid(container_id string) (containerNs ns.NetNS,
...
+	return
 }
 
 // Create container with the giving containerId when containerId is not empty
+//
+// Deprecated: Please call CreateContainerNamespace and then RunCNIPluginWithID directly.
 func CreateContainerWithId(netconf, podName, podNamespace, ip, containerId string) (container_id string, session *gexec.Session, contVeth netlink.Link, contAddr []netlink.Addr, contRoutes []netlink.Route, targetNs ns.NetNS, err error) {
-	targetNs, container_id, err = CreateContainerNamespaceWithCid(containerId)
-
+	targetNs, container_id, err = CreateContainerNamespace()
```
caseydavenport: `container_id` is python style (underscores) - I know it was already this way, but....

I'd suggest calling this something like `containerID` and then the arg that's passed in can be `givenContainerID` or something like that to differentiate them.

caseydavenport: (Similar for `targetNs` which should be `targetNS`)

testutils/utils.go
```diff
@@ -250,17 +242,30 @@ func CreateContainerNamespaceWithCid(container_id string) (containerNs ns.NetNS,
 }
 
 func CreateContainer(netconf, podName, podNamespace string, ip string) (container_id string, session *gexec.Session, contVeth netlink.Link, contAddr []netlink.Addr, contRoutes []netlink.Route, targetNs ns.NetNS, err error) {
-	return CreateContainerWithId(netconf, podName, podNamespace, ip, "")
```
caseydavenport: Not for this PR, but for future clean up - these functions are VERY similar, and both take a lot of arguments and return even more.

It would be great to use structs for argument and return values.

e.g. 

```go
func CreateContainer(args CreateArgs) (ContainerParams, error) {
   . . .
}

type CreateArgs struct {
    NetConf string
    PodName string
    PodNamespace string
    IP string
    ContainerID string // Optional
}

type ContainerParams struct {
    ID string
    Session *gexec.Session
    Veth netlink.Link
    Addrs []Addrs
    Routes []Route
    Namespace ns.NetnS
}
```

Not a trivial amount of work, but I think could clean up the code nicely if you're ever bored.

ozdanborne: Agreed, I'm working towards that.

ozdanborne: I actually think `CreateContainer` should be removed entirely as well, and test flow should look like this:

1. Create & configure container ns, check for errors
2. Run calico plugin against it, check for errors

By combining those two things in a single function like CreateContainer does, the error from 1 and 2 are less distinguishable. It will introduce some redundancy in the tests but I think that can be resolved by moving it into beforeEach.

caseydavenport: Sounds fine to me

PR #526, opened 2018-05-22
BUG FIX: Make install-cni.sh check if the etcd cert file exist.

bjhaid:
This will prevent it from terminating because `stat` returns non-zero
when called a non-existing file and also stat_output ends up being an
unbound variable.

Affected Component: install-cni container

@tmjd PTAL not sure if this needs a release note thanks!

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -190,7 +190,7 @@ while [ "$should_sleep" == "true"  ]; do
 	# Kubernetes Secrets can be updated.  If so, we need to install the updated
 	# version to the host. Just check the timestamp on the certificate to see if it
 	# has been updated.  A bit hokey, but likely good enough.
-	if [ "$(ls ${SECRETS_MOUNT_DIR} 2>/dev/null)" ];
+	if [ -e ${SECRETS_MOUNT_DIR}/etcd-cert ];
 	then
         stat_output=$(stat -c%y ${SECRETS_MOUNT_DIR}/etcd-cert 2>/dev/null)
         sleep 10;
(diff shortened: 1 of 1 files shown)


k8s-install/scripts/install-cni.sh
```diff
@@ -191,14 +191,16 @@ while [ "$should_sleep" == "true"  ]; do
 	# version to the host. Just check the timestamp on the certificate to see if it
 	# has been updated.  A bit hokey, but likely good enough.
 	if [ "$(ls ${SECRETS_MOUNT_DIR} 2>/dev/null)" ];
```
tmjd: I think you should just change this line instead of adding a new check.  I don't think there is a need for 2 checks.

bjhaid:
@tmjd 

> Could you clean up the first several lines of the while loop to be spaces too instead of tabs?

The file has more places where it uses tabs and it's a mix of tabs and spaces for indentation should I just clean the entire file and make it use all spaces?

bjhaid:
Added a second commit to consistently use spaces vs tabs

tmjd:
Yeah I'd be fine with cleaning up the whole file.  I just wanted at least that loop to be consistent. Thanks.

> not sure if this needs a release note

I don't think it is necessary.

k8s-install/scripts/install-cni.sh
```diff
@@ -187,18 +187,18 @@ echo "Created CNI config ${CNI_CONF_NAME}"
...
-        if [ "$stat_output" != "$(stat -c%y ${SECRETS_MOUNT_DIR}/etcd-cert 2>/dev/null)" ]; then
-            echo "Updating installed secrets at: $(date)"
-            cp -p ${SECRETS_MOUNT_DIR}/* /host/etc/cni/net.d/calico-tls/
-        fi
-    else
-        sleep 10
+  # Kubernetes Secrets can be updated.  If so, we need to install the updated
+  # version to the host. Just check the timestamp on the certificate to see if it
+  # has been updated.  A bit hokey, but likely good enough.
+  if [ "$(ls ${SECRETS_MOUNT_DIR} 2>/dev/null)" ];
```
tmjd: Your second commit reverted the changes in your first.

bjhaid: doh, fixed thanks!

tmjd:
@caseydavenport WDYT?

PR #527, opened 2018-05-23
Enable config of K8s CA in kubeconfig

tmjd:
## Description

Load the K8s CA into the kubeconfig if it exists.  This also provides the ability to set SKIP_TLS_VERIFY to true which will sets `insecure-skip-tls-verify: true` which was the previous behavior.

## Release Note

```release-note
None required
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -101,17 +101,26 @@ ${CNI_NETWORK_CONFIG:-}
 EOF
 fi
 
+SERVICE_ACCOUNT_PATH=/var/run/secrets/kubernetes.io/serviceaccount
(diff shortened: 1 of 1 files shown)


k8s-install/scripts/install-cni.sh
```diff
@@ -114,6 +117,12 @@ if [ -f "/var/run/secrets/kubernetes.io/serviceaccount/token" ]; then
     echo "KUBERNETES_SERVICE_PORT not set"; exit 1;
   fi
 
+  if [ "$SKIP_TLS_VERIFY" == "true" ]; then
```
neiljerram: FYI, as 'true' and 'false' are commands, you could simplify this to
```
  if $SKIP_TLS_VERIFY; then
```
OTOH, you may still prefer it as it is.

tmjd: I think what is currently there would have better behavior if something other than `true` or `false` was entered, as something else would probably cause a failed command.

PR #528, opened 2018-05-24
Tag amd64 builds with latest for CI

tmjd:
## Description

The semaphore job depends on having calico/cni:latest after the build to tag and push to the repos. With a recent change the build only produced calico/cni:latest-amd64 causing the incorrect images to be tagged and pushed.

This change simply tags amd64 builds as `latest` (in addition to the `latest-amd64`).

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,6 +261,10 @@ test-watch: $(DIST)/calico $(DIST)/calico-ipam run-etcd run-k8s-apiserver
 
 $(DEPLOY_CONTAINER_MARKER): Dockerfile.$(ARCH) build-containerized fetch-cni-bins
 	docker build -f Dockerfile.$(ARCH) -t $(CONTAINER_NAME):latest-$(ARCH) .
+ifeq ($(ARCH),amd64)
+	# Need amd64 builds tagged as :latest because Semaphore depends on that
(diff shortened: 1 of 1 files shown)


deitch:
@tmjd can you hold on this for a few? We had the same issue with `kube-controller`, @fasaxc and I think we have a more generic solution.

deitch:
Also, can you share a link to a recent failed semaphore?

tmjd:
Semaphore isn't failing but it is pushing the incorrect images for the master branch.
For CNI the latest, master, and a couple other recent images are all the same and they should not be.

deitch:
I did some of the latest changes on this, but they shouldn't have been that disruptive. I am trying to understand what they should be and what they became?

deitch:
Oh heck. If this fixes it, just do it. We are going to clean up the ci later anyways. But for now, let's get it in. We do this in `felix` anyways. Go for it.

tmjd:
The semaphore job tags the image `calico/cni` to create the master and version tags for the build and then pushes those.  Currently the build produces calico/cni:latest-amd64 which is not being used when tagging with `docker tag calico/cni calico/cni:master`.

Issue #520, opened 2018-05-15
Race condition: docker rm loses IP allocations


caseydavenport:
Saw this when playing around with `docker rm` commands. Deleting the sandbox container managed to end up in a state where two pods had IP addresses, but the host-local plugin wasn't aware of the allocations.

i.e.

```
$ ip route
default via 10.128.0.1 dev ens4 
10.128.0.1 dev ens4  scope link 
172.17.0.0/16 dev docker0  proto kernel  scope link  src 172.17.0.1 linkdown 
192.168.0.0/24 via 10.128.0.34 dev tunl0  proto bird onlink 
192.168.1.0/24 via 10.128.0.36 dev tunl0  proto bird onlink 
blackhole 192.168.2.0/24  proto bird 
192.168.2.2 dev cali7f038ce39ed  scope link 
192.168.2.3 dev calie3a7f661282  scope link
```

```
$ ls /var/lib/cni/networks/k8s-pod-network/
192.168.2.3  last_reserved_ip.0
```
## Expected Behavior

The number of allocated IPs should always match the number of running pods. 

## Possible Solution

My guess is that this is the bit of code that's not behaving properly - https://github.com/projectcalico/cni-plugin/blob/master/k8s/k8s.go#L62-L73

I think it needs to release and re-allocate the IP address for that pod. Otherwise the IP is still attached to the old container ID and will eventually be GC'd. 

This likely applies to Calico IPAM too as of this change: https://github.com/projectcalico/cni-plugin/pull/445 

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* CNI plugin version (e.g. `/opt/cni/bin/calico --version`):
* Orchestrator version (e.g. kubernetes, mesos, rkt):
* CNI config:
* Operating System and version:
* Link to your project (optional):

calixwu:
I don't understand why this would happen with the codes below:

```
if endpointAlreadyExisted {
		// This happens when Docker or the node restarts. K8s calls CNI with the same parameters as before.
		// Do the networking (since the network namespace was destroyed and recreated).
		// There's an existing endpoint - no need to create another. Find the IP address from the endpoint
		// and use that in the response.
		result, err = utils.CreateResultFromEndpoint(endpoint)
		if err != nil {
			return nil, err
		}
		logger.WithField("result", result).Debug("Created result from existing endpoint")
		// If any labels changed whilst the container was being restarted, they will be picked up by the policy
[...]

PR #529, opened 2018-05-24
Always allocate a new IP on k8sCmdAdd

caseydavenport:
## Description

Fixes https://github.com/projectcalico/cni-plugin/issues/520 by always assigning a new IP address to a pod when ADD is called.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -33,6 +33,7 @@ import (
 	k8sconversion "github.com/projectcalico/libcalico-go/lib/backend/k8s/conversion"
 	calicoclient "github.com/projectcalico/libcalico-go/lib/clientv3"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
 	"github.com/projectcalico/libcalico-go/lib/options"
 	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


k8s/k8s.go
```diff
@@ -58,235 +58,221 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
...
-			if len(result.IPs) == 0 {
-				utils.ReleaseIPAllocation(logger, conf.IPAM.Type, args.StdinData)
-				return nil, errors.New("IPAM plugin returned missing IP config")
-			}
+	// Switch based on which annotations are passed or not passed.
+	switch {
+	case ipAddrs == "" && ipAddrsNoIpam == "":
+		// If the endpoint already exists, we need to attempt to release the previous IP addresses here
+		// since the ADD call will fail when it tries to reallocate the same IPs.
+		// TODO
```
gunjan5: 👀

k8s/k8s.go
```diff
@@ -58,235 +58,221 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
...
+	}
+	endpoint.Name = epIDs.WEPName
+	endpoint.Namespace = epIDs.Namespace
+	endpoint.Labels = labels
+	endpoint.Spec.Endpoint = epIDs.Endpoint
+	endpoint.Spec.Node = epIDs.Node
+	endpoint.Spec.Orchestrator = epIDs.Orchestrator
+	endpoint.Spec.Pod = epIDs.Pod
+	endpoint.Spec.Ports = ports
+	endpoint.Spec.IPNetworks = []string{}
```
gunjan5: what's this for?

k8s/k8s.go
```diff
@@ -58,235 +58,221 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
...
+	// maybeReleaseIPAM cleans up any IPAM allocations on failure.
 	maybeReleaseIPAM := func() {
 		logger.Debug("Checking if we need to clean up IPAM.")
-		logger := logger.WithField("IPs", endpoint.Spec.IPNetworks)
-		if endpointAlreadyExisted {
-			logger.Info("Not cleaning up IPAM allocation; this was a pre-existing endpoint.")
-			return
-		}
-		logger.Info("Releasing IPAM allocation after failure")
+		logger := logger.WithField("endpointIPs", endpoint.Spec.IPNetworks)
```
gunjan5: I think you can just do `logger.WithField("...").Info("...")` here

k8s/k8s.go
```diff
@@ -336,17 +322,16 @@ func CmdDelK8s(ctx context.Context, c calicoclient.Interface, epIDs utils.WEPIde
...
 		}
 
-		// Check if ActiveInstanceID is populated (it will be an empty string "" if it was populated
+		// Check if ContainerID is populated (it will be an empty string "" if it was populated
 		// before this field was added to the API), and if it is there then compare it with ContainerID
 		// passed by the orchestrator to make sure they are the same, return without deleting if they aren't.
 	} else if wep.Spec.ContainerID != "" && args.ContainerID != wep.Spec.ContainerID {
-		logger.WithField("WorkloadEndpoint", wep).Warning("CNI_CONTAINERID does not match WorkloadEndpoint ActiveInstanceID so ignoring the DELETE cmd.")
-		return nil
+		logger.WithField("WorkloadEndpoint", wep).Warning("CNI_CONTAINERID does not match WorkloadEndpoint ConainerID, don't delete WEP.")
```
gunjan5: shouldn't we return after this log? (according to the comment above)

k8s/k8s.go
caseydavenport: whoops

k8s/k8s.go
caseydavenport: We need to reset the IPNetworks, since we're assigning a new IP address. We'll populate the nets from the result a few lines below. 

If we didn't do this, the endpoint would keep the old IP from the previous iteration in its spec.

k8s/k8s.go
caseydavenport: Nope, the comment is out of date now. We used to return, but now we let execution continue below so we can clean up the IPAM info.

k8s/k8s.go
```diff
@@ -388,7 +377,36 @@ func CmdDelK8s(ctx context.Context, c calicoclient.Interface, epIDs utils.WEPIde
...
+	ipList, err := validateAndExtractIPs(ipAddrs, "cni.projectcalico.org/ipAddrs", logger)
+	if err != nil {
+		return err
+	}
+
+	// For each IP, call out to Calico IPAM to release it.
+	for _, ip := range ipList {
+		log := logger.WithField("IP", ip)
+		log.Info("Releasing explicitly requested address")
+		unallocated, err := calico.IPAM().ReleaseIPs(context.Background(), []cnet.IP{{ip}})
```
gunjan5: wondering if you can avoid this loop if you just did `ReleaseIPs(ctx, []cnet.IP{ipList...}` something like that? since ReleaseIPs can release multiple IPs

k8s/k8s.go
gunjan5: I thought so, but was thinking of a case (haven't fully thought of it, just brainstorming with you here) if we "unassign" the IP here and if releasing the IP fails, what happens then

k8s/k8s.go
```diff
@@ -313,46 +305,41 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
...
+		logger.WithField("WorkloadEndpoint", epIDs.WEPName).Warning("WorkloadEndpoint does not exist in the datastore, moving forward with the clean up")
 	} else if wep.Spec.ContainerID != "" && args.ContainerID != wep.Spec.ContainerID {
-		logger.WithField("WorkloadEndpoint", wep).Warning("CNI_CONTAINERID does not match WorkloadEndpoint ActiveInstanceID so ignoring the DELETE cmd.")
-		return nil
-
-		// Delete the WorkloadEndpoint object from the datastore.
-		// In case of k8s, where we are deleting the WEP we got from the Datastore,
-		// this Delete is a Compare-and-Delete, so if *any* field in the WEP changed from
-		// the time we get WEP until here then the Delete operation will fail.
+		// If the ContainerID is populated and doesn't match the CNI_CONATINERID provided for this execution, then
```
gunjan5: lol I was waiting to see when you'd move the comments inside the `if...` block :trollface:

k8s/k8s.go
gunjan5: that's what the `...` is for, it'll expand the list. Haven't tried it, but it's ok if that doesn't work

gunjan5:
don't see a release note, just see the label.

k8s/k8s.go
caseydavenport: But the contents of ipList aren't `cnet.IP` objects, they're `net.IP` objects, so we'd need to do something like 

`[]cnet.IP{cnet.IP{ip} for net in ipList}`, so we're iterating over the list anyway.

calixwu:
Won't it cause "IP leak" if assigning a new IP address to a pod when ADD is called?

If I run the ``docker rm`` command to delete the sandbox container, kubelet will call ADD command of cni again and be allocated a new IP. But kubelet seems not to call DEL command of cni at that meantime for releasing the old IP.

PR #535, opened 2018-05-25
Add GCR images to standard tag/push targets

caseydavenport:
## Description

Will make publishing updates to GKE easier. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,9 +104,21 @@ endif
 push: imagetag
 	docker push $(CONTAINER_NAME):$(IMAGETAG)-$(ARCH)
 	docker push quay.io/$(CONTAINER_NAME):$(IMAGETAG)-$(ARCH)
+
+	# Push images to gcr.io, used by GKE.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -107,6 +107,12 @@ push: imagetag
 ifeq ($(ARCH),amd64)
 	docker push $(CONTAINER_NAME):$(IMAGETAG)
 	docker push quay.io/$(CONTAINER_NAME):$(IMAGETAG)
+
+	# Push images to gcr.io, used by GKE.
+	gcloud --project projectcalico-org docker -- push gcr.io/projectcalico-org/$(CONTAINER_NAME):$(IMAGETAG)
```
fasaxc: I think you should probably push the "-<arch>" suffixed versions too; it'll make the multi-arch migration easier later.

caseydavenport: @fasaxc done

deitch:
Oh, I like this. And I like how easy it is to do with the targets of `make tagimages` and `make push`. :-)

PR #547, opened 2018-06-10
Replicate all semaphore activities in ci and cd targets

deitch:
## Description

Replicate the workflow from kube-controllers PR here. It already had a `ci` target (which Semaphore already used), so just then `cd` side.

* Create a make cd target that only tags images and pushes to registries

As discussed with @caseydavenport @fasaxc @tomdee

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -301,11 +301,23 @@ test-install-cni: image k8s-install/scripts/install_cni.test
 	cd k8s-install/scripts && CONTAINER_NAME=$(CONTAINER_NAME) ./install_cni.test
 
 ###############################################################################
-# CI
+# CI/CD
 ###############################################################################
 .PHONY: ci
 ## Run what CI runs
 ci: clean static-checks test-cni-versions image test-install-cni
+
+## Deploys images to registry
+cd:
+ifndef CONFIRM
+	$(error CONFIRM is undefined - run using make <target> CONFIRM=true)
+endif
+ifndef BRANCH_NAME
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -307,6 +307,18 @@ test-install-cni: image k8s-install/scripts/install_cni.test
 ## Run what CI runs
 ci: clean static-checks test-cni-versions image test-install-cni
 
+## Deploys images to registry
```
caseydavenport: @deitch I think we're going with these big block headings now - could you add one here for CD (I believe this is the convention @tomdee has been adding throughout)

deitch: Sure. The double- comment `##` causes it to show up when you do `make help` (from [this line](https://github.com/deitch/calico-cni-plugin/blob/96659bfc872f2e38a32b9cf50af826aa347008f0/Makefile#L424) ) and happy to add the block heading too. Coming in one minute.

deitch: Wait a minute, there is one here. See [this](https://github.com/deitch/calico-cni-plugin/blob/96659bfc872f2e38a32b9cf50af826aa347008f0/Makefile#L303-L305) . Is this not good enough?

tomdee: Yep, I've just been adding /CD to that block heading (https://github.com/projectcalico/confd/blob/master/Makefile#L295) so lets go with that

deitch: Done

PR #571, opened 2018-08-01
[WiP] Add support for host-local IPAM routes.

fasaxc:
## Description

If the host-local IPAM plugin is in use and some routes
are specified then we only program those routes, not out default
0.0.0.0/0 and ::/0 routes.

The gateway of the route is ignored and we use the Calico gateway
instead since that's the only valid peer on the link.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


calico_cni_k8s_test.go
```diff
@@ -842,95 +847,229 @@ var _ = Describe("CalicoCni", func() {
...
+					  },
+					  "policy": {"type": "k8s"},
+					  "log_level":"debug"
+					}`,
+					expectedV4Routes: []string{
+						regexp.QuoteMeta("default via 169.254.1.1 dev eth0"),
+						regexp.QuoteMeta("169.254.1.1 dev eth0 scope link"),
+					},
+					expectedV6Routes: []string{
+						"dead:beef::. dev eth0  metric 256",
```
neiljerram: Wondering what `.` will match here.  Is it just `0`?

calico_cni_k8s_test.go
```diff
@@ -842,95 +847,229 @@ var _ = Describe("CalicoCni", func() {
...
+					          }
+					       ],
+					       [
+					          {
+					            "subnet": "dead:beef::/96"
+					          }
+					       ]
+					    ],
+					    "routes": [
+					      {"dst": "10.123.0.0/16"},
```
neiljerram: Can there be a gateway field here?  I understand that we don't use it, but should it be here to check that we can ignore it?

calico_cni_k8s_test.go
```diff
@@ -842,95 +847,229 @@ var _ = Describe("CalicoCni", func() {
...
 					  "type": "calico",
 					  "etcd_endpoints": "http://%s:2379",
 					  "datastore_type": "%s",
 					  "ipam": {
 					    "type": "host-local",
-						"subnet": "usePodCidr"
+					    "ranges": [
+					       [
+					          {
+					            "subnet": "usePodCidr"
```
neiljerram: Am I right that PodCidr is always v4, and that we only support `usePodCidr` for substituting a v4 CIDR?

fasaxc: I think so, that's certainly the only requirement that I'm aware of here.

utils/network.go
```diff
@@ -29,8 +29,39 @@ import (
...
+	// IPv6AllNet represents the IPv6 all-addresses CIDR ::/0.
+	IPv6AllNet    *net.IPNet
+	DefaultRoutes []*net.IPNet
+)
+
+func init() {
+	var err error
+	_, IPv4AllNet, err = net.ParseCIDR("0.0.0.0/0")
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
fasaxc: The app will crash with a panic.

utils/utils.go
```diff
@@ -165,6 +170,59 @@ func CleanUpIPAM(conf types.NetConf, args *skel.CmdArgs, logger *logrus.Entry) e
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func replaceHostLocalIPAMPodCIDR(logger *logrus.Entry, rawIpamData interface{}, getPodCidr func() (string, error)) error {
+	logrus.WithField("ipamData", rawIpamData).Debug("Examining IPAM data for usePodCidr")
+	ipamData := rawIpamData.(map[string]interface{})
```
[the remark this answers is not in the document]
fasaxc: There _some_ logic.  Where I've used `_`, I'm handling the error case by checking for the zero value instead.  For example
```
subnet, _ := ipamData["subnet"].(string)
if subnet == ...
```
I.e. we only substitute the string "usePodCidr", otherwise we defer to the IPAM plugin to validate that field.

The unchecked cast of `stdinData["ipam"]` came from the existing code that I was updating but I might as well make that consistent.

calico_cni_k8s_test.go
neiljerram: So is it OK to rely on that number being between 0 and 9?  You'd need e.g. `.*` to allow multiple digits?

utils/network.go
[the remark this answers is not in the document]
neiljerram: I guess I was worrying if that would be a new scenario; if we know how k8s as a whole reacts if that happens...

But never mind, because this is a panic that really *won't* ever happen anyway.

PR #572, opened 2018-08-02
Support a configuring the config template as a file

fasaxc:
## Description

Previously, the config had to be specified inline as an environment variable.

While adding the tests, I found that the existing tests were silently failing to read files and hence weren't actually testing anything!  I've fixed that up.

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


k8s-install/scripts/install_cni_test.go
```diff
@@ -143,38 +144,7 @@ var _ = Describe("install-cni.sh tests", func() {
...
 				err := runCniContainer()
 				Expect(err).NotTo(HaveOccurred())
-
-				expectedFile := "expected_10-calico.conf"
-
-				var expected = []byte{}
-				if file, err := os.Open(expectedFile); err != nil {
-					Fail("Could not open " + expectedFile + " for reading")
-				} else {
-					_, err := file.Read(expected)
```
fasaxc: expected was a 0-length slice at this point and `file.Read()`'s contract says that it reads at most `len(<input slice>)` bytes; hence it was reading 0 bytes!

k8s-install/scripts/expected_10-calico.conf.alternate
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "alternate",
+    "type": "calico",
+    "etcd_endpoints": "",
+    "etcd_key_file": "",
```
neiljerram: Does the use case that you're addressing require substitution of alternate `__ETCD_*__` variables here?  If so, if would be good to test that with a case where the substitution is not `""`.

k8s-install/scripts/install_cni_test.go
```diff
@@ -39,6 +39,7 @@ func runCniContainer(extraArgs ...string) error {
 		"-e", "SLEEP=false",
 		"-e", "KUBERNETES_SERVICE_HOST=127.0.0.1",
 		"-e", "KUBERNETES_SERVICE_PORT=8080",
+		"-v", cwd + ":/template",
```
neiljerram: Hmm, so files under `cwd + "/tmp"` will be mapped into the container in two different places.  Does that work?  Does it matter?  It feels slightly inelegant to me; could you introduce another subdirectory of `cwd`, and only map that subdir to `/template`?

k8s-install/scripts/install_cni_test.go
```diff
@@ -246,3 +207,13 @@ var _ = Describe("install-cni.sh tests", func() {
...
 })
+
+func expectFilesEqual(filenameExpected, filenameActual string) {
+	expected, err := ioutil.ReadFile(filenameExpected)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "failed to read expected file "+filenameExpected)
+	actual, err := ioutil.ReadFile(filenameActual)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred(), "failed to read actual file "+filenameActual)
+
+	ExpectWithOffset(1, actual).To(Equal(expected), fmt.Sprintf(
+		"actual file (%s) differed from expected file (%s)", filenameActual, filenameExpected))
```
neiljerram: Can you log out the content diff here?  It feels likely someone would want to see that.

k8s-install/scripts/install_cni_test.go
neiljerram: :-) And unnecessary anyway, given that they should use ioutil.ReadFile.

k8s-install/scripts/install_cni_test.go
fasaxc: Gomega does that

k8s-install/scripts/expected_10-calico.conf.alternate
fasaxc: I'm only really testing that _some_ values get substituted here; the substitution code is the same whatever the source of the data since the data gets written out to the temporary file before being processed.  I think etcd substitutions can be safely tested orthogonally to my change.

k8s-install/scripts/install_cni_test.go
neiljerram: D'oh, of course.

k8s-install/scripts/install_cni_test.go
fasaxc: Moved to a subdir

PR #573, opened 2018-08-10
Fix that CNI plugin allowed at most one IPv4 and one IPv6 address.

fasaxc:
## Description

- Move the per-IP -version set-up (sysctls and routes) out of the loop over IPs.
- Simplify the loop over IPs (since there's no reliance on IP version in there any more)
- Add asserts to existing tests to verify number of IPs assigned.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


neiljerram:
Oops, is this something that I missed in my previous CNI reviews this sprint?

PR #581, opened 2018-08-27
Exit if the plugin is taking too long

tomdee:
- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
Shut down CNI plugin if execution takes over 30 seconds
```

Fixes projectcalico/calico#2098

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -121,11 +121,7 @@ func CleanUpNamespace(args *skel.CmdArgs, logger *logrus.Entry) error {
 					return err
 				})
 
-				if err != nil {
-					ch <- err
-				} else {
-					ch <- nil
-				}
+				ch <- err
 			}()
 
 			select {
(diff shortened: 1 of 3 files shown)


tomdee:
Based on some more digging, the offending line is https://github.com/projectcalico/cni-plugin/blob/5938bbbf1a2fc8ac4f4a7f742459abd0c9dc8e90/utils/utils.go#L115 (or possibly the lien above).

This code is being called when the CNI `delete` method is called (when the container is torn down). 

It's the last things that happens and it only occurs _after_ the IP address is released.

utils/utils.go
```diff
@@ -111,13 +112,31 @@ func CleanUpNamespace(args *skel.CmdArgs, logger *logrus.Entry) error {
...
 
-			if err != nil {
-				return err
+			go func() {
+				err := ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+					_, err := ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+					return err
+				})
+
+				if err != nil {
```
caseydavenport: This could be simplified to be `ch <- err`

utils/utils.go
```diff
@@ -111,13 +112,31 @@ func CleanUpNamespace(args *skel.CmdArgs, logger *logrus.Entry) error {
...
+
+			select {
+			case err := <-ch:
+				if err != nil {
+					return err
+				} else {
+					fmt.Fprintf(os.Stderr, "Calico CNI deleted device in netns %s\n", args.Netns)
+				}
+			case <-time.After(5 * time.Second):
+				fmt.Fprintf(os.Stderr, "Calico CNI timed out deleting device in netns %s\n", args.Netns)
```
caseydavenport: Did you consider returning an error in this scenario?

I suppose it's probably OK, but might we end up leaking devices this way?

PR #587, opened 2018-09-04
Always allocate a new IP

caseydavenport:
## Description

Reimplementing #529 on the v2.6 branch.

- Always allocate a new IP address for pods, even when the wep exists.
- Always deallocate IP addresses for a given container ID, even if the wep doesn't exist or doesn't match.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix a bug in the CNI plugin where IP allocations were not properly updated on container restart
```

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -1045,16 +1045,22 @@ var _ = Describe("CalicoCni", func() {
 					Expect(err).ShouldNot(HaveOccurred())
 				})
 
-				It("a second ADD for the same container should be a no-op", func() {
-					// Try to create the same container (so CNI receives the ADD for the same endpoint again)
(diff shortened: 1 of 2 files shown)


calixwu:
Won't it cause "IP leak" if assigning a new IP address to a pod when ADD is called?

For example. If I run the ``docker rm`` command to delete the sandbox container, kubelet will call ADD command of cni again and allocate a new IP to that pod. But kubelet seems not to call DEL command of cni at that meantime for releasing the old IP.

caseydavenport:
Modern kubelets (as of many releases ago) will call DEL for exited containers.

But yes, if you cleanup the sandbox yourself kubelet won't know to do it, and will leak an IP. You shouldn't clean up the sandbox by hand, though!

PR #604, opened 2018-09-26
Resolve IP Pool names to CIDRs in IPAM plugin

caseydavenport:
## Description

Support using IP Pool names in the current namespace / pod annotations and CNI config instead of requiring the use of a CIDR explicitly. 

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Support specifying IP Pool by name in CNI config and Kubernetes annotations
```


caseydavenport:
@tomdee Still working on tests, but I think at this point it's worth looking at the approach.

PR #605, opened 2018-09-27
Disable floating IPs in k8s by default

caseydavenport:
## Description

Disable this feature by default, since most users won't need it, and it allows for arbitrary pods to pretend to be other pods.

This feature is new in Calico v3.3.0 (unlreleased) as of this PR: https://github.com/projectcalico/cni-plugin/pull/419

Can be enabled in the CNI config:

```
feature_control: {
  floating_ips: true,
},
```

Corresponding docs: https://github.com/projectcalico/calico/pull/2213

Also added in some releaseIPAM() calls, since we were missing those branches.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/k8s/k8s.go
+++ b/k8s/k8s.go
@@ -390,12 +390,19 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 	endpoint.Spec.InterfaceName = hostVethName
 	endpoint.Spec.ContainerID = epIDs.ContainerID
 	logger.WithField("endpoint", endpoint).Info("Added Mac, interface name, and active container ID to endpoint")
+
 	// List of DNAT ipaddrs to map to this workload endpoint
 	floatingIPs := annot["cni.projectcalico.org/floatingIPs"]
 
 	if floatingIPs != "" {
+		// If floating IPs are defined, but the feature is not enabled, return an error.
+		if !conf.FeatureControl.FloatingIPs {
+			releaseIPAM()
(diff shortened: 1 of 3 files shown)


k8s/k8s.go
```diff
@@ -390,12 +390,19 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
...
 	endpoint.Spec.ContainerID = epIDs.ContainerID
 	logger.WithField("endpoint", endpoint).Info("Added Mac, interface name, and active container ID to endpoint")
+
 	// List of DNAT ipaddrs to map to this workload endpoint
 	floatingIPs := annot["cni.projectcalico.org/floatingIPs"]
 
 	if floatingIPs != "" {
+		// If floating IPs are defined, but the feature is not enabled, return an error.
+		if !conf.FeatureControl.FloatingIPs {
+			releaseIPAM()
```
neiljerram: Hmm, a 400-line function....

These `releaseIPAM()` calls seem to be crying out for a `defer` being used instead.  Is there a point in this function where it becomes necessary to call `releaseIPAM` (or `ReleaseIPAllocation`) on return?  If so, can we just put a `defer` at that point, and then delete all the other calls?

caseydavenport: Yeah, I'd desperately like to find some time to refactor some of this code - it's a bit of a mess right now, but it works.

I've pushed a new commit that uses `defer` - It's not obvious to me that it's less error prone though, since we now need to make sure not to mask the `err` variable. WDYT?

PR #617, opened 2018-10-13
Support update-libcalico on branches other than master

caseydavenport:
## Description

@fasaxc WDYT? Until now, we've only supported this on master.

Adding this means we should be able to have Semaphore automtaically update libcalico for us on all release branches.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


fasaxc:
SGTM

PR #621, opened 2018-10-17
Rev CNI dependencies to recent master.

fasaxc:
## Description

Pick up API changes on way to v0.4 support.

Bit of refactoring to make it easier to support other platforms in future.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/calico.go
+++ b/calico.go
@@ -500,5 +500,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	skel.PluginMain(cmdAdd, cmdDel, cniSpecVersion.All)
+	skel.PluginMain(cmdAdd, nil, cmdDel,
+		cniSpecVersion.PluginSupports("0.1.0", "0.2.0", "0.3.0", "0.3.1"),
+		"Calico CNI plugin "+VERSION)
 }
--- a/utils/network_linux.go
+++ b/utils/network_linux.go
@@ -19,6 +19,7 @@ import (
 	"net"
 	"os"
 	"syscall"
+	"time"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types/current"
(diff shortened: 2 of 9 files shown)


utils/network_linux.go
```diff
@@ -416,3 +417,47 @@ func writeProcSys(path, value string) error {
...
+		})
+
+		if devErr == nil {
+			fmt.Fprintf(os.Stderr, "Calico CNI deleting device in netns %s\n", args.Netns)
+			// Deleting the veth has been seen to hang on some kernel version. Timeout the command if it takes too long.
+			ch := make(chan error, 1)
+
+			go func() {
+				err := ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+					return ip.DelLinkByName(args.IfName)
```
fasaxc: Note to reviewer: this line changed vs the version in utils.go that I removed.

Old line was `_, err := ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)` but that method changed signature so I had to fix up and it looked like `ip.DelLinkByName()` was a better fit for what we wanted here anyway?

calico.go
```diff
@@ -500,5 +500,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	skel.PluginMain(cmdAdd, cmdDel, cniSpecVersion.All)
+	skel.PluginMain(cmdAdd, nil, cmdDel,
+		cniSpecVersion.PluginSupports("0.1.0", "0.2.0", "0.3.0", "0.3.1"),
```
fasaxc: Note: this excludes v0.4.0 so the framework will not allow us to be called for CHECK (until we add that later).

utils/network_linux.go
caseydavenport: Seems reasonable.

PR #626, opened 2018-10-24
Support interop with Azure CNI IPAM plugin.

caseydavenport:
## Description

Add support for Azure IPAM, which requires a few things:
- that we pass in the allocated subnet from the previous call.
- on delete, also pass in any IP addresses associated with the container.

In order to do that, we need to store information about azure endpoints as well as networks so that we can provide that information to the IPAM plugin when we call it.

This PR implements that by writing configuration to disk in `/var/run/calico/azure/<network>/`

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Calico CNI support for the azure-vnet-ipam plugin
```


k8s/k8s.go
```diff
@@ -279,6 +295,18 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
...
+			subnetPrefix := result.IPs[0].Address
+			subnetPrefix.IP = subnetPrefix.IP.Mask(subnetPrefix.Mask)
+			fn := fmt.Sprintf(azureFilenameTemplate, conf.Name)
+			if err := ioutil.WriteFile(fn, []byte(subnetPrefix.String()), 0644); err != nil {
+				return nil, err
+			}
+			logger.Infof("Stored azure subnet on disk: %s", subnetPrefix)
+		}
+
 	case ipAddrs != "" && ipAddrsNoIpam != "":
```
fasaxc: Worth erroring-out if the user tries to mix Azure IPAM with incompatible options?

k8s/k8s.go
```diff
@@ -828,3 +867,30 @@ func getPodCidr(client *kubernetes.Clientset, conf types.NetConf, nodename strin
 	}
 	return node.Spec.PodCIDR, nil
 }
+
+// handleAzureIPAM mutates the given args to include the Azure IPAM subnet, if appropriate.
+func handleAzureIPAM(logger *logrus.Entry, args *skel.CmdArgs, filename string) error {
```
fasaxc: `handle` always seems like a copy out :-P.  How about `adjustAzureIPAMConfig` or `fixUpAzureIPAMConfig` or something like that?

internal/pkg/azure/endpoint.go
```diff
@@ -0,0 +1,56 @@
...
+	Interface   string
+	Addresses   []string
+}
+
+func (ae *AzureEndpoint) Write() error {
+	bytes, err := json.Marshal(ae)
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(ae.filename(), bytes, 0600); err != nil {
```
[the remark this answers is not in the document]
fasaxc: ah, looks like it gets created elsewhere

caseydavenport:
> Where does the Azure IPAM plugin store its state; if it stores on disk too then this approach would be a nice fit. OTOH, if it stores its state in k8s CRDs.... (I suspect the answer is neither?)

Yup, the Azure IPAM plugin stores its state to disk as well in the `/var/run` directory, which is why I've done the same for this PR and thought it was a safe bet :)

From my own experimentation, it seems rebooting the node clears all IP allocations in the Azure IPAM plugin. The IPAM plugin keeps a local pool of IPs and manages them locally rather than relying on allocations being stored centrally.

fasaxc:
OK, that sounds fine then.

PR #642, opened 2018-11-07
Adding MTU tests in reference to projectcalico/calico #2200

insequent:
## Description
This is a simple testing change that specifically checks custom MTU values. It was created in an attempt to replicate https://github.com/projectcalico/calico/issues/2200

### DIFF
--- a/calico_cni_k8s_test.go
+++ b/calico_cni_k8s_test.go
@@ -398,6 +398,83 @@ var _ = Describe("Kubernetes CNI tests", func() {
 			})
 		})
 
+		Context("when calico-config contains a custom mtu", func() {
+			mtuNetconfTemplate := `
+			{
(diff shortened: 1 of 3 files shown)


insequent:
Try as I might, I cannot get the git diff to not look crazy for `calico_cni_test.go`, despite there being minimal changes. Anyways, changes since last PR:

* Removed `Describe("Run Calico CNI plugin"` from all tests
* Rearranged `Context("deprecate Hostname for nodename"` tests (there were 2) to be together under the same context.
* Minor comment update to utils' DetermineNodeName. It seems priorities had changed there, but the comment hadn't. Side note: if it's a priority list, we should start with the highest priority and return if a value is found, instead of rewriting the same variable and checking everything? Trying not to let my scope creep too much in this PR...

PR #652, opened 2018-11-08
Make CNI plugin importable; standardise package structure.

fasaxc:
## Description

Making the CNI plugin importable will make it possible to pull it in as an import in the node repo more easily.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


Makefile
```diff
@@ -196,8 +196,8 @@ $(BIN)/calico $(BIN)/calico-ipam: $(SRCFILES) vendor
 	-w /go/src/$(PACKAGE_NAME) \
 	-e GOCACHE=/go-cache \
 		$(CALICO_BUILD) sh -c '\
-			go build -v -o $(BIN)/calico -ldflags "-X main.VERSION=$(GIT_VERSION) -s -w" calico.go ; \
-            go build -v -o $(BIN)/calico-ipam -ldflags "-X main.VERSION=$(GIT_VERSION) -s -w" ipam/calico-ipam.go'
+			go build -v -o $(BIN)/calico -ldflags "-X github.com/projectcalico/cni-plugin/pkg/cniplugin.VERSION=$(GIT_VERSION) -s -w" ./cmd/calico && \
```
caseydavenport: I think in our other projects we keep the VERSION as part of the main package. 

https://github.com/projectcalico/felix/blob/master/cmd/calico-felix/calico-felix.go#L40
https://github.com/projectcalico/node/blob/master/cmd/calico-node/main.go#L88

fasaxc: I think those a both counter-examples :-P but I do prefer `-X main.VERSION` over the convoluted package name (and make it easier to build from the other package) so making that change.

I don't think it makes sense to move argument parsing into the main package for now; we  can always factor that out later (splitting `Main()` into `Main()`, which does arg parsing and a second method that skips it, if it becomes useful.

cmd/calico/calico-main.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2018 Tigera, Inc. All rights reserved.
```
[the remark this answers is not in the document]
fasaxc: I don't mind; I wanted a different name to the other calico.go so it's easier to search for in goland.  I've renamed the other file  to plugin.go to match it's new package name.

Technically, go wants the directory name to match and the file name is irrelevant.

fasaxc:
Back to you @caseydavenport; I spotted and cleaned up some mixed whitespace errors in the Makefile too.

Makefile
caseydavenport: Hah, yeah you're right they are! Sigh...

PR #664, opened 2018-11-17
Add a target to help automate cherry-picks

caseydavenport:
## Description

Wanted to try this out and see what folks thought. Kubernetes has a pretty nice tool for helping create cherry-pick PRs, and it seems to "just work" against Calico repos too!

Trying it as a make target so that we handle the magic of downloading the latest script, removing the need to check it in.

Example:

```
make FORK_REMOTE=cd4 GITHUB_USER=caseydavenport cherry-pick TARGET_BRANCH=release-v3.3 PR_NUMBER=662
```

Here's an example PR created this way: https://github.com/projectcalico/cni-plugin/pull/663

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ install_cni.test
 *.swp
 *.coverprofile
 report/*.xml
+cherry_pick_pull.sh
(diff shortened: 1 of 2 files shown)


caseydavenport:
Going to close this for now and think a little bit more, but it's an option if we want it.

PR #666, opened 2018-11-23
Automatic update of GO_BUILD to v0.20

fasaxc:
The new linter was complaining about all our `Fprintf` calls, since they return an error, which we ignore.  Might be missing something but I can't see a good reason not to use the logger now it's threaded through the code; I guess the Fprintf output is a bit more "crafted" but at this stage it's mixed in with a lot of log output so maybe better to be consistent and use the logger everywhere?


caseydavenport:
@fasaxc yeah I think we should just use the logger now.

For context, I believe this is something we did a long while back so the output when running `rkt` containers was pretty - these "milestone" logs were printed to screen and other logs were not.

But that's not really common.

PR #680, opened 2019-02-12
Store additional attributes with IP allocations

caseydavenport:
Stores pod name/namespace as well as the node of allocation.

Requires this guy: https://github.com/projectcalico/libcalico-go/pull/1035

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/pkg/ipamplugin/ipam_plugin.go
+++ b/pkg/ipamplugin/ipam_plugin.go
@@ -109,6 +109,13 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	// We attach important attributes to the allocation.
+	attrs := map[string]string{ipam.AttributeNode: nodename}
+	if epIDs.Pod != "" {
+		attrs[ipam.AttributePod] = epIDs.Pod
+		attrs[ipam.AttributeNamespace] = epIDs.Namespace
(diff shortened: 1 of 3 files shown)


briansan:
What is the intended result of added these 3 fields to the Attributions map on AutoAssign? IIUC, the only code I found using this map is in the `findOrAddAttribute` method n `lib/ipam/ipam_block.go` which seems to be doing things only with the handleID..

briansan:
Ah, I see that this goes along with https://github.com/projectcalico/kube-controllers/pull/340

pkg/ipamplugin/ipam_plugin.go
```diff
@@ -109,6 +109,16 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	// We attach important attributes to the allocation.
+	attrs := map[string]string{
+		"node": nodename,
+	}
+	if epIDs.Pod != "" {
+		attrs["pod"] = epIDs.Pod
```
caseydavenport: Need to decide if we should just include the "wep" field and rebuild the pod / namespace from that, or if it is more useful to include the pod/namespace directly.

PR #683, opened 2019-02-14
calico-ipam-upgrade Implementation

briansan:
## Description
- adds a new binary `calico-ipam-upgrade` for migrating the data necessary to upgrade KDD-backed calico installations from using host-local IPAM to Calico IPAM

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

### DIFF
--- a/internal/pkg/utils/utils.go
+++ b/internal/pkg/utils/utils.go
@@ -454,15 +454,15 @@ func GetIdentifiers(args *skel.CmdArgs, nodename string) (*WEPIdentifiers, error
 	return &epIDs, nil
 }
 
-func GetHandleID(netName string, containerID string, workload string) (string, error) {
+func GetHandleID(netName, containerID, workload string) string {
 	handleID := fmt.Sprintf("%s.%s", netName, containerID)
 	logrus.WithFields(logrus.Fields{
-		"Network":     netName,
-		"ContainerID": containerID,
-		"Workload":    workload,
 		"HandleID":    handleID,
+		"Network":     netName,
+		"Workload":    workload,
+		"ContainerID": containerID,
 	}).Debug("Generated IPAM handle")
-	return handleID, nil
+	return handleID
 }
 
 func CreateClient(conf types.NetConf) (client.Interface, error) {
--- a/pkg/upgrade/migrate.go
+++ b/pkg/upgrade/migrate.go
@@ -1,0 +1,341 @@
+// Copyright 2015 Tigera Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package upgrade
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/containernetworking/plugins/plugins/ipam/host-local/backend/disk"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	bapi "github.com/projectcalico/libcalico-go/lib/backend/api"
+	"github.com/projectcalico/libcalico-go/lib/backend/k8s"
+	client "github.com/projectcalico/libcalico-go/lib/clientv3"
+	"github.com/projectcalico/libcalico-go/lib/errors"
+	"github.com/projectcalico/libcalico-go/lib/ipam"
+	"github.com/projectcalico/libcalico-go/lib/net"
+	cnet "github.com/projectcalico/libcalico-go/lib/net"
+	"github.com/projectcalico/libcalico-go/lib/options"
+
+	"github.com/projectcalico/cni-plugin/internal/pkg/utils"
+)
+
+const (
+	ipAllocPath = "/var/lib/cni/networks/k8s-pod-network"
+)
+
+var (
+	binariesToDisable = []string{
+		"/host/opt/cni/bin/calico",
+		"/host/opt/cni/bin/host-local",
+	}
+)
+
+type accessor interface {
+	Backend() bapi.Client
+}
+
+func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
+	// k8sClient directly calls the k8s apiserver.
+	k8sClient := c.(accessor).Backend().(*k8s.KubeClient).ClientSet
+
+	// Check to see if the system is still using host-local
+	// by checking the existence of the path.
+	log.Info("checking host-local IPAM data dir dir existence...")
+	if _, err := os.Stat(ipAllocPath); err != nil && os.IsNotExist(err) {
+		log.Info("host-local IPAM data dir dir not found; no migration necessary, successfully exiting...")
+		return nil
+	}
+
+	// Get node resource to check for IPIP tunnel address.
+	log.Info("retrieving node for IPIP tunnel address")
+	node, err := c.Nodes().Get(ctxt, nodename, options.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get calico node resource: %s", err)
+	}
+
+	// Migrate IPIP tunnel address if the field is empty.
+	if node.Spec.BGP.IPv4IPIPTunnelAddr == "" {
+		log.Info("IPIP tunnel address not found, assigning...")
+
+		// Fetch k8s node.
+		k8sNode, err := k8sClient.CoreV1().Nodes().Get(nodename, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get k8s node resource: %s", err)
+		}
+
+		// Parse PodCIDR.
+		ip, cidr, err := net.ParseCIDR(k8sNode.Spec.PodCIDR)
+		if err != nil {
+			return fmt.Errorf("PodCIDR %s did not parse successfully: %s", k8sNode.Spec.PodCIDR, err)
+		} else if cidr.Version() == 4 {
+			// We need to get the IP for the podCIDR and increment it to the
+			// first IP in the CIDR.
+			tunIp := ip.To4()
+			if tunIp == nil {
+				return fmt.Errorf("Cannot pick an IPv4 tunnel address from the given CIDR: %s", k8sNode.Spec.PodCIDR)
+			}
+			tunIp[3]++
+			node.Spec.BGP.IPv4IPIPTunnelAddr = tunIp.String()
+
+			// Assign the address via Calico IPAM.
+			ipipTunnelAddr := cnet.ParseIP(node.Spec.BGP.IPv4IPIPTunnelAddr)
+			if err = c.IPAM().AssignIP(ctxt, ipam.AssignIPArgs{
+				IP:       *ipipTunnelAddr,
+				Hostname: nodename,
+				Attrs: map[string]string{
+					"node":           nodename,
+					"ipipTunnelAddr": "true",
+				},
+			}); err != nil {
+				if _, ok := err.(errors.ErrorResourceAlreadyExists); !ok {
+					return fmt.Errorf("failed to get add IPIP tunnel addr %s: %s", node.Spec.BGP.IPv4IPIPTunnelAddr, err)
+				}
+				log.Info("IPIP tunnel address already assigned, continuing...")
+			}
+
+			// Save the calico node object to the datastore.
+			if node, err = c.Nodes().Update(ctxt, node, options.SetOptions{}); err != nil {
+				return fmt.Errorf("failed to save newly updated node object: %s", err)
+			}
+
+			log.Info("Successfully assigned and updated IPIP tunnel address")
+		}
+	}
+	log.WithField("ip", node.Spec.BGP.IPv4IPIPTunnelAddr).Info("node has IPIP tunnel address")
+
+	// Open k8s-pod-directory to check for emptiness.
+	log.Info("checking if host-local IPAM data dir dir is empty...")
+	ipamDir, err := os.Open(ipAllocPath)
+	if err != nil {
+		return fmt.Errorf("failed to open host-local IPAM data dir dir: %s", err)
+	}
+
+	// Check if the directory is empty.
+	if _, err = ipamDir.Readdirnames(1); err != nil {
+		if err == io.EOF {
+			log.Info("host-local IPAM data dir empty; no migration necessary...")
+			log.Info("removing host-local IPAM data directory...")
+			if err = os.Remove(ipAllocPath); err != nil {
+				log.WithError(err).Error("failed to remove host-local IPAM data dir directory")
+			}
+			log.Info("successfully removed host-local IPAM data directory!")
+			return nil
+		}
+		if closeErr := ipamDir.Close(); closeErr != nil {
+			return fmt.Errorf("failed to close host-local IPAM data dir directory on read failure: %s", err)
+		}
+		return fmt.Errorf("failed to read host-local IPAM data dir names: %s", err)
+	}
+	log.Info("host-local IPAM data dir is not empty, migrating...")
+
+	// Close the host-local IPAM data directory file pointer.
+	if closeErr := ipamDir.Close(); closeErr != nil {
+		return fmt.Errorf("failed to close host-local IPAM data dir directory: %s", err)
+	}
+
+	// Disable cni by setting datastoreReady to false.
+	log.Info("setting datastore readiness to false")
+	clusterInfo, err := c.ClusterInformation().Get(ctxt, "default", options.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to fetch cluster information: %s", err)
+	}
+	if *clusterInfo.Spec.DatastoreReady {
+		*clusterInfo.Spec.DatastoreReady = false
+		for i := uint(0); i < 5; i++ {
+			if clusterInfo, err = c.ClusterInformation().Update(ctxt, clusterInfo, options.SetOptions{}); err != nil {
+				if _, ok := err.(errors.ErrorResourceUpdateConflict); ok {
+					log.Info("Encountered update conflict, retrying...")
+					time.Sleep((1 << i) * time.Second)
+				}
+				return fmt.Errorf("failed to disable cluster: %s", err)
+			}
+			break
+		}
+	}
+	log.Info("successfully set datastore readiness to false!")
+
+	// Also disable cni by deleting the binaries.
+	log.Info("removing cni binaries...")
+	for _, binary := range binariesToDisable {
+		if err = os.Remove(binary); err != nil {
+			return fmt.Errorf("failed to remove binary %s: %s", binary, err)
+		}
+		log.WithField("binary", binary).Info("successfully removed cni binary!")
+	}
+	time.Sleep(5 * time.Second)
+
+	// Establishing a mapping of IP addresses to Pods on this node.
+	log.Info("mapping pod ips to pods...")
+	podList, err := k8sClient.CoreV1().Pods("").List(metav1.ListOptions{
+		FieldSelector: fmt.Sprintf("spec.nodeName=%s", nodename),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list pods: %s", err)
+	}
+	podIPMap := make(map[string]*corev1.Pod)
+	for i := 0; i < len(podList.Items); i++ {
+		pod := &podList.Items[i]
+		log.WithFields(log.Fields{"pod": pod.Name, "IP": pod.Status.PodIP, "namespace": pod.Namespace}).Info("mapping in pod")
+		podIPMap[pod.Status.PodIP] = pod
+	}
+	log.Info("successfully mapped pod ips to pods!")
+
+	// Acquire a lock on the host-local cni backend
+	log.Info("acquiring lock on host-local backend...")
+	hostLocal, err := disk.New(ipAllocPath, "")
+	if err != nil {
+		return fmt.Errorf("failed to initialize host local backend: %s", err)
+	}
+	if err = hostLocal.Lock(); err != nil {
+		return fmt.Errorf("failed to acquire lock on host local backend: %s", err)
+	}
+	log.Info("successfully acquired lock on host-local backend!")
+	defer func() {
+		// Release the lock on host local backend
+		log.Info("releasing lock on host-local backend...")
+		if err = hostLocal.Unlock(); err != nil {
+			log.WithError(err).Error("failed to release lock on host local backend")
+		}
+		log.Info("successfully released lock on host-local backend!")
+	}()
+
+	// Read in all the files in the host-local directory.
+	log.Info("reading files from host-local IPAM data dir...")
+	files, err := ioutil.ReadDir(ipAllocPath)
+	if err != nil {
+		return fmt.Errorf("failed to read path %s: %s", ipAllocPath, err)
+	}
+
+	// For each file, convert it into an IP allocation and then delete the file.
+	for _, f := range files {
+		logCtxt := log.WithField("file", f.Name())
+		logCtxt.Info("processing file...")
+		fname := path.Join(ipAllocPath, f.Name())
+
+		// Delete and skip the last reserved IP.
+		if strings.Contains(f.Name(), "last") {
+			logCtxt.Info("skipping and removing last reserved ip file...")
+			if err = os.Remove(fname); err != nil {
+				return fmt.Errorf("failed to remove file %s: %s", fname, err)
+			}
+			logCtxt.Info("successfully removed last reserved ip file!")
+			continue
+		}
+
+		// Skip the lock
+		if f.Name() == "lock" {
+			logCtxt.Info("skipping the lock...")
+			continue
+		}
+
+		// The name of the file is its IP address.
+		ip, _, err := cnet.ParseCIDR(fmt.Sprintf("%s/32", f.Name()))
+		if err != nil {
+			return fmt.Errorf("failed to parse IP %s: %s", f.Name(), err)
+		}
+
+		// The contents are the container ID.
+		b, err := ioutil.ReadFile(fname)
+		if err != nil {
+			return fmt.Errorf("failed to read file %s: %s", fname, err)
+		}
+		containerID := string(b)
+
+		// See if the IP hasn't already been assigned for the handleID.
+		handleID := utils.GetHandleID("k8s-pod-network", containerID, "")
+		ips, err := c.IPAM().IPsByHandle(ctxt, handleID)
+		if err == nil {
+		} else if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			// Return error if it is NOT does not exist, otherwise continue.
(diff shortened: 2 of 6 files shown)


cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	// Delete the directory
+	if err := os.Remove(ipAllocPath); err != nil {
+		log.WithField("path", ipAllocPath).WithError(err).Fatal("failed to remove directory")
+	}
+
+	// Check DaemonSet rolling update status to see if this is the last node to update.
+	// If so, re-enable auto-allocation of blocks
+	k8sCfg, err := clientcmd.BuildConfigFromFlags("", "")
```
caseydavenport: Might be worth building these up-front, since they are prone to failing as a result of misconfiguration.

Basically, if the code is misconfigured we probably want it to fail before doing anything.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,193 @@
...
+			StrictAffinity:     false,
+		}); err != nil {
+			log.WithError(err).Fatal("failed to re-enable auto-allocation of block")
+		}
+	}
+}
+
+func assignCalicoIP(c client.Interface, ip cnet.IP, containerID, nodename string) error {
+	handleID := fmt.Sprintf("%s.%s", "k8s-pod-network", containerID)
+	return c.IPAM().AssignIP(context.Background(), ipam.AssignIPArgs{
```
caseydavenport: Ah, this is an interesting one... this PR adds some attributes: https://github.com/projectcalico/cni-plugin/pull/680

Which, the cleanup controller relies on. We should figure out what to do in this code. Options include:
- Query the apiserver to figure out the pod this IP belongs to
- Add some other attribute the allocation cleanup controller can use to clean this up.
- ???


We should at least include the node in the attributes though (as the linked PR does).

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,193 @@
...
+
+		// Store allocation to datastore
+		err = assignCalicoIP(calicoClient, *ip, string(b), nodename)
+		if err != nil {
+			log.WithField("nodename", nodename).WithError(err).Fatal("failed to assign IP to calico backend")
+		}
+
+		// Delete the file
+		if err := os.Remove(fname); err != nil {
+			log.WithField("file", fname).WithError(err).Fatal("failed to remove file")
```
caseydavenport: I think rather than exit the process on each of these error conditions, we should implement some retry logic.

We could wrap this in another function which orchestrates the retries, e.g.:

```go
  for i:=0; i< numretries; i++ {
    if err := doUpgrade(); err != nil { continue}
    break
  }
```

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,193 @@
...
+		log.WithError(err).Fatal("failed to load api client config")
+	}
+	cfg.Spec.DatastoreType = apiconfig.Kubernetes
+	calicoClient, err := client.New(*cfg)
+	if err != nil {
+		log.WithError(err).Fatal("failed to initialize api client")
+	}
+
+	// Disable auto-allocation of blocks
+	if err = calicoClient.IPAM().SetIPAMConfig(context.Background(), ipam.IPAMConfig{
```
caseydavenport: Let's swap this over to using the ready flag for now - I think given these lines of code this won't work the way I originally imagined: https://github.com/projectcalico/libcalico-go/blob/9ccf86fcbbe260dd6ba0c31afaa9686e3025868e/lib/ipam/ipam.go#L1357-L1360

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// First, disable the necessary binaries so it can't be used.
+	for _, binary := range binariesToDisable {
+		if err = toggleFile(binary, true); err != nil {
+			log.WithField("file", binary).WithError(err).Fatal("failed to disable binary")
+		}
+	}
+	defer func() {
+		if err != nil {
+			rollback()
```
caseydavenport: I like the idea, but as you mentioned earlier this might leave us in a worse state than simply failing the upgrade altogether. Might be better to simply log out a clear error and stop the upgrade, so users can correct manually.

caseydavenport: Also, we definitely don't want to do this on success - once this script is done executing, we shouldn't allow any more host-local allocations to take place, otherwise that data will be lost.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+const (
+	ipAllocPath = "/var/lib/cni/networks/k8s-pod-network"
+)
+
+func main() {
+	// nodename associates IPs to this node.
+	nodename := os.Getenv("KUBERNETES_NODE_NAME")
+	logCtxt := log.WithField("node", nodename)
+
+	if msg, err := Main(nodename); err != nil {
```
caseydavenport: Hm, this pattern of passing a message + error back isn't something I've seen before.

Usually I'd expect the called code to wrap any errors, e.g. `return fmt.Errorf("failed to init client: %s", err)` and then the function simply returns an error.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+		msg = "failed to fetch cluster information"
+		return
+	}
+	if *clusterInfo.Spec.DatastoreReady {
+		*clusterInfo.Spec.DatastoreReady = false
+		if clusterInfo, err = calicoClient.ClusterInformation().Update(ctxt, clusterInfo, options.SetOptions{}); err != nil {
+			msg = "failed to disable cluster"
+			return
+		}
+	}
```
caseydavenport: I think we need a reasonable sleep here to allow any in-progress CNI executions to finish - maybe 5 seconds?

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Establishing a mapping of IP addresses to Pods on this node
+	if podList, err = k8sClient.CoreV1().Pods("").List(v1.ListOptions{
+		LabelSelector: fmt.Sprintf("spec.nodeName = %s", nodename),
```
caseydavenport: I don't think the label selector will work here - I think you need to use the field selector.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+	}); err != nil {
+		msg = "failed to list pods"
+		return
+	}
+	podIPMap := make(map[string]*corev1.Pod)
+	for i := 0; i < len(podList.Items); i++ {
```
caseydavenport: why not do a `for pod := range podList.Items`?

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+		podIPMap[podList.Items[i].Status.PodIP] = &podList.Items[i]
```
caseydavenport: We should handle the case where the pod IP is not present.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	// Migrate IPIP tunnel address.
+	if node, err = calicoClient.Nodes().Get(ctxt, nodename, options.GetOptions{}); err != nil {
```
caseydavenport: Let's discuss the strategy here - not sure if this will work.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+			log.WithField("ip", f.Name()).Info("pod not found for IP, deleting and continuing...")
+			if err = os.Remove(fname); err != nil {
+				msg = "failed to remove file " + fname
+				return
+			}
+			continue
+		}
+
+		// Store allocation to datastore
+		handleID := fmt.Sprintf("%s.%s", "k8s-pod-network", string(b))
```
caseydavenport: Probably should use this utility function: https://github.com/projectcalico/cni-plugin/blob/master/internal/pkg/utils/utils.go#L457

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+			IP:       *ip,
+			HandleID: &handleID,
+			Hostname: nodename,
+			Attrs: map[string]string{
+				"node":      nodename,
+				"pod":       pod.Name,
+				"namespace": pod.Namespace,
+			},
+		}); err != nil {
+			msg = "failed to assign IP to calico backend"
```
caseydavenport: What happens if we succeed here, but then fail to delete the file?

I think we'll retry, but when we get to this point a second time it will error since the IP is already allocated. We should probably special case the "Already exists" error case.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+
+		// Delete the file
+		if err = os.Remove(fname); err != nil {
+			msg = "failed to remove file " + fname
+			return
+		}
+	}
+
+	// Delete the directory
+	if err = os.Remove(ipAllocPath); err != nil {
```
caseydavenport: Can we formulate a way to ensure this directory is empty before deleting? Maybe just a double check that there are no more files in the directory?

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Check DaemonSet rolling update status to see if this is the last node to update.
+	// If so, re-enable auto-allocation of blocks
+	if calicoDS, err = k8sClient.ExtensionsV1beta1().DaemonSets("kube-system").Get("calico-node", v1.GetOptions{}); err != nil {
+		msg = "failed to get daemonsets/kube-system.calico-node"
+		return
+	}
+
+	// UpdatedNumberScheduled tracks the total number of nodes that are running updated daemon pod.
+	if calicoDS.Status.UpdatedNumberScheduled == calicoDS.Status.DesiredNumberScheduled-1 {
```
caseydavenport: Wonder if we should look at `maxUnavailable` here.

e.g., if someone decides to upgrade 5 pods at a time, then it's possible none of them will detect they are the last one. 

Probably better if all of them think they are the last one.

cmd/calico-ipam-upgrade/main.go
briansan: Yeah, I was trying something out and it tickled me in a similar way. I think the way you laid out is more standard. I'll make the switch

cmd/calico-ipam-upgrade/main.go
briansan: Oopsyy~

cmd/calico-ipam-upgrade/main.go
briansan: I'm pretty sure that Go reuses the variable when enumerating via range and for the sake of minimizing memory footprint and guarding from opaque language internals, I think it's better to explicitly store the reference to the individual pod elements using the standard for loop iteration.

cmd/calico-ipam-upgrade/main.go
briansan: Well this section of the code is just building up the map of IPs to Pods. Handling non-existing pod IPs is done on L171.

cmd/calico-ipam-upgrade/main.go
briansan: That is a really odd looking function. It takes in an arg that never gets used and returns an error that never occurs...

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+type Accessor interface {
+	Backend() bapi.Client
+}
+
+func main() {
+	log.Info("migrating from host-local to calico-ipam...")
+	ctxt := context.Background()
+
+	// nodename associates IPs to this node.
+	nodename := os.Getenv("KUBERNETES_NODE_NAME")
```
caseydavenport: Might be worth a check that this isn't an empty string - if it is we want to error out.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	// Execute migration
+	if err := Migrate(ctxt, calicoClient, nodename); err != nil {
+		logCtxt.WithError(err).Fatal("failed to migrate ipam")
+	}
+}
+
+func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
+	// Set up logging formatting.
```
caseydavenport: These should be at the beginning of execution, so that all logs get this formatting.

As it is, we're emitting some logs before these lines get executed.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+	log.SetFormatter(&logutils.Formatter{})
+
+	// Install a hook that adds file/line no information.
+	log.AddHook(&logutils.ContextHook{})
+
+	// k8sClient directly calls the k8s apiserver.
+	k8sClient := c.(Accessor).Backend().(*k8s.KubeClient).ClientSet
+
+	var err error
+	defer func() {
```
caseydavenport: What's this for?

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+				return fmt.Errorf("failed to save newly updated node object: %s", err)
+			}
+
+			log.Info("Successfully assigned and updated IPIP tunnel address")
+		}
+	}
+
+	// Check to see if the system is still using host-local
+	// by checking the existence of the path.
+	if _, err := os.Stat(ipAllocPath); err != nil && os.IsNotExist(err) {
```
caseydavenport: Shouldn't we do this first? We shouldn't try to migrate the IPIP tunnel address at all if there's no host-local directory IIUC.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	// Check if the directory is empty.
+	ipamDir, err := os.Open(ipAllocPath)
+	if err != nil {
+		return fmt.Errorf("failed to open k8s-pod-network dir: %s", err)
+	}
+	if _, err = ipamDir.Readdirnames(1); err != nil {
+		if err == io.EOF {
+			log.Info("k8s-pod-network dir empty; no migration necessary, successfully exiting...")
```
caseydavenport: Should we delete the directory in this case, so we don't hit this branch ever again?

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+			return nil
+		}
+		return fmt.Errorf("failed to read k8s-pod-network dir names: %s", err)
+	}
+	defer func() {
```
caseydavenport: Should we execute this defer up below where we open the `ipamDir`?

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+	}()
+
+	// Disable cni by setting datastoreReady to false.
+	clusterInfo, err := c.ClusterInformation().Get(ctxt, "default", options.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to fetch cluster information: %s", err)
+	}
+	if *clusterInfo.Spec.DatastoreReady {
+		*clusterInfo.Spec.DatastoreReady = false
+		if clusterInfo, err = c.ClusterInformation().Update(ctxt, clusterInfo, options.SetOptions{}); err != nil {
```
caseydavenport: I think we should wrap this logic in a retry loop - it's likely we'll hit `ErrorResourceUpdateConflict` here. In that case, we can simply retry setting to `false` rather than exit the entire container and start from the beginning.

cmd/calico-ipam-upgrade/main.go
caseydavenport: @briansan I don't see a sleep... probably should be below the part where we delete the binaries.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+			return fmt.Errorf("failed to disable cluster: %s", err)
+		}
+	}
+
+	// Also disable cni by disabling the binaries.
+	for _, binary := range binariesToDisable {
+		dir, file := path.Split(binary)
+		disabledBinary := path.Join(dir, file+".disabled")
+		if err = os.Rename(binary, disabledBinary); err != nil {
```
caseydavenport: Won't this fail if the binary isn't present? For example, if we've already run the script to this point and then it gets restarted? 

We probably want to be resilient to the binaries themselves not existing.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+
+	podList, err := k8sClient.CoreV1().Pods("").List(metav1.ListOptions{
+		FieldSelector: fmt.Sprintf("spec.nodeName=%s", nodename),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list pods: %s", err)
+	}
+	podIPMap := make(map[string]*corev1.Pod)
+	for i := 0; i < len(podList.Items); i++ {
+		podIPMap[podList.Items[i].Status.PodIP] = &podList.Items[i]
```
caseydavenport: I think it will be useful for debugging to log out the mappings here for each iteration of the loop (including those with no IP address).

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read path %s: %s", ipAllocPath, err)
+	}
+
+	// For each file, convert it into an IP allocation and then delete the file.
+	for _, f := range files {
+		fname := path.Join(ipAllocPath, f.Name())
+
+		// Delete and skip the last reserved IP.
+		if strings.Contains(f.Name(), "last") || f.Name() == "lock" {
```
caseydavenport: I think we should look at how the host-local plugin uses the lock file. We may want to respect it - e.g., wait for the lock to be available and then grab it ourselves before reading in all these files.

caseydavenport: See here: https://github.com/containernetworking/plugins/blob/master/plugins/ipam/host-local/backend/store.go#L19

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+		b, err := ioutil.ReadFile(fname)
+		if err != nil {
+			return fmt.Errorf("failed to read file %s: %s", fname, err)
+		}
+		containerID := string(b)
+
+		// See if the IP hasn't already been assigned for the handleID.
+		handleID := utils.GetHandleID("k8s-pod-network", containerID)
+		/*
+			ips, err := c.IPAM().IPsByHandle(ctxt, handleID)
```
caseydavenport: What's this for? Can we delete it?

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+			continue
+		}
+
+		// Store allocation to datastore
+		if err = c.IPAM().AssignIP(ctxt, ipam.AssignIPArgs{
+			IP:       *ip,
+			HandleID: &handleID,
+			Hostname: nodename,
+			Attrs: map[string]string{
+				"node":      nodename,
```
caseydavenport: Just added some constants for these which would be good to use: https://github.com/projectcalico/libcalico-go/pull/1035/files

internal/pkg/utils/utils.go
```diff
@@ -454,15 +454,10 @@ func GetIdentifiers(args *skel.CmdArgs, nodename string) (*WEPIdentifiers, error
 	return &epIDs, nil
 }
 
-func GetHandleID(netName string, containerID string, workload string) (string, error) {
+func GetHandleID(netName, containerID string) string {
 	handleID := fmt.Sprintf("%s.%s", netName, containerID)
-	logrus.WithFields(logrus.Fields{
```
caseydavenport: I think we do actually want to keep this log as it is - it's really useful to have all this information in the CNI log output.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+)
+
+var (
+	binariesToDisable = []string{
+		"/host/opt/cni/bin/calico",
+		"/host/opt/cni/bin/host-local",
+	}
+)
+
+type Accessor interface {
```
caseydavenport: Does this need to be a public interface?

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,298 @@
...
+	Backend() bapi.Client
+}
+
+func main() {
```
caseydavenport: What's the plan for getting this into the Docker image?

I think we have a couple of options:
- Simply add the new binary to the docker image via the Dockerfile.
- Instead of building a new binary, add this logic behind a flag on `calico-ipam`, e.g., `calico-ipam -upgrade`

I'm happy with either - the latter has the nice benefit of keeping the image size a bit smaller.

cmd/calico-ipam-upgrade/main.go
briansan: This is the best approach I could think to give the user time to view the error in the logs if one occurs without it getting overwritten by another error in a future execution.

cmd/calico-ipam-upgrade/main.go
briansan: Forgot to finish implementing the logic to check if the IP hasn't already been assigned for the given handle

internal/pkg/utils/utils.go
briansan: No problem reverting the change, but isn't it sufficient to just log out the handleID since it's constructed using the netName and containerID?

cmd/calico-ipam-upgrade/main.go
briansan: Hmmm...well, the current implementation uses the first option you describe but it seems to have bumped up the binary size from 27MB to 37MB which is a 37% increase in size. I think it's worth to make the change. WDYT?

caseydavenport: SGTM.

cmd/calico-ipam-upgrade/main.go
caseydavenport: Rather than the defer, we should at least put it in the error handling when we call `Migrate` above. I think that's easier to reason about.

But, I'm not sure a sleep is really the best thing here - why don't we just implement the retry logic ourselves? What is the value in having the init container exit?

e.g.

```go
for {
  err := Migrate()
  if err != nil {
    // log error
    time.Sleep(1 * time.Second)
    continue
  }
}
```

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,333 @@
...
+	// TODO: Should check DaemonSet rolling update status to see if this is the last node to update and
+	// only re-enable datastoreReady if the number of UpdatedNumberScheduled is MaxUnavailable less than the DesiredNumberScheduled.
+	// calicoDS, err := k8sClient.AppsV1().DaemonSets("kube-system").Get("calico-node", metav1.GetOptions{})
+	// if (calicoDS.Status.UpdatedNumberScheduled == calicoDS.Status.DesiredNumberScheduled-calicoDS.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.IntVal)
+	*clusterInfo.Spec.DatastoreReady = true
+	if _, err = c.ClusterInformation().Update(ctxt, clusterInfo, options.SetOptions{}); err != nil {
+		return fmt.Errorf("failed to re-enable cluster: %s", err)
+	}
+
+	return nil
```
caseydavenport: I think if we exit successfully we should remove the `.disabled` binaries so they don't sit around forever.

cmd/calico-ipam-upgrade/main.go
```diff
@@ -0,0 +1,333 @@
...
+		disabledBinary := path.Join(dir, file+".disabled")
+		if err = os.Rename(binary, disabledBinary); err != nil {
+			if os.IsNotExist(err) {
+				log.WithField("binary", file).Info("binary not found, probably already disabled, continuing...")
+				continue
+			}
+			return fmt.Errorf("failed to disable binary %s: %s", binary, err)
+		}
+	}
+	time.Sleep(3 * time.Second)
```
caseydavenport: I think we should up this to 5 seconds to be safe. It's somewhat arbitrary ("magic number") but 3 feels like it might not be quite enough.

internal/pkg/utils/utils.go
caseydavenport: I think the main thing is we want a log which includes the WorkloadID as well so that we can tie the two together.

cmd/calico-ipam-upgrade/main.go
briansan: agreed

pkg/upgrade/migrate.go
```diff
@@ -0,0 +1,359 @@
...
+	log.Info("checking if host-local IPAM data dir dir is empty...")
+	ipamDir, err := os.Open(ipAllocPath)
+	if err != nil {
+		return fmt.Errorf("failed to open host-local IPAM data dir dir: %s", err)
+	}
+
+	// Check if the directory is empty.
+	if _, err = ipamDir.Readdirnames(1); err != nil {
+		if err == io.EOF {
+			log.Info("host-local IPAM data dir empty; no migration necessary, removing directory and exiting...")
```
caseydavenport: Doesn't look like we're actually deleting the dir here, unlike what the log suggests.

pkg/upgrade/migrate.go
```diff
@@ -0,0 +1,341 @@
...
+		handleID := utils.GetHandleID("k8s-pod-network", containerID, "")
+		ips, err := c.IPAM().IPsByHandle(ctxt, handleID)
+		if err == nil {
+		} else if _, ok := err.(errors.ErrorResourceDoesNotExist); !ok {
+			// Return error if it is NOT does not exist, otherwise continue.
+			return fmt.Errorf("failed to fetch ips for handleID %s: %s", handleID, err)
+		}
+		logCtxt.WithField("handleID", handleID).Info("file contains handle ID")
+
+		alreadyExists := false
```
caseydavenport: IIUC, I think a more robust way to do this check would be to try to assign the IP address and if you get an AlreadyExists error, you continue.

It means you could remove a bunch of code by getting rid of the `IPsByHandle` call above altogether, as well as this cross-correlation code to check if it exists. TBH I don't really understand what this loop is meant to check, so I could definitely be wrong.

pkg/upgrade/migrate.go
```diff
@@ -0,0 +1,341 @@
...
+		logCtxt.Info("removing file...")
+		if err = os.Remove(fname); err != nil {
+			return fmt.Errorf("failed to remove file %s: %s", fname, err)
+		}
+		logCtxt.Info("successfully removed file!")
+	}
+
+	// Delete the host-local IPAM data directory
+	log.Info("removing host-local IPAM data directory...")
+	if err = os.Remove(ipAllocPath); err != nil {
```
caseydavenport: Will this succeed without us deleting the lock file? I think we need to release the lock and delete the file right before this.

pkg/upgrade/migrate.go
```diff
@@ -0,0 +1,341 @@
...
+		return fmt.Errorf("failed to acquire lock on host local backend: %s", err)
+	}
+	log.Info("successfully acquired lock on host-local backend!")
+	defer func() {
+		// Release the lock on host local backend
+		log.Info("releasing lock on host-local backend...")
+		if err = hostLocal.Unlock(); err != nil {
+			log.WithError(err).Error("failed to release lock on host local backend")
+		}
+		log.Info("successfully released lock on host-local backend!")
```
caseydavenport: This log is going to be emitted even if we fail to release the lock.

PR #684, opened 2019-02-21
Update makefile to test kdd as well

lwr20:
## Description
Make tests run in KDD mode with calico-ipam

## Todos
- [ ] Make it work

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,8 +126,6 @@ endif
 
 LIBCALICOGO_PATH?=none
 
-DATASTORE_TYPE?=etcdv3
-
 LOCAL_USER_ID?=$(shell id -u $$USER)
 
 .PHONY: clean
@@ -309,6 +307,10 @@ foss-checks: vendor
 ###############################################################################
 ## Run the unit tests.
 ut: run-k8s-controller build $(BIN)/host-local
+	$(MAKE) ut-datastore DATASTORE_TYPE=etcdv3
+	$(MAKE) ut-datastore DATASTORE_TYPE=kubernetes
+
+ut-datastore:
 	# The tests need to run as root
 	docker run --rm -t --privileged --net=host \
 	-e ETCD_IP=$(LOCAL_IP_ENV) \
@@ -319,12 +321,22 @@ ut: run-k8s-controller build $(BIN)/host-local
 	-e CNI_SPEC_VERSION=$(CNI_SPEC_VERSION) \
 	-e DATASTORE_TYPE=$(DATASTORE_TYPE) \
 	-e ETCD_ENDPOINTS=http://$(LOCAL_IP_ENV):2379 \
+	-e K8S_API_ENDPOINT=http://127.0.0.1:8080 \
 	-v $(CURDIR):/go/src/$(PACKAGE_NAME):rw \
 	$(LOCAL_BUILD_MOUNTS) \
 	$(CALICO_BUILD) sh -c '\
 			cd  /go/src/$(PACKAGE_NAME) && \
 			ginkgo -cover -r -skipPackage vendor -skipPackage k8s-install $(GINKGO_ARGS)'
+
+ut-etcd: run-k8s-controller build $(BIN)/host-local
+	$(MAKE) ut-datastore DATASTORE_TYPE=etcdv3
 	make stop-etcd
+	make stop-k8s-controller
+
+ut-kdd: run-k8s-controller build $(BIN)/host-local
+	$(MAKE) ut-datastore DATASTORE_TYPE=kubernetes
+	make stop-etcd
+	make stop-k8s-controller
 
 ## Run the tests in a container (as root) for different CNI spec versions
 ## to make sure we don't break backwards compatibility.
@@ -338,12 +350,16 @@ test-cni-versions:
 run-k8s-apiserver: stop-k8s-apiserver run-etcd
 	docker run --detach --net=host \
 	  --name calico-k8s-apiserver \
+	  -v `pwd`/vendor/github.com/projectcalico/libcalico-go/test/crds.yaml:/crds.yaml \
 	  -v `pwd`/internal/pkg/testutils/private.key:/private.key \
 	  gcr.io/google_containers/hyperkube-$(ARCH):$(K8S_VERSION) \
 	  /hyperkube apiserver \
 	    --etcd-servers=http://$(LOCAL_IP_ENV):2379 \
 	    --service-cluster-ip-range=10.101.0.0/16 \
 	    --service-account-key-file=/private.key
+	# Wait until the apiserver is accepting requests.
+	while ! docker exec calico-k8s-apiserver kubectl get nodes; do echo "Waiting for apiserver to come up..."; sleep 2; done
+	docker exec calico-k8s-apiserver kubectl apply -f /crds.yaml
 
 ## Kubernetes controller manager used for tests
 run-k8s-controller: stop-k8s-controller run-k8s-apiserver
@@ -537,7 +553,7 @@ run-kube-proxy:
 ## Run the unit tests, watching for changes.
 test-watch: $(BIN)/calico $(BIN)/calico-ipam run-etcd run-k8s-apiserver
 	# The tests need to run as root
-	sudo CGO_ENABLED=0 ETCD_IP=127.0.0.1 PLUGIN=calico GOPATH=$(GOPATH) $(shell which ginkgo) watch -skipPackage k8s-install -skipPackage vendor
+	CGO_ENABLED=0 ETCD_IP=127.0.0.1 PLUGIN=calico GOPATH=$(GOPATH) $(shell which ginkgo) watch -skipPackage k8s-install -skipPackage vendor
 
 .PHONY: help
 help: # Some kind of magic from https://gist.github.com/rcmachado/af3db315e31383502660
--- a/internal/pkg/utils/utils.go
+++ b/internal/pkg/utils/utils.go
@@ -1,4 +1,4 @@
-// Copyright 2015-2018 Tigera Inc
+// Copyright (c) 2015-2019 Tigera, Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/tests/calico_cni_k8s_test.go
+++ b/tests/calico_cni_k8s_test.go
@@ -1,3 +1,5 @@
+// Copyright (c) 2015-2019 Tigera, Inc. All rights reserved.
+
 package main_test
 
 import (
@@ -22,7 +24,6 @@ import (
 	"github.com/projectcalico/cni-plugin/internal/pkg/testutils"
 	"github.com/projectcalico/cni-plugin/internal/pkg/utils"
 	"github.com/projectcalico/cni-plugin/pkg/types"
-	"github.com/projectcalico/libcalico-go/lib/apis/v3"
 	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	k8sconversion "github.com/projectcalico/libcalico-go/lib/backend/k8s/conversion"
 	client "github.com/projectcalico/libcalico-go/lib/clientv3"
@@ -36,6 +37,7 @@ import (
 	"github.com/vishvananda/netlink"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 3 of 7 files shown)


tests/calico_cni_k8s_test.go
```diff
@@ -52,6 +54,25 @@ func ensureNamespace(clientset *kubernetes.Clientset, name string) {
 	Expect(err).NotTo(HaveOccurred())
 }
 
+func ensurePodDeleted(clientset *kubernetes.Clientset, ns string, podName string) {
+	err := clientset.CoreV1().Pods(ns).Delete(podName, metav1.NewDeleteOptions(0))
+	if kerrors.IsNotFound(err) {
+		return
+	}
+	if err != nil {
```
caseydavenport: Probably better to `Expect(err).NotTo(BeNil())`

If we can't do that here, then we'll want to return the error and have the calling code handle it.

tests/calico_cni_k8s_test.go
```diff
@@ -52,6 +54,25 @@ func ensureNamespace(clientset *kubernetes.Clientset, name string) {
...
+		panic(err)
+	}
+	Eventually(func() bool {
+		pod, err := clientset.CoreV1().Pods(ns).Get(podName, metav1.GetOptions{})
+		if kerrors.IsNotFound(err) {
+			return true
+		}
+		Expect(err).NotTo(HaveOccurred())
```
caseydavenport: Hm, I'm not sure if you can do an Expect inside an Eventually, I haven't seen that before.

Usually for this pattern we use a `func() error` and then expect it to be nil, returning an error whenever we haven't met the desired criteria.

tests/calico_cni_k8s_test.go
```diff
@@ -65,25 +86,31 @@ var _ = Describe("Kubernetes CNI tests", func() {
...
 	BeforeEach(func() {
-		testutils.WipeK8sPods()
-		testutils.WipeEtcd()
+		testutils.WipeDatastore()
 
 		// Create the node for these tests. The IPAM code requires a corresponding Calico node to exist.
-		var err error
-		n := v3.NewNode()
-		n.Name, err = names.Hostname()
+		var name string
```
caseydavenport: think we can just do a `name, err := names.Hostname()` here instead of predeclaring.

tests/calico_cni_k8s_test.go
```diff
@@ -248,6 +282,10 @@ var _ = Describe("Kubernetes CNI tests", func() {
 					Type:      syscall.RTN_UNICAST,
 				})))
 
+			// Delete pod
+			ensurePodDeleted(clientset, testutils.K8S_TEST_NS, name)
```
caseydavenport: Why do we need this change? I'm not sure I'm following this.

caseydavenport: I think we need the Pod to exist at this point - `testutils.DeleteContainer` is calling the CNI plugin, which expects the pod to exist in the API.

tests/calico_cni_k8s_test.go
```diff
@@ -719,6 +775,19 @@ var _ = Describe("Kubernetes CNI tests", func() {
 							PodCIDR: "10.0.0.0/24",
 						},
 					})
+					if err != nil {
```
caseydavenport: I'm a bit confused by this as well - why do we need this now but didn't need it before?

tests/calico_cni_k8s_test.go
```diff
@@ -1669,6 +1781,9 @@ var _ = Describe("Kubernetes CNI tests", func() {
 			// Delete the container.
 			_, err = testutils.DeleteContainer(netconfCalicoIPAM, netNS.Path(), name, testutils.K8S_TEST_NS)
 			Expect(err).ShouldNot(HaveOccurred())
+
+			// Delete pod
+			ensurePodDeleted(clientset, testutils.K8S_TEST_NS, name)
```
caseydavenport: Can we do this in an afterEach as well?

caseydavenport: I'm still a bit curious as to why these passed before, but now need this.. feel like I must be missing something!

tests/calico_cni_k8s_test.go
```diff
@@ -2210,6 +2361,9 @@ var _ = Describe("Kubernetes CNI tests", func() {
 			Expect(err).NotTo(HaveOccurred())
 			containerID2 := "random-cid"
 			defer func() {
+				// Delete pod
+				ensurePodDeleted(clientset, testutils.K8S_TEST_NS, name2)
```
caseydavenport: I think if we're going to this we need to make sure it occurs _after_ we call the CNI plugin (e.g., `DeleteContainerWithId` below).

tests/calico_cni_test.go
```diff
@@ -50,7 +60,7 @@ var _ = Describe("CalicoCni", func() {
 
 	cniVersion := os.Getenv("CNI_SPEC_VERSION")
 
-	Context("using host-local IPAM", func() {
+	Context("song using host-local IPAM", func() {
```
caseydavenport: stray "song"

Makefile
```diff
@@ -308,7 +306,35 @@ foss-checks: vendor
...
+	-e BIN=/go/src/$(PACKAGE_NAME)/$(BIN) \
+	-e CNI_SPEC_VERSION=$(CNI_SPEC_VERSION) \
+	-e DATASTORE_TYPE=etcdv3 \
+	-e ETCD_ENDPOINTS=http://$(LOCAL_IP_ENV):2379 \
+	-v $(CURDIR):/go/src/$(PACKAGE_NAME):rw \
+	$(LOCAL_BUILD_MOUNTS) \
+	$(CALICO_BUILD) sh -c '\
+			cd  /go/src/$(PACKAGE_NAME) && \
+			ginkgo -cover -r -skipPackage vendor -skipPackage k8s-install $(GINKGO_ARGS)'
+	sudo mv ./report/azure_suite.xml ./report/azure_suite_etcd.xml
```
caseydavenport: I don't think these need `sudo` - we should remove any instances of `sudo` from the Makefile.

Makefile
```diff
@@ -308,7 +306,35 @@ foss-checks: vendor
...
+	sudo mv ./report/cni_suite.xml ./report/cni_suite_etcd.xml
+	make stop-etcd
+	make stop-k8s-controller
+
+ut-kdd: run-k8s-controller build $(BIN)/host-local
+    # Need to sleep here to allow k8s apiserver to start
+	sleep 10
```
caseydavenport: Better than a sleep would be to do something like this: https://github.com/projectcalico/libcalico-go/blob/master/Makefile#L220-L221

I think we could move that inside the `run-k8s-controller` target, so anyone depending on that target can guarantee the API is up by the time it gets called.

internal/pkg/utils/utils.go
```diff
@@ -614,3 +619,50 @@ func ResolvePools(ctx context.Context, c client.Interface, pools []string, isv4
 	}
 	return result, nil
 }
+
+func AddNode(c client.Interface, kc *kubernetes.Clientset, host string) error {
```
caseydavenport: These should go in the `internal/pkg/testutils` package since they are just meant for test code.

Makefile
```diff
@@ -308,7 +306,35 @@ foss-checks: vendor
 # Unit Tests
 ###############################################################################
 ## Run the unit tests.
-ut: run-k8s-controller build $(BIN)/host-local
+ut:
+	make ut-etcd
```
caseydavenport: I think we can just depend on these.

e.g. `ut: ut-etcd ut-kdd`

Makefile
```diff
@@ -308,7 +306,35 @@ foss-checks: vendor
...
+	make ut-kdd
+
+ut-etcd: run-k8s-controller build $(BIN)/host-local
```
caseydavenport: ut-kdd and ut-etcd both have the same dependencies.

It would be nice if we didn't need to restart the k8s components twice when running `make ut`.

Can we format this in a way to avoid that?

maybe something like this:

- `ut`: starts API, etc, and then runs UTs for all datastores.
- `ut-datastore`: runs UTs for a particular datastore, requires DATASTORE_TYPE argument, assumes everything is set up.
- `ut-kdd` starts API, etc, and then runs UTs for KDD by calling `ut-datastore`
- `ut-etcd` starts API, etc, and then runs UTs for etcd by calling `ut-datastore`

Makefile
caseydavenport: Resolved in this commit: https://github.com/projectcalico/cni-plugin/pull/698/commits/67ece493d341032c9b39c0ee13824a3e680d1e39

Makefile
caseydavenport: Resolved in this commit: https://github.com/projectcalico/cni-plugin/pull/698/commits/67ece493d341032c9b39c0ee13824a3e680d1e39

Makefile
caseydavenport: Resolved in this commit: https://github.com/projectcalico/cni-plugin/pull/698/commits/67ece493d341032c9b39c0ee13824a3e680d1e39

tests/calico_cni_k8s_test.go
```diff
@@ -2004,12 +2142,18 @@ var _ = Describe("Kubernetes CNI tests", func() {
...
 			endpointSpec = endpoints.Items[0].Spec
-			Expect(endpointSpec.ContainerID).Should(Equal(containerID))
+			if os.Getenv("DATASTORE_TYPE") != "kubernetes" {
+				Expect(endpointSpec.ContainerID).Should(Equal(containerID))
+			}
 			checkIPAMReservation()
 		})
 
 		AfterEach(func() {
+			// Delete pod
```
caseydavenport: I think this actually breaks the testcase right below with this error:

```
  Expected error:

      <*types.Error | 0xc000f4bd40>: {

          Code: 100,

          Msg: "pods \"run601964353\" not found",

          Details: "",

      }

      pods "run601964353" not found

  not to have occurred
```

tests/calico_cni_k8s_test.go
song-jiang: Updated code to do a `Get` first. And expect we do not have errors after `Delete`.

tests/calico_cni_k8s_test.go
song-jiang: I think I tried to delete pod here because we are checking WEP count later. I have made changes that we only check WEP count if it is etcd datastore and added 
`defer ensurePodDeleted(clientset, testutils.K8S_TEST_NS, name)`  to make sure we cleanup pod.

tests/calico_cni_k8s_test.go
song-jiang: @lwr20 Could you address this comment?

tests/calico_cni_k8s_test.go
song-jiang: > Can we do this in an afterEach as well?

Yes. I have updated the code so that, if there is only one test case, I would use 
```defer ensurePodDeleted(clientset, testutils.K8S_TEST_NS, name)``` right after pod been created. 

Otherwise, `ensurePodDeleted` is calling from `afterEach` for multiple test cases.

Comment (song-jiang) [reply tests/calico_cni_k8s_test.go]: > I'm still a bit curious as to why these passed before, but now need this.. feel like I must be missing something!

Test passed before with etcd datastore. This is because `wipeDatastore` cleanup calico resources include WEPs. In kubernetes datastore, we need to clean up Pod as WEP resources for each test.

File: tests/calico_cni_k8s_test.go (hunk 5 above)
Comment (lwr20) [reply tests/calico_cni_k8s_test.go]: Fine.  Leave it out and see what breaks :)

File: tests/calico_cni_k8s_test.go (hunk 14 above)
Comment (song-jiang) [reply tests/calico_cni_k8s_test.go]: I did not reproduce your issue here. But I will adjust the order for AfterEach:
```
		AfterEach(func() {
			_, err = testutils.DeleteContainerWithId(netconf, contNs.Path(), name, testutils.K8S_TEST_NS, containerID)
			Expect(err).ShouldNot(HaveOccurred())

			ensurePodDeleted(clientset, testutils.K8S_TEST_NS, name)
		})
```

File: tests/calico_cni_test.go (hunk 8 above)
Comment (song-jiang) [reply tests/calico_cni_test.go]: Ouch.. Removed.

File: tests/calico_cni_k8s_test.go (hunk 5 above)
Comment (song-jiang) [reply tests/calico_cni_k8s_test.go]: I guess we added `AddNode` in `BeforeEach`, so when creating a node with same name but different PodCIDR, the node could still exists. `DeleteNode` will take some time to remove a node. 

I added function `ensureNodeDeleted` similar to `ensurePodDeleted` and use that to make sure node is not exits before creating a new node.

File: tests/calico_cni_k8s_test.go (hunk 1 above)
Comment (caseydavenport) [reply tests/calico_cni_k8s_test.go]: I think either way we don't want `panics()` in the test code if we can avoid them - they make it a lot harder to read / understand test output and end the test short rather than running all to completion.

File: tests/calico_cni_k8s_test.go (hunk 4 above)
Comment (caseydavenport) [reply tests/calico_cni_k8s_test.go]: This case isn't doing a defer though, so I don't think it's quite right. I think so long as we do this at the end it is OK but we should make sure they're all properly deferred or, ideally, in an `AfterEach`.

File: tests/calico_cni_k8s_test.go (hunk 5 above)
Comment (caseydavenport) [reply tests/calico_cni_k8s_test.go]: > I guess we added AddNode in BeforeEach

But, we always used to create a node in the BeforeEach AFAICT, we just called it directly. My questions is why do we suddenly need this new logic? It doesn't feel quite right to me but I might just be misunderstanding.

File: tests/calico_cni_k8s_test.go (hunk 6 above)
Comment (caseydavenport) [reply tests/calico_cni_k8s_test.go]: > defer ensurePodDeleted(clientset, testutils.K8S_TEST_NS, name) right after pod been created.

My preference is to simply use BeforeEach and AfterEach anyway, since we might add additional tests to those contexts later and the mix of defers / AfterEach adds cognitive overhead. Let's not worry about adjusting that now, but we can clean it up in the future.

File: tests/calico_cni_k8s_test.go [hunk 15]
```diff
@@ -2125,6 +2269,9 @@ var _ = Describe("Kubernetes CNI tests", func() {
 			// Make sure the pod gets cleaned up, whether we fail or not.
 			expectedIfaceName := "eth0"
 			defer func() {
+				// Delete pod
```
caseydavenport: This also should go after the `DeleteContainerWithIdAndIfaceName`

tests/calico_cni_k8s_test.go
song-jiang: Yes. Ideally all defer should be in `AfterEach`. However, there are cases that two pod with different names been created where we use `defer`

tests/calico_cni_k8s_test.go
song-jiang: We used to create a calico node but not a k8s node I think. 
https://github.com/projectcalico/cni-plugin/blob/master/tests/calico_cni_k8s_test.go#L78

That is why we need a logic now to make sure node is deleted before we create it again.

PR #720, opened 2019-03-29
Fix nil pointer exception for node bgp spec

kahou82:
Fix of issue: https://github.com/projectcalico/cni-plugin/issues/719


caseydavenport:
@kahou82 thanks for this. However, I think this isn't the right fix. I think we probably shouldn't be executing this script at all for v6 only clusters.

I suspect there's something else going on here.

caseydavenport:
Alternative fix included in this PR (with some other changes as well): https://github.com/projectcalico/cni-plugin/pull/726

PR #723, opened 2019-04-15
adding fixed IPs for Mesos like k8s annotations

newffy:
This adds fixed IPs support to Mesos similar to k8s annotations via NetworkInfo.Labels and also structures code to cni_default and k8s modules.

Example of usage (fragment of Marathon's app decription):
```
"networks": [
    {
      "name": "cni-net-fixedpool",
      "mode": "container",
      "labels": {
        "cni.projectcalico.org/ipAddrs": "[\"10.11.12.13\"]"
      }
    }
  ],
  "upgradeStrategy": {
    "maximumOverCapacity": 0,
    "minimumHealthCapacity": 0
  }
```

### DIFF
--- a/pkg/cni_default/cni_default.go
+++ b/pkg/cni_default/cni_default.go
@@ -1,0 +1,294 @@
+// Copyright 2015 Tigera Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cni_default
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types/current"
+	"github.com/projectcalico/cni-plugin/internal/pkg/utils"
+	"github.com/projectcalico/cni-plugin/pkg/types"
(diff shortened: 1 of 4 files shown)


newffy:
> This looks like a lot more lines changed than I would expect for this - @newffy how much of this is required for this feature?

it's basically refactoring. I moved common functions from k8s package to new cni_default package and moved some code from plugin.go to cni_default package to make plugin.go cleaner ( I speak about
```	if wepIDs.Orchestrator == api.OrchestratorKubernetes {
		if result, err = k8s.CmdAddK8s(ctx, args, conf, *wepIDs, calicoClient, endpoint); err != nil {
			return err
		}
	} else {
		// Default CNI behavior
		if result, err = cni_default.CmdAddCNIDefault(ctx, args, conf, *wepIDs, calicoClient, endpoint); err != nil {
			return err
		}
	}
``` 
and the same for DEL command:
```if epIDs.Orchestrator == api.OrchestratorKubernetes {
 		return k8s.CmdDelK8s(ctx, calicoClient, *epIDs, args, conf, logger)		 		return k8s.CmdDelK8s(ctx, calicoClient, *epIDs, args, conf, logger)
 	} else {
 		return cni_default.CmdDelCNIDefault(ctx, calicoClient, *epIDs, args, conf, logger)
 	}		 	
}```
where `cni_default.CmdAddCNIDefault` and `cni_default.CmdDelCNIDefault` mostly copy-paste from original plugin.go code. 
Also I moved common code for k8s and default cni to `cni_default.ProcessIpAddrs`

newffy:
..Maybe I shouldn't do that, I can try to formulate more concise pull-request. 
I just tried to make affected more clean and readable.. for me at least .)

caseydavenport:
> ..Maybe I shouldn't do that, I can try to formulate more concise pull-request.
I just tried to make affected more clean and readable.. for me at least .)

Yeah, I'd like to do this with a few lines of code changes as possible - it makes it a lot harder to review  :)

I think this should be a pretty small change.

caseydavenport:
Closing this for now just because of inactivity and the substantial number of changes. Would be happy to review a change set which doesn't include such a substantial refactor.

PR #726, opened 2019-04-17
Handle nil pointers better in migration script

caseydavenport:
Fixes https://github.com/projectcalico/calico/issues/2516

- [ ] Tests
- [ ] Documentation
- [ ] Release note

```release-note
None required
```

### DIFF
--- a/pkg/upgrade/migrate.go
+++ b/pkg/upgrade/migrate.go
@@ -90,7 +90,7 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
 			return fmt.Errorf("PodCIDR %s did not parse successfully: %s", k8sNode.Spec.PodCIDR, err)
 		} else if cidr.Version() == 4 {
 			// We need to get the IP for the podCIDR and increment it to the
-			// first IP in the CIDR.
+			// first IP in the CIDR to match the behavior used by Calico when using host-local IPAM.
 			tunIp := ip.To4()
 			if tunIp == nil {
 				return fmt.Errorf("Cannot pick an IPv4 tunnel address from the given CIDR: %s", k8sNode.Spec.PodCIDR)
@@ -116,7 +116,7 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
 				if _, ok := err.(errors.ErrorResourceAlreadyExists); !ok {
 					return fmt.Errorf("failed to get add IPIP tunnel addr %s: %s", node.Spec.BGP.IPv4IPIPTunnelAddr, err)
 				}
-				log.Info("IPIP tunnel address already assigned, continuing...")
+				log.Info("IPIP tunnel address already assigned in IPAM, continuing...")
 			}
 
 			// Save the calico node object to the datastore.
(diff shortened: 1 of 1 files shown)


pkg/upgrade/migrate.go
```diff
@@ -160,27 +161,29 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
...
+			return fmt.Errorf("failed to fetch cluster information: %s", err)
+		}
+		if clusterInfo.Spec.DatastoreReady == nil || *clusterInfo.Spec.DatastoreReady {
+			f := false
+			clusterInfo.Spec.DatastoreReady = &f
 			if clusterInfo, err = c.ClusterInformation().Update(ctxt, clusterInfo, options.SetOptions{}); err != nil {
 				if _, ok := err.(errors.ErrorResourceUpdateConflict); ok {
 					log.Info("Encountered update conflict, retrying...")
-					time.Sleep((1 << i) * time.Second)
+					time.Sleep(1 * time.Second)
```
jpalardy: no more exponential backoff?

pkg/upgrade/migrate.go
```diff
@@ -333,7 +336,8 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
 	// calicoDS, err := k8sClient.AppsV1().DaemonSets("kube-system").Get("calico-node", metav1.GetOptions{})
 	// if (calicoDS.Status.UpdatedNumberScheduled == calicoDS.Status.DesiredNumberScheduled-calicoDS.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.IntVal)
 	log.Info("setting Calico datastore readiness to true...")
-	*clusterInfo.Spec.DatastoreReady = true
+	t := true
+	clusterInfo.Spec.DatastoreReady = &t
```
jpalardy: that might be a question of style, or a golang nuance: is there an advantage to doing it this way rather than `*clusterInfo.Spec.DatastoreReady = true` ?

pkg/upgrade/migrate.go
```diff
@@ -160,27 +161,29 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
...
+	// Disable cni by setting DatastoreReady to false.
 	log.Info("setting datastore readiness to false")
-	clusterInfo, err := c.ClusterInformation().Get(ctxt, "default", options.GetOptions{})
-	if err != nil {
-		return fmt.Errorf("failed to fetch cluster information: %s", err)
-	}
-	if *clusterInfo.Spec.DatastoreReady {
-		*clusterInfo.Spec.DatastoreReady = false
-		for i := uint(0); i < 5; i++ {
+	var clusterInfo *v3.ClusterInformation
```
jpalardy: why `clusterInfo` outside the for-loop? I don't see it used afterward.

pkg/upgrade/migrate.go
```diff
@@ -160,27 +161,29 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
...
+	for i := uint(0); i < 5; i++ {
+		clusterInfo, err = c.ClusterInformation().Get(ctxt, "default", options.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to fetch cluster information: %s", err)
+		}
+		if clusterInfo.Spec.DatastoreReady == nil || *clusterInfo.Spec.DatastoreReady {
```
jpalardy: I'm unclear on why `DatastoreReady` was a `*bool` in the first place. As a plain bool, we wouldn't have had this nil problem.

pkg/upgrade/migrate.go
caseydavenport: Nope, I don't think it's what we want here. 

This shouldn't fail, and if it does it shouldn't be persistent - the exponential backoff will just make this take longer than it should.

pkg/upgrade/migrate.go
caseydavenport: I think it's probably fine the way it was... I could probably revert, but I like this better.

I generally like this better because it doesn't dereference a pointer, which might be nil. I think in this case it wouldn't be.. but in other cases we might panic.

pkg/upgrade/migrate.go
caseydavenport: Yeah, same. I think it was to distinguish between "not set" which should mean "true", and explicitly set to "false", but in retrospect it probably could have just been a bool.

PR #728, opened 2019-04-19
Adding -C option to curl to ensure download of cni-plugins wont fail

GeorgeGuo2018:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ TEST_SRCFILES=$(shell find tests -name '*.go')
 LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
 
 # fail if unable to download
-CURL=curl -sSf
+CURL=curl -C - -sSf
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -71,6 +71,7 @@ LOCAL_IP_ENV?=$(shell ip route get 8.8.8.8 | head -1 | awk '{print $$7}')
 
 # fail if unable to download
 CURL=curl -sSf
+CURL_S=curl -C - -sSf
```
caseydavenport: I think with this change, `$CURL` is no longer used anywhere - why don't we drop the `CURL_C` variable and just add the `-C` argument to `CURL`?

caseydavenport:
@GeorgeGuo2018 you'll need to sign the CLA to make the bot happy before we can merge. Thanks!

Makefile
GeorgeGuo2018: You‘re right.

PR #730, opened 2019-04-24
Retry on IPv6 listing

CallMeFoxie:
## Description
A fix for https://github.com/projectcalico/calico/issues/2553

Detailed info what is happening is in the link, this is probably the easiest fix of all. I've been testing the code (or at least its calico 3.3 branch) for a while now and it seems to be working great (actually triggered several retries). 

```
Apr 24 09:10:17 <hostname redacted> kubelet[13796]: 2019-04-24 09:10:17.326 [INFO][19173] network.go 212: No IPv6 set on interface, retrying.. ContainerID="f5276c4c91358bfa76ee56c654e5ce83e3aca22dccec5104f0c581632a87510a" Namespace="default" Pod="test-ipv6-fvirtbase-1556089800-lggbv" WorkloadEndpoint="<hostname redacted>-k8s-test--ipv6--fvirtbase--1556089800--lggbv-eth0"
```

The bug was there always, just a recent change either in kubelet or something has made the logic faster and triggered the bug :).

IMHO it should be backported to previous branches as well as not everybody runs the latest Calico (we have 3.3 in production atm). 

Thanks for looking at this,
Ashley

## Release Note

```release-note
None required
```

### DIFF
--- a/internal/pkg/utils/network_linux.go
+++ b/internal/pkg/utils/network_linux.go
@@ -187,20 +187,35 @@ func DoNetworking(
 				return fmt.Errorf("failed to set net.ipv6.conf.lo.disable_ipv6=0: %s", err)
 			}
 
-			// No need to add a dummy next hop route as the host veth device will already have an IPv6
-			// link local address that can be used as a next hop.
-			// Just fetch the address of the host end of the veth and use it as the next hop.
-			addresses, err := netlink.AddrList(hostVeth, netlink.FAMILY_V6)
-			if err != nil {
-				logger.Errorf("Error listing IPv6 addresses for the host side of the veth pair: %s", err)
-				return err
+			// Retry several times as the LL can take a several micro/miliseconds to initialize and we may be too fast
+			// after these sysctls
+			var err error
+			var addresses []netlink.Addr
+			for i := 0; i < 10; i++ {
+				// No need to add a dummy next hop route as the host veth device will already have an IPv6
+				// link local address that can be used as a next hop.
+				// Just fetch the address of the host end of the veth and use it as the next hop.
+				addresses, err = netlink.AddrList(hostVeth, netlink.FAMILY_V6)
+				if err != nil {
+					logger.Errorf("Error listing IPv6 addresses for the host side of the veth pair: %s", err)
+				}
+
+				if len(addresses) < 1 {
+					// If the hostVeth doesn't have an IPv6 address then this host probably doesn't
+					// support IPv6. Since a IPv6 address has been allocated that can't be used,
+					// return an error.
+					err = fmt.Errorf("failed to get IPv6 addresses for host side of the veth pair")
+				}
+				if err == nil {
+					break
+				}
+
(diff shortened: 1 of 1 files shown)


internal/pkg/utils/network_linux.go
```diff
@@ -187,20 +187,33 @@ func DoNetworking(
...
 			}
 
-			// No need to add a dummy next hop route as the host veth device will already have an IPv6
-			// link local address that can be used as a next hop.
-			// Just fetch the address of the host end of the veth and use it as the next hop.
-			addresses, err := netlink.AddrList(hostVeth, netlink.FAMILY_V6)
-			if err != nil {
-				logger.Errorf("Error listing IPv6 addresses for the host side of the veth pair: %s", err)
-				return err
+			var err error
```
fasaxc: Please can you add a comment explaining why the retry is needed?   Something like `// Sometimes the IPv6 address takes a short time to appear after the sysctls are enabled.  Retry a few times.`

CallMeFoxie: added :)

internal/pkg/utils/network_linux.go
```diff
@@ -187,20 +187,35 @@ func DoNetworking(
...
+					logger.Errorf("Error listing IPv6 addresses for the host side of the veth pair: %s", err)
+				}
+
+				if len(addresses) < 1 {
+					// If the hostVeth doesn't have an IPv6 address then this host probably doesn't
+					// support IPv6. Since a IPv6 address has been allocated that can't be used,
+					// return an error.
+					err = fmt.Errorf("failed to get IPv6 addresses for host side of the veth pair")
+				}
+				if err == nil {
```
aojea: what err is checking this condition?

CallMeFoxie: Whoops misread.
This is the one checking if we can break after less than 10 retries.

fasaxc: It terminates the retry loop if there is no error at both 198 and 207

CallMeFoxie: Yeah, check my 2nd reply, I was looking at the wrong err.

It just makes it not retry 10 times (as in that case there may have just been time.Sleep(500ms)) but continue as fast as possible.

CallMeFoxie: Should I still change it to flip the `if len(addresses)` or is this ok? :) (missed that due to the edit)

fasaxc: I think it's OK, it's how the old code worked so no _need_ to change it.

caseydavenport:
^ I've submitted some backport PRs to v3.7, v3.6, and v3.5.

CallMeFoxie:
Thanks, I will either upgrade our ancient 3.3 to something newer or keep the update I tested it with :)

Can I change the release note somehow? Or will it get auto-set from the cherry picks?

caseydavenport:
I think the release note is all set - it should get pulled in from the cherry picks.

PR #746, opened 2019-05-22
Changed order of releasing IP and removing interface

jadech32:
## Description

Bug fix: Changed ordering of events so that IP is released after the interface is removed to prevent potential race condition / two interfaces sharing one IP

## Todos
- [x] Tests
- [ ] Documentation
- [ ] Release note

### DIFF
--- a/pkg/k8s/k8s.go
+++ b/pkg/k8s/k8s.go
@@ -473,16 +473,16 @@ func CmdDelK8s(ctx context.Context, c calicoclient.Interface, epIDs utils.WEPIde
 		}
 	}
 
-	// Release the IP address for this container by calling the configured IPAM plugin.
-	logger.Info("Releasing IP address(es)")
-	ipamErr := utils.DeleteIPAM(conf, args, logger)
-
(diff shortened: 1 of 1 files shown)


pkg/k8s/k8s.go
```diff
@@ -473,17 +473,17 @@ func CmdDelK8s(ctx context.Context, c calicoclient.Interface, epIDs utils.WEPIde
...
 	err = utils.CleanUpNamespace(args, logger)
 	if err != nil {
 		return err
 	}
 
+	// Release the IP address for this container by calling the configured IPAM plugin.
+	logger.Info("Releasing IP address(es)")
+	ipamErr := utils.DeleteIPAM(conf, args, logger)
+
 	// Return the IPAM error if there was one. The IPAM error will be lost if there was also an error in cleaning up
```
caseydavenport: This comment isn't correct any more - I think we can just remove it?

Also, we can rename `ipamErr` to `err` now since we're not dealing with two errors at a time.

Issue #749, opened 2019-05-28
upgrade-ipam initContainer fails when calico-node is restarted


maxbischoff:
When upgrading from v3.5.4 to v3.7.2 we sometimes observed an issue where calico-node on single nodes sometimes fails to restart when its pod was killed due to upgrade-ipam failing, as migrate.go attempts to write IP addresses into ` IPAMBlock` which already contain those addresses. A symptom of this is upgrade-ipam not skipping migration, because `/var/lib/cni/networks/k8s-pod-network` is not properly cleaned.

## Expected Behavior

upgrade-ipam should skip migration in subsequent restarts of calico-node pods and therefore not fail.

## Current Behavior

Migration fails in `ipam.go` when assignin IPs in IPAMBlocks, see log below:

```
019-05-27 15:41:45.340 [INFO][1] migrate.go 64: checking host-local IPAM data dir dir existence...
2019-05-27 15:41:45.340 [INFO][1] migrate.go 71: retrieving node for IPIP tunnel address
2019-05-27 15:41:45.343 [INFO][1] migrate.go 134: checking if host-local IPAM data dir dir is empty...
2019-05-27 15:41:45.343 [INFO][1] migrate.go 157: host-local IPAM data dir is not empty, migrating...
2019-05-27 15:41:45.343 [INFO][1] migrate.go 165: setting datastore readiness to false
2019-05-27 15:41:45.357 [INFO][1] migrate.go 186: successfully set datastore readiness to false
2019-05-27 15:41:45.357 [INFO][1] migrate.go 189: removing cni binaries...
2019-05-27 15:41:45.357 [INFO][1] migrate.go 194: successfully removed cni binary! binary="/host/opt/cni/bin/calico"
2019-05-27 15:41:45.357 [INFO][1] migrate.go 194: successfully removed cni binary! binary="/host/opt/cni/bin/host-local"
2019-05-27 15:41:50.357 [INFO][1] migrate.go 201: acquiring lock on host-local IPAM
2019-05-27 15:41:50.357 [INFO][1] migrate.go 209: successfully acquired lock on host-local IPAM
2019-05-27 15:41:50.357 [INFO][1] migrate.go 222: mapping pod ips to pods...
2019-05-27 15:41:50.372 [INFO][1] migrate.go 232: mapping in pod IP="100.64.3.14" namespace="ams-testing-dev" pod="nc-pod-830748-5b7d8c44d7-zpzrv"
[...]

PR #750, opened 2019-05-28
Fix upgrade-ipam lockin wrong directory

maxbischoff:
## Description

This is a bug fix for an issue observed when upgrading from Calico 3.5.4 to 3.7.2 (see #749).
I am currently manually testing this fix and am open to any suggestions.

### Explanation
The old call to `disk.New` causes creation of `/var/lib/cni/networks/var/lib/cni/networks/k8s-pod-network`. This is because supplying `""` as `dataDir` causes the default of `/var/lib/cni/networks` to be used. The `network` value of `"/var/lib/cni/networks/k8s-pod-network/"` is then concatenated to that default value.
This in turn causes the wrong directory to be locked.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/upgrade/migrate.go
+++ b/pkg/upgrade/migrate.go
@@ -199,7 +199,7 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
 	// against racing with any remaining host-local processes which might be allocating
 	// IP addresses.
 	log.Info("acquiring lock on host-local IPAM")
-	hostLocal, err := disk.New(ipAllocPath, "")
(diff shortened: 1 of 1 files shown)


caseydavenport:
@maxbischoff thanks for the PR, and for digging into this.

This definitely looks like a bug - please let me know if you discover this to be the cause of 
https://github.com/projectcalico/cni-plugin/issues/749 or if it is something different. At a glance, it isn't obvious to me.

PR #754, opened 2019-06-05
Handle migrating an IP that is already migrated

neiljerram:
This is a partial fix for https://github.com/projectcalico/cni-plugin/issues/749, q.v.

Clearly the intent of the code here is to be resilient in the case
where an IP (as indicated by a host-local file) has already been
assigned in the Calico data model; but the condition used to detect this
case was insufficient.  Here's a log excerpt that shows what the actual
error is here:
```
2019-05-27 15:31:52.774 [INFO][1] migrate.go 238: reading files from host-local IPAM data dir...
2019-05-27 15:31:52.774 [INFO][1] migrate.go 247: processing file... file="100.64.3.11"
2019-05-27 15:31:52.774 [INFO][1] migrate.go 291: assinging pod IP to Calico IPAM... file="100.64.3.11"
2019-05-27 15:31:52.774 [INFO][1] ipam.go 581: Assigning IP 100.64.3.11 to host: kworker-be-sandbox-iz2-bap004
2019-05-27 15:31:52.791 [ERROR][1] ipam.go 635: Failed to assign address 100.64.3.11: Address already assigned in block
```

### DIFF
--- a/pkg/upgrade/migrate.go
+++ b/pkg/upgrade/migrate.go
@@ -1,4 +1,4 @@
-// Copyright 2015 Tigera Inc
+// Copyright 2015-2019 Tigera Inc
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


caseydavenport:
Sounds reasonable to me

PR #772, opened 2019-08-15
Update OSS semaphore jobs that depend on glide

rafaelvanoni:


Makefile
```diff
@@ -46,6 +46,15 @@ LOCAL_GROUP_ID:=$(shell id -g)
 
 EXTRA_DOCKER_ARGS	+= -e GO111MODULE=on
 
+ifdef GOPATH
+ifneq ($(GOPATH),)
+ifndef CIRCLECI
+	LOCAL_GOPATH = $(shell echo $(GOPATH) | cut -d':' -f1)
```
ozdanborne:
```suggestion
	# If the environment is using multiple comma-separated directories for gopath, use the first one, as that 
	# is the default one used by go modules.
	LOCAL_GOPATH = $(shell echo $(GOPATH) | cut -d':' -f1)
```

Makefile
```diff
@@ -46,6 +46,15 @@ LOCAL_GROUP_ID:=$(shell id -g)
 
 EXTRA_DOCKER_ARGS	+= -e GO111MODULE=on
 
+ifdef GOPATH
+ifneq ($(GOPATH),)
+ifndef CIRCLECI
```
ozdanborne:
```suggestion
# CircleCI's gopath is readonly and readonly gopaths are incompatible with go modules so don't
# volume mount the cache in that environment
ifndef CIRCLECI
```

Makefile
```diff
@@ -46,6 +46,15 @@ LOCAL_GROUP_ID:=$(shell id -g)
 
 EXTRA_DOCKER_ARGS	+= -e GO111MODULE=on
 
+ifdef GOPATH
```
ozdanborne:
```suggestion
# Volume-mount gopath into the build container if it's defined for persistent caching.
ifdef GOPATH
```

ozdanborne: I think this if statement is overkill. Golang treats gopath the same whether it's set to an empty string or unset.

```
(⎈ N/A:N/A)❯ export GOPATH=""                                     (turk@tigera.io:tigera-dev)

~/tmp
(⎈ N/A:N/A)❯ go env GOPATH                                        (turk@tigera.io:tigera-dev)
/home/turk/go

~/tmp
(⎈ N/A:N/A)❯ unset GOPATH                                         (turk@tigera.io:tigera-dev)

~/tmp
(⎈ N/A:N/A)❯ go env GOPATH                                        (turk@tigera.io:tigera-dev)
/home/turk/go
```

Our makefile logic shouldn't care either.

Can we just rely on the following `ifneq` statement?

rafaelvanoni: I think so

PR #777, opened 2019-08-28
Re-enable golangci-lint

rafaelvanoni:


internal/pkg/testutils/utils.go
```diff
@@ -269,8 +269,6 @@ func CreateContainer(netconf, podName, podNamespace, ip string) (containerID str
 }
 
 // Create container with the giving containerId when containerId is not empty
-//
-// Deprecated: Please call CreateContainerNamespace and then RunCNIPluginWithID directly.
```
caseydavenport: Why did you remove this comment? I think we don't want people using the function in the test code.

internal/pkg/testutils/utils.go
```diff
@@ -584,12 +581,11 @@ func AddNode(c client.Interface, kc *kubernetes.Clientset, host string) error {
 }
 
 func DeleteNode(c client.Interface, kc *kubernetes.Clientset, host string) error {
-	var err error
-	err = nil
+	var err error = nil
```
caseydavenport: I don't see us ever using this - shouldn't we just change line 596 to `return nil`?

internal/pkg/testutils/utils.go
rafaelvanoni: One of the linters enforces `Deprecated` comments, but I spoke to @ozdanborne and he told me to ignore it and stick with the current call.

internal/pkg/testutils/utils.go
rafaelvanoni: We still use `err` on line 593 and return that value.

caseydavenport:
@rafaelvanoni LGTM - could you squash before merging?

PR #807, opened 2019-11-11
Add timeouts to cmdAdd() and cmdDel()

rafaelvanoni:
Proposed fix for https://github.com/projectcalico/calico/issues/2902

### DIFF
--- a/pkg/ipamplugin/ipam_plugin.go
+++ b/pkg/ipamplugin/ipam_plugin.go
@@ -155,6 +155,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 	}
 
 	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 120*time.Second)
+	defer cancel()
+
 	r := &current.Result{}
 	if ipamArgs.IP != nil {
 		logger.Infof("Calico CNI IPAM request IP: %v", ipamArgs.IP)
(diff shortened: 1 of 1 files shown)


pkg/ipamplugin/ipam_plugin.go
```diff
@@ -207,12 +207,16 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 		logger.Infof("Calico CNI IPAM request count IPv4=%d IPv6=%d", num4, num6)
 
-		v4pools, err := utils.ResolvePools(ctx, calicoClient, conf.IPAM.IPv4Pools, true)
+		ctx2, cancel := context.WithTimeout(ctx, 10*time.Second)
```
tmjd: I think we should just make ctx be a `context.WithTimeout`, I think this whole cmdAdd function could use that one context and should all be done in 10 seconds.
I think it would be reasonable to make the same changes in the cmdDel.

rafaelvanoni: Fixed, please take a look.

pkg/ipamplugin/ipam_plugin.go
```diff
@@ -155,6 +155,9 @@ func cmdAdd(args *skel.CmdArgs) error {
 	}
 
 	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, 120*time.Second)
```
tmjd: I suggested switching the context for the cmdAdd to have a timeout (instead of just on the IPPools call). As I understand it, @rafaelvanoni had to increase the timeout from 10s to 120s to get the tests to pass.
@caseydavenport do you have any idea why there would be that much time spent in here?

@rafaelvanoni Looking back at the build failures, I see in build 4 that it shows many of the failures took 0.0x seconds. Do you know if that build had the shorter timeout?

rafaelvanoni: That was either a 10 or 30 sec timeout

tmjd: With the 10 sec timeout again. We've got 1 failure, 2 passes so far. I've got another run going.

PR #809, opened 2019-11-20
Move common Makefile targets to Makefile.common

rafaelvanoni:


Makefile
```diff
@@ -173,57 +62,15 @@ clean:
...
 # Building the binary
 ###############################################################################
 ## Build the Calico network plugin and ipam plugins
+BIN=bin/$(ARCH)
 $(BIN)/calico $(BIN)/calico-ipam: local_build $(SRC_FILES)
-	$(DOCKER_RUN) \
-	-v $(CURDIR)/$(BIN):/go/src/$(PACKAGE_NAME)/$(BIN):rw \
-	    $(CALICO_BUILD) sh -c '\
+	$(DOCKER_RUN) $(CALICO_BUILD) sh -c '\
 		go build -v -o $(BIN)/calico -ldflags "-X main.VERSION=$(GIT_VERSION) -s -w" ./cmd/calico && \
```
ozdanborne: I tried this out locally and after a successful `make build`, my binary works! But it's missing the version:
```
bin/amd64/calico-ipam -v

```

ozdanborne:
One more issue: previously, there was a `default` build target so if just run `make` with no target, the binary is built. Doesn't seem to be happening anymore. Could you restore that functionality?

PR #819, opened 2019-12-17
Connect to datastore before installing CNI config

spikecurtis:
Signed-off-by: Spike Curtis <spike@tigera.io>

## Description

If we can't connect to the datastore, but install the CNI config, Kubelet may start scheduling pods, which will just fail on this node.  So, before adding the CNI config, test a connection to the datastore. Once that is successful, proceed to add the CNI config.  If the datastore connection fails we retry in a loop.

## Todos
- [ ] Tests
- [x] Documentation
- [ ] Release note

## Release Note

```release-note
wait for successful connect to datastore before configuring kubelet with CNI plugin
```

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -19,8 +19,10 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"runtime"
+	"time"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	cnitypes "github.com/containernetworking/cni/pkg/types"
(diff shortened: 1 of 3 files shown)


pkg/plugin/plugin.go
```diff
@@ -36,15 +38,67 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("error getting ClusterInformation: %v", err)
+	}
+	if !*ci.Spec.DatastoreReady {
+		logrus.Info("Upgrade may be in progress, ready flag is not set")
+		return fmt.Errorf("Calico is currently not ready to process requests")
+	}
+
+	// If we have a kubeconfig, test connection to the APIServer
+	if conf.Kubernetes.Kubeconfig != "" {
```
caseydavenport: If we're using KDD mode, then we'll have already tested this above. Maybe worth gating this check on etcd mode only? Then again, it doesn't _hurt_ to have two checks 🤔

spikecurtis:
I think that the `[It] Should install bins and config` test case at https://github.com/projectcalico/cni-plugin/blob/64c92d35c881b384e4c565fb663186bc7dc90579/k8s-install/scripts/install_cni_test.go#L126 will exercise all this code in the "datastore available" case

PR #820, opened 2019-12-18
Remove blank and comment lines

xieyanker:
## Description

Remove blank and comment lines before `mv "$TMP_CONF" /host/etc/cni/net.d/"${CNI_CONF_NAME}"`.
This has the benefit of allowing comments and blank lines in the TMP_CONF file.

In some situations, I have to use `ansible` to update `calico-cni`'s config file, but it might add some comment lines automatically, this will cause error: `error parsing configuration list: invalid character '#' looking for beginning of object key string`. So I want to remove blank and comment lines before convert temp file to target file.

I have tested the effect of command `sed -i '/^$/d;/^#/d'` as follows:
```shell
root@master1:[/root]cat test_file 
#This is a comment line
1234

5678
foo
#Another comment line
bar
root@master1:[/root]sed -i '/^$/d;/^#/d' test_file 
root@master1:[/root]cat test_file 
1234
5678
foo
bar
```

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -179,6 +179,9 @@ sed -i s~__ETCD_ENDPOINTS__~"${ETCD_ENDPOINTS:-}"~g $TMP_CONF
 sed -i s~__ETCD_DISCOVERY_SRV__~"${ETCD_DISCOVERY_SRV:-}"~g $TMP_CONF
 sed -i s~__LOG_LEVEL__~"${LOG_LEVEL:-warn}"~g $TMP_CONF
 
+# Remove blank and comment lines
+sed -i '/^$/d;/^#/d' $TMP_CONF
+
 CNI_CONF_NAME=${CNI_CONF_NAME:-10-calico.conf}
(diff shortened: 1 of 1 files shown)


tmjd:
It seems wrong to add something to work around an issue with Ansible. Is there not a way to have ansible not put those lines you're trying to remove or have it remove those lines? From my ansible understanding couldn't you write the template to a file and then run that `sed` on that file then use the contents of that file.

xieyanker:
@tmjd Thanks for your reply. In fact, I can avoid this issue by better using ansible. However, what I want to resolve is not the matter caused by ansible. I think it's necessary to allow blank and comment lines exist in template file, this can improve fault tolerance.

caseydavenport:
Thanks for the PR and the idea, but I don't think Calico should support comments within the template I agree with @tmjd - this is extending the scope of Calico's capabilities beyond what I personally think is appropriate. 

You should make sure that the template you are generating in ansible is valid json so that it will be correctly parsed by the CNI runtime. Calico shouldn't be in the business of taking invalid json and "fixing" it to be valid.

PR #828, opened 2020-01-03
release-v3.10 - Update libcalico-go

lmm:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/natefinch/atomic v0.0.0-20150920032501-a62ce929ffcc
 	github.com/onsi/ginkgo v1.6.0
 	github.com/onsi/gomega v1.4.2
-	github.com/projectcalico/libcalico-go v0.0.0-20191104213956-8f81e1e344ce
+	github.com/projectcalico/libcalico-go v1.7.2-0.20191223230650-31f24eab280c
```
caseydavenport: This doesn't seem right - why does it say v1.7.2?

lmm: We talked about this in Slack. This is due to different go versions used when the go.mod was updated. Both pseudo versions are valid:
https://sum.golang.org/lookup/github.com/projectcalico/libcalico-go@v0.0.0-20191223230708-3d65d3751012
https://sum.golang.org/lookup/github.com/projectcalico/libcalico-go@v1.7.2-0.20191223230708-3d65d3751012

The `v1.7.2` is because that's the latest release tag in libcalico-go that go recognizes.

PR #833, opened 2020-01-16
Fix release image push

lmm:
## Description

The invocation ` make tag-images-all RELEASE=true IMAGETAG=<tag>` was not tagging and pushing the image to gcr.io registries.

This PR addresses part of that: the `PUSH_IMAGES` variable needs to be defined before `Makefile.common` is included, otherwise `PUSH_IMAGES` will lose its initial default values. The second part is fixed in the common Makefile in https://github.com/projectcalico/go-build/pull/95

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PACKAGE_NAME=github.com/projectcalico/cni-plugin
-GO_BUILD_VER=v0.31
+GO_BUILD_VER=v0.34
 
 # This needs to be evaluated before the common makefile is included.
 # This var contains some default values that the common makefile may append to.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,5 +1,9 @@
 PACKAGE_NAME=github.com/projectcalico/cni-plugin
-GO_BUILD_VER=v0.31
+GO_BUILD_VER=v0.34
```
lmm: A to-be-released version of go-build

PR #836, opened 2020-02-10
Put spaces around inside of `[ ]` for bash compat

### LINKED ISSUE #835: install-cni.sh bash syntax error
## Current Behavior
`install-cni.sh` fails with this error
```
/install-cni.sh: line 195: [-z: command not found
```

## Possible Solution
Bash syntax needs spaces inside the square brackets here.
https://github.com/projectcalico/cni-plugin/blob/95276869e7b6536403f4f387fa086f8d7c1e4d90/k8s-install/scripts/install-cni.sh#L195

## Your Environment
* CNI plugin version: v3.12.0 (but installed with v3.3.6 manifests, it doesn't work but that's a separate problem)
* Orchestrator version: kubernetes
* CNI config:
* Operating System and version: Fedora CoreOS 31.20200113.3.1

spikecurtis:
Fixes #835 

Nominally, we support `sh` compatibility, not `bash`, but this change is so trivial we might as well for consistency if nothing else.

## Todos
- [x] Tests (no new tests required)
- [x] Documentation (not required)
- [x] Release note (not required)

## Release Note

```release-note
None required
```


spikecurtis:
Actually, there is more going on here.  Turns out this was bugged after all, even for `sh`, and install-cni.sh script isn't actually testing for datastore connectivity.  This needs a bigger PR to fix, including for-real tests.

PR #837, opened 2020-02-12
Install cni bash compat

### LINKED ISSUE #835: install-cni.sh bash syntax error

spikecurtis:
## Description

Fixes #835 - which will make the install script actually wait for the data store connection.

Also fixes up tests so that we are actually testing this functionality.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -192,7 +192,7 @@ sed -i s/__SERVICEACCOUNT_TOKEN__/"${SERVICEACCOUNT_TOKEN:-}"/g $TMP_CONF
 # to the datastore. Once config is in place, Kubelet will start scheduling pods
 # and these will fail if we can't reach the data store for some reason.
 SKIP_DATASTORE_CONNECTION_CHECK=${SKIP_DATASTORE_CONNECTION_CHECK:-""}
-if [-z "${SKIP_DATASTORE_CONNECTION_CHECK}"]; then
+if [ -z "${SKIP_DATASTORE_CONNECTION_CHECK}" ]; then
   until /opt/cni/bin/calico -t < $TMP_CONF
   do
     echo "Failed to reach datastore"; sleep 1;
--- a/k8s-install/scripts/install_cni_test.go
+++ b/k8s-install/scripts/install_cni_test.go
@@ -41,6 +41,8 @@ func runCniContainer(extraArgs ...string) error {
 		"-e", "SLEEP=false",
 		"-e", "KUBERNETES_SERVICE_HOST=127.0.0.1",
 		"-e", "KUBERNETES_SERVICE_PORT=8080",
+		// These tests run without a datastore
+		"-e", "SKIP_DATASTORE_CONNECTION_CHECK=true",
 		"-v", cwd + "/test-templates:/template",
 		"-v", cwd + "/tmp/bin:/host/opt/cni/bin",
 		"-v", cwd + "/tmp/net.d:/host/etc/cni/net.d",
@@ -123,7 +125,7 @@ var _ = Describe("install-cni.sh tests", func() {
 
 	Describe("Run install-cni", func() {
 		Context("With default values", func() {
-			It("Should install bins and config", func() {
+			It("Should install bins and config", func(done Done) {
 				err := runCniContainer()
 				Expect(err).NotTo(HaveOccurred())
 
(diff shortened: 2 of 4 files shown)


spikecurtis:
The spacing causes errors in both `bash` and `sh` --- so the result is that we never hit the code inside that check. I.e. we never actually do the connectivity check.

So, this PR does change current behavior inasmuch as it fixes a bug to actually do the thing we designed it to do.

The changes `k8s-install/scripts/install_cni_test.go` are needed because these tests run without a datastore.  They used to succeed because the spacing bug ensured we never checked the datastore anyway.  I also took the liberty of adding timeouts to the test, so they can't just run forever, if, say, someone accidentally re-enables the datastore check without a datastore.

The other test changes are to actually test the code that connects to the datastore. I had mistakenly thought that the `k8s-install/scripts/install_cni_test.go` tests would do this, but they don't as discussed above.

neiljerram:
@spikecurtis @tmjd I think this PR has caused trouble, because the install-cni container doesn't have any environment that would allow it to connect to the datastore.
- In the KDD case, it's missing `DATASTORE_TYPE=kubernetes`.
- In the etcd case, it's missing `ETCD_ENDPOINTS`.

So for me, trying to bring up an etcd cluster with a manifest that worked last week, it hangs running install-cni.sh like this:
```
neil@glaurung:~$ docker exec -it kind-worker2 bash
root@kind-worker2:/# crictl ps
CONTAINER           IMAGE               CREATED              STATE               NAME                ATTEMPT             POD ID
7cac6b52c2b2d       1809ad5f71faf       About a minute ago   Running             install-cni         0                   fdb3d8fe3012a
cfaf6ad0f8871       2ee81de3993a3       2 minutes ago        Running             kube-proxy          0                   cd143e1d8cc59
root@kind-worker2:/# crictl logs 7cac6b52c2b2d 2>&1 | tail -30
2020-02-17 17:32:53.375 [WARNING][942] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:32:53.375 [ERROR][942] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:32:54.394 [WARNING][955] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:32:54.394 [ERROR][955] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:32:55.430 [WARNING][965] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:32:55.430 [ERROR][965] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:32:56.452 [WARNING][975] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:32:56.452 [ERROR][975] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:32:57.488 [WARNING][987] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:32:57.489 [ERROR][987] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:32:58.507 [WARNING][998] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:32:58.507 [ERROR][998] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:32:59.543 [WARNING][1009] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:32:59.543 [ERROR][1009] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:33:00.583 [WARNING][1020] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:33:00.583 [ERROR][1020] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:33:01.623 [WARNING][1030] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:33:01.623 [ERROR][1030] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
2020-02-17 17:33:02.642 [WARNING][1040] etcdv3.go 69: No etcd endpoints specified in etcdv3 API config
2020-02-17 17:33:02.642 [ERROR][1040] plugin.go 561: data store connection failed error=no etcd endpoints specified
Failed to reach datastore
```

I guess the explanation is that the previous `if [-z "${SKIP...` block never executed at all, because there was no such program as `[-z`.

caseydavenport:
```
In the KDD case, it's missing DATASTORE_TYPE=kubernetes.
In the etcd case, it's missing ETCD_ENDPOINTS.
```

My understanding was that these should be set in the $TMP_CONF that is passed to the CNI plugin, rather than through environment variables. 

IIUC, our manfiests _do_ set these, so it's curious that you're seeing that behavior. Here's where we set it for etcd: https://github.com/projectcalico/calico/blob/master/_includes/charts/calico/templates/calico-node.yaml#L94

The datastore type should be set in the calico-config ConfigMap: https://github.com/projectcalico/calico/blob/master/_includes/charts/calico/templates/calico-config.yaml#L107

neiljerram:
@caseydavenport Thanks; maybe I got the detail wrong.  But the CNI plugin (i.e. the calico binary) was definitely failing to connect to the datastore at this point.

Another factor in the etcd case is that I think we can get a deadlock with the etcd datastore:
- calico-etcd DS exists but does not schedule any pods because it's waiting for nodes to be Ready
- CNI plugin can't connect to datastore, because there aren't yet any etcd pods; therefore nodes do not become Ready.

We can work around that by adding
```
        - key: node.kubernetes.io/not-ready
          effect: NoSchedule
```
to the calico-etcd DS; but that's a further change, so I'd want to be really sure that we need the connection check and that it will work as intended.

caseydavenport:
> calico-etcd DS 

Ah yeah, if a user is trying to install etcd on k8s itself, that might be a problem. But we've long since removed the calico-etcd DS from our docs because it wasn't suitable for production (or even really proof-of-concept deployments) so I don't expect real users to be using it.

Issue #818, opened 2019-12-06
Dataplane interface for the CNI plugin?


AloysAugustin:
Hi Calico maintainers,

We are working on integrating Calico with an alternate dataplane (FD.io VPP). We noticed that there is a dataplane interface for felix but there is no equivalent for the CNI plugin. Is it something that you would be open to adding? This would allow us (and presumably most users of felix's dataplane interface) not to fork the entire CNI plugin.
We have a working prototype and are happy to contribute code if this is something you're interested in.

caseydavenport:
@AloysAugustin that sounds reasonable to me. Before you get started, might be useful to sketch out the changes you're proposing in the next level of detail here in this issue?

AloysAugustin:
Thanks @caseydavenport, sure.
Looking at the code, I think the right point to insert this interface is at the level of the `DoNetworking` / `CleanUpNamespace` pair of functions in [internal/pkg/utils/network_linux.go](https://github.com/projectcalico/cni-plugin/blob/cd887812815fbd260881d046a4574b876335c513/internal/pkg/utils/network_linux.go).
We can define a gRPC service with two rpcs whose arguments and return values match those of these functions. The CNI would then act as a client to this service, performing RPCs each time it is called. 
Regarding the configuration of this feature, I think it should be specified in the network definition, and it could look like this:
```
    "dataplane_options": {
        "type": "external",
        "socket": "unix:///var/run/calico/cni-server.sock"
    }
```
This allows several networks with different dataplanes to coexist, and the socket argument can be passed directly to `grpc.Dial`, allowing the use of unix or tcp sockets for the requests as required. The grpc dial options can be passed through additional parameters in this config section.

Do you have any questions or concerns about this approach?

caseydavenport:
Seems reasonable to me - so we'd keep the existing implementation compiled in and default if no `dataplane_options` are provided?

@fasaxc you might be interested in this as well

AloysAugustin:
> Seems reasonable to me - so we'd keep the existing implementation compiled in and default if no `dataplane_options` are provided?

Yes, that's the idea.
If that sounds fine to you I will clean up my fork, make the .proto file more consistent with the [Felix proto](https://github.com/projectcalico/felix/blob/master/proto/felixbackend.proto) and open a PR here.

caseydavenport:
@AloysAugustin I think that sounds great.

PR #859, opened 2020-03-24
Add dataplane interface and gRPC dataplane

AloysAugustin:
## Description
This PR adds a dataplane interface for the CNI plugin, and an external gRPC dataplane. The dataplane interface has only 2 functions, DoNetworking and CleanupNamespace. The gRPC dataplane simply forwards the arguments to these functions to an external gRPC server.
The dataplane can be configured in the network definition. An example configuration for the gRPC dataplane is:
```
{
      "name": "k8s-pod-network",
      "cniVersion": "0.3.1",
      "plugins": [
        {
          "type": "calico",
          ...
          "dataplane_options": {
            "type": "grpc",
            "socket": "unix:///var/run/calico/cni-server.sock"
          }
        }
    }
```
If no configuration is specified, the default dataplane is the linux dataplane, which implements the current behaviour.

Addresses #818 .

## Todos
- [x] Tests
- [ ] Documentation -> Pointers to where this should be documented are welcome :)
- [ ] Release note

## Release Note

```release-note
Is one required?
```


dataplane/grpc/grpc_dataplane.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return "", "", fmt.Errorf("cannot connect to grpc dataplane: %v", err)
+	}
+
+	c := proto.NewCniDataplaneClient(conn)
+
+	request := &proto.AddRequest{
+		InterfaceName:            args.IfName,
+		Netns:                    args.Netns,
+		DesiredHostInterfaceName: desiredVethName,
```
Brian-McM: The desired desiredVethName could be, and will be if the orchestrator is not k8s, empty. The "linuxDataplane" defaults this to
`hostVethName = "cali" + args.ContainerID[:utils.Min(11, len(args.ContainerID))]`. I suggest factoring that out and using it here as a default.

dataplane/grpc/proto/cnibackend.pb.go
```diff
@@ -0,0 +1,376 @@
+// Code generated by protoc-gen-gogo. DO NOT EDIT.
```
Brian-McM: I think we may want to consider adding the tooling for contributors to regenerate this code, possibly adding this to the standard "make build" target. @caseydavenport, wdyt?

dataplane/grpc/grpc_dataplane.go
```diff
@@ -0,0 +1,127 @@
...
+		Settings: &proto.ContainerSettings{
+			AllowIpForwarding: d.allowIPForwarding,
+			Mtu:               int32(d.mtu),
+		},
+		ContainerIps:    make([]*proto.IPConfig, 0),
+		ContainerRoutes: make([]string, 0),
+	}
+	for _, ipConf := range result.IPs {
```
Brian-McM: You'll see in the linuxDataplane that there is this code
`
// Add a connected route to a dummy next hop so that a default route can be set
if gw == nil {
	gw = net.IPv4(169, 254, 1, 1)
}
`
I don't really know anything about GRPC, but would we still need to add this dummy gateway so we can add the default route?

AloysAugustin: I think this is a dataplane specific detail that comes from the fact that veth are L2 interfaces. For the gRPC dataplane, the actual dataplane (behind the gRPC server) should take care of adding the gw if needed.

dataplane/dataplane.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 Cisco Systems Inc
```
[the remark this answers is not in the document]
AloysAugustin: dataplane moved to pkg/dataplane

dataplane/grpc/grpc_dataplane.go
AloysAugustin: I moved this default name computation to plugin.go to use the same default value for non-k8s orchestrators for all dataplanes. This ensures DoNetworking never receives an empty value in desiredVethName.

I also modified the k8s code to take into account the name returned by DoNetworking, as not all dataplanes may be able to comply with the requested name (this is the case for VPP in particular, where we don't get to choose the name of the interface on the VPP side).

Brian-McM: Good thinking.

dataplane/grpc/grpc_dataplane.go
Brian-McM: Gotcha, thanks for the explanation.

pkg/dataplane/dataplane.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 Cisco Systems Inc
```
Brian-McM: Can you update the Copyright info for the new files to match the existing copy right info.

AloysAugustin: AFAIK the standard procedure with the apache license is to put the name of the copyright holder in the license block (which is Cisco for the files I write, even if they are contributed to a Tigera project). That does not change any of Tigera’s rights regarding this file, and you can add a Tigera line as soon as you modify them. Let me know if you’re ok with that.

caseydavenport:
@AloysAugustin I think I'm good to merge this - OK if I squash commits?

AloysAugustin:
@caseydavenport Sure, please go ahead :)

PR #886, opened 2020-05-23
namespaces support for mesos

vixns:
## Description

Unlike k8s, Apache Mesos does not provide annotations, only labels.
This proposal is to use the `projectcalico.org/namespace` label value as namespace.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note


caseydavenport:
@vixns sorry for the long delay on this!

This looks good to me - I'll add it to the v3.16.0 milestone. Thanks for the PR!

PR #889, opened 2020-06-02
k8s-install: install-cni.sh now aborts when calico -v doesn't succeed

### LINKED ISSUE #877: initContainer succeeds even though copy of calico binary fails
## Expected Behavior
If the calico binary cannot be copied, I expect the install-cni.sh sript to fail

## Current Behavior
We're running very frequent installations and in 99% of the cases everything works fine. I just had the case that the binary was not copied but install-cni.sh succeeded anyway so no retry was triggered. Output:
```
/install-cni.sh: line 96: /host/opt/cni/bin/calico: No such file or directory
CNI plugin version: 
```

## Steps to Reproduce (for bugs)
I'm not sure if it's possible to reproduce this.
```bash
for path in /opt/cni/bin/*;
  do
    filename="$(basename "$path")"
    tmp=",$filename,"
    if [ "${SKIP_CNI_BINARIES#*$tmp}" != "$SKIP_CNI_BINARIES" ];
    then
      echo "$filename is in SKIP_CNI_BINARIES, skipping"
      continue
    fi
    if [ "${UPDATE_CNI_BINARIES}" != "true" ] && [ -f $dir/"$filename" ];
    then
      echo "$dir/$filename is already here and UPDATE_CNI_BINARIES isn't true, skipping"
      continue
    fi
    cp "$path" $dir/ || exit_with_error "Failed to copy $path to $dir. This may be caused by selinux configuration on the host, or something else."
  done

  echo "Wrote Calico CNI binaries to $dir"
  echo "CNI plugin version: $($dir/calico -v)"
```

It's unclear to me how the `cp` could succeed but it would be nice if `$($dir/calico -v)` would fail if 
the file cannot be found (`/install-cni.sh: line 96: /host/opt/cni/bin/calico: No such file or directory`). I guess it doesn't fail because it's executed in a subshell). Would it be possible to implement it in a way that the whole script fails if `calico -v` doesn't work?

## Your Environment

* CNI plugin version (e.g. `/opt/cni/bin/calico --version`): 3.13
* Orchestrator version (e.g. kubernetes, mesos, rkt): Kubernetes 1.17
* Operating System and version: Ubuntu

sbueringer:
## Description
Fixes #877

Apparently echo masks the error. Can be reproduced with:

```bash
#!/usr/bin/env bash

set -e

echo "false in echo does not abort $(false)"

output=$(false)
echo "false in variable assignment does ${output}"
```

```release-note
install-cni.sh now also fails if `calico -v` doesn't work after copying the calico binary
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -97,7 +97,8 @@ do
   done
 
   echo "Wrote Calico CNI binaries to $dir"
-  echo "CNI plugin version: $($dir/calico -v)"
+  calico_version=$(${dir}/calico -v)
+  echo "CNI plugin version: ${calico_version}"
 done
 
 TMP_CONF='/calico.conf.tmp'


sbueringer:
@jperville I also added release notes above. Let me know if there's anything else I can/should do :)

jperville:
Perfect, thank you very much @sbueringer

PR #902, opened 2020-06-24
Fix vulnerabilities

gianlucam76:
```release-note
None required
```

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1-407
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2-301.1592810506 AS ubi
 
 ARG GIT_VERSION=unknown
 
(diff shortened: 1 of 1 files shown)


Dockerfile.amd64
```diff
@@ -31,6 +31,38 @@ ADD k8s-install/scripts/calico.conf.default /calico.conf.tmp
 RUN microdnf install hostname && mkdir /licenses
 COPY LICENSE /licenses
 
+RUN microdnf update
+
+# Remove packages currently known to have vulnerabilities and dependencies
+# Remove packages deemed known to be unsafe by thread defense team
```
doublek, line 37: nit: s/thread/threat.

Dockerfile.amd64
```diff
@@ -31,6 +31,38 @@ ADD k8s-install/scripts/calico.conf.default /calico.conf.tmp
...
+    libidn2 \
+    libpsl \
+    libnghttp2 \
+    libcrypto  glibc-langpack-en glibc-minimal-langpack  json-c libcom_err \
+    libgcc libstdc++ libstdc libpcap libxml2 lz4-libs openssl platform-python-pip python3-pip-wheel systemd sqlite-libs \
+    libuuid \
+    libblkid libmount \
+    libsmartcols
+
+FROM scratch
```
bcreane, line 63: please comment on motivation, e.g.
```
# Copy everything into a fresh scratch image so that naive CVE scanners don't pick up binaries and libraries
# that have been removed in our later layers.
```

caseydavenport, line 63: Could you add a comment explaining this?

Something like this: "Image CVE scanners will sometimes incorrectly report CVEs across the union of base image layers, even if the affected software has been removed. Copy to a fresh scratch image so that binaries and libraries that have been removed are not incorrectly included in CVE scanes"

Dockerfile.amd64
```diff
@@ -31,6 +31,38 @@ ADD k8s-install/scripts/calico.conf.default /calico.conf.tmp
 RUN microdnf install hostname && mkdir /licenses
 COPY LICENSE /licenses
 
+RUN microdnf update
+
+# Remove packages currently known to have vulnerabilities and dependencies
+# Remove packages deemed known to be unsafe by threat defense team
+RUN microdnf remove cyrus-sasl-lib \
```
doublek, line 38: Does the order matter here? I saw someplace else that it mattered. Curious if this is the case here.

gianlucam76: It does not appear to matter. microdnf seems to be smart enough to understand (or sort itself based on dependencies).

Dockerfile.amd64
gianlucam76: I don't know how whitesource works to be honest. Doing this seems to cause issue and image scan fails. So I am going to remove this.

Issue #874, opened 2020-05-09
mv requires care on SELinux enforcing systems


dghubble:
`install-cni.sh` uses `mv` to place the CNI configs. Use of `mv` affects the way SELinux context's are set for placed configs and binaries.

Related: With https://github.com/projectcalico/cni-plugin/pull/849 (v3.14.0), `install-cni.sh` switched from placing CNI binaries with `cp` to using `mv` which may cause related issues, depending on the setup.

## Current Behavior

Using `mv`, the SELinux context of moved files is kept as-is when its moved from temporary to final locations (mounted from host). That means the SELinux file context is determined when the temp file is written, which depends on the SELinux context the `calico/cni` container is run with.

With `cp`, the SELinux context of the target directory or host transition rules would determine the SELinux context of palced binary and conf files. 

Best described in https://danwalsh.livejournal.com/56534.html

## Expected Behavior

`calico/node` may run multiple times (due to pod restarts, anything, etc.) and must continue to be able to manage/write the CNI config and CNI binaries its responsible for. To do that, keeping a stable context across restarts is important ("shared" in container terms).

## Steps to Reproduce (for bugs)

On an SELinux enforicing system, I find this can be observed most readily in the CNI config file, if you restart a Calico pod (but not reboot the instance, as Kubelet service is often setup to do some relabling that corrects the issue).

```
mv: inter-device move failed: '/calico.conf.tmp' to '/host/etc/cni/net.d/10-calico.conflist'; unable to remove target: Permission denied
Failed to mv files. This may be caused by selinux configuration on the host, or something else.
```

Here, the host SELinux config isn't to blame imo. The first Calico pod runs install-cni as an init container and the written file carries the SELinux context of that container (random MCS pair). When Calico is restarted, the container will have a different pair and be unable to access the file (install-cni tries to write it every time) which causes the init container to fail over and over.

```
-rw-r--r--. 1 root root system_u:object_r:container_file_t:s0:c443,c515  646 May  9 22:07 10-calico.conflist 
-rw-------. 1 root root system_u:object_r:container_file_t:s0           2786 May  9 22:07 calico-kubeconfig
```

## Possible Solution

[...]

PR #910, opened 2020-07-09
Restore binary SELinux context when possible

realgaurav:
## Description
Restore binary SELinux context when available during installation.

## Todos
- [ ] Testing

## Release Note

```release-note
Restore SELinux context when possible.
```

### DIFF
--- a/k8s-install/scripts/install-cni.sh
+++ b/k8s-install/scripts/install-cni.sh
@@ -91,9 +91,18 @@ do
     fi
     # Copy the binary to a temporary location, and then mv it into place. This safely
     # distributes the binary even if the existing executable is currently running.
-    rm -f $dir/$filename.tmp
-    cp $path $dir/$filename.tmp
-    mv $dir/$filename.tmp $dir/$filename || exit_with_error "Failed to copy $dir/$filename.tmp to $dir/$filename. This may be caused by selinux configuration on the host, or something else."
+    cp -f $path $dir/$filename.tmp
+    # Adjust security context of temporary file so the move operation below can
+    # delete temporary file without violation.
+    if [ -e "/sbin/restorecon" ]; then
+        /sbin/restorecon $dir/$filename.tmp
+    fi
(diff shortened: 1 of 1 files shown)


k8s-install/scripts/install-cni.sh
```diff
@@ -91,9 +91,18 @@ do
     fi
     # Copy the binary to a temporary location, and then mv it into place. This safely
     # distributes the binary even if the existing executable is currently running.
-    rm -f $dir/$filename.tmp
-    cp $path $dir/$filename.tmp
-    mv $dir/$filename.tmp $dir/$filename || exit_with_error "Failed to copy $dir/$filename.tmp to $dir/$filename. This may be caused by selinux configuration on the host, or something else."
+    cp -f $path $dir/$filename.tmp
+    # Adjust security context of temporary file so the move operation below can
+    # delete temporary file without violation.
+    if [ -e "/sbin/restorecon" ]; then
```
caseydavenport, line 97: Do we need to worry about this binary existing in different locations on different platforms?

realgaurav: Good point. I took a quick look earlier and saw many examples pointing to `/sbin`.
Now I looked at package search for rhel/centos/fedora/ubuntu etc. (e.g. https://www.rpmfind.net/linux/rpm2html/, ) and noticed `/usr/sbin` and `/sbin` as only possible locations. I have updated the PR with these 2 path details.

Let me know if there is a better way to handle it.

dghubble:
By itself, this didn't change the effect of a calico-node Pod restart causing it to be in a CrashLoop - the `mv` will exit with an error (without the solution used in https://github.com/projectcalico/cni-plugin/issues/874). This wants users to mount the host's `/usr/sbin` into the init container at `/usr/sbin`? I think you'd want to expect the host mount at `/host/sbin/restorecon`.

Then the question is whether a pod running under a new MCS pair has the ability to call restorecon. Not sure.

PR #913, opened 2020-07-14
Handle panics in the CNI plugin gracefully

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Handle panics in the CNI plugin more gracefully
```

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -1,4 +1,4 @@
-// Copyright 2015 Tigera Inc
+// Copyright 2015,2020 Tigera Inc
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pkg/plugin/plugin.go
```diff
@@ -447,57 +473,77 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 		}
 		logrus.Debug("/var/lib/calico/nodename exists")
 	}
 
 	// Determine which node name to use.
 	nodename := utils.DetermineNodename(conf)
 
-	epIDs, err := utils.GetIdentifiers(args, nodename)
+	var epIDs *utils.WEPIdentifiers
+	epIDs, err = utils.GetIdentifiers(args, nodename)
```
lmm, line 517: Do we need to change the "short declaration" syntax here? I thought that Go would redeclare existing vars (`err`) in this case and not overwrite them but I'm not sure.

pkg/plugin/plugin.go
```diff
@@ -39,6 +39,7 @@ import (
 	"github.com/projectcalico/cni-plugin/pkg/k8s"
```
lmm, line 39: Could you update the license header?

pkg/plugin/plugin.go
lmm: I guess the long form is more obviously correct; in that case, maybe we can do that same in cmdAdd too

PR #914, opened 2020-07-15
Move to scratch

gianlucam76:


Dockerfile.amd64
```diff
@@ -12,7 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1-407
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2-301.1592810506 AS base
+
+RUN mkdir /licenses
+ADD licenses/ /licenses
+ADD LICENSE /licenses/
+RUN mkdir -p /opt/cni/bin
```
robbrockbank, line 20: Couple of questions:

- Do we need the mkdir `/opt/cni/bin` - we don't have this for the other dockerfiles below, so seems odd that we would need it here.  If we do need it here, could we just add the binary to / instead?
-  I'm still trying to understand why we need the Ubi base image.  AFAICT it's really just for the license. Is that correct?  Is there something special about that copy of the license file?

Couldn't we rationalize all of these docker files to be from scratch, copy in a license file from the local file system and then just copy across the appropriate build of the binary.  That way all of the docker files look pretty similar.

Dockerfile.arm64
```diff
@@ -1,14 +1,12 @@
-FROM arm64v8/debian:9.8-slim
-
-LABEL maintainer "Casey Davenport <casey@tigera.io>"
+FROM scratch
 
 ADD licenses/ /licenses
```
robbrockbank, line 3: this won't work will it?

Dockerfile.amd64
gianlucam76: We do need to create /opt/cni/bin. The copy later on in scratch (ADD bin/amd64/ /opt/cni/bin/) will fail otherwise.
Regarding UBI, I am not the best to answer. It was there and I left it there.

lmm: While you're here, could you please change the maintainer label to `maintainers@projectcalico.org`

robbrockbank: Oh I see, it's because of the multistage build.  Gotcha.  I'm happy to go with the status quo for this, but my personal preference would be to copy in license from local file if that's really all we are using that base image for.  I don't think you should change this unless there was general agreement on that approach though.

Dockerfile.arm64
robbrockbank: Oh d'oh - sorry this IS a local copy.  My bad!

robbrockbank: So in that case we seem to have a mixture of using local license file and copying in from ubi.  I really think we should at least converge on one method otherwise it raises questions about why they are different.

gianlucam76: I pinged @lmm  regarding this. I locally verified cluster comes up with that. Once @lmm  acks RH validation is still OK with such image, I will update this PR.

pkg/install/install_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+		// Create token file for the Kubernetes client.
+		k8sSecret := []byte("my-secret-key")
+		var tokenFile = fmt.Sprintf("%s/serviceaccount/token", tempDir)
+		err = ioutil.WriteFile(tokenFile, k8sSecret, 0755)
+		if err != nil {
+			Fail(fmt.Sprintf("Failed to write k8s secret file: %v", err))
+		}
+
+		// Create a ca.crt - requires to the in cluster config works correctly. Since this is just test code,
```
robbrockbank, line 127: nit:  to->so

pkg/install/install.go
```diff
@@ -0,0 +1,467 @@
...
+	// Directory where we expect that TLS assets will be mounted into the calico/cni container.
+	TLSAssetsDir string `envconfig:"TLS_ASSETS_DIR"`
+
+	// SkipCNIBinaries is a comma-separated list of binaries. Each binary in the list
+	// will be skipped when installing to the host.
+	SkipCNIBinaries string `envconfig:"SKIP_CNI_BINARIES"`
+
+	// UpdateCNIBinaries controls whether or not to overwrite any binaries with the same name
+	// on the host.
+	UpdateCNIBinaries bool
```
robbrockbank, line 36: nit:  Perhaps OverwriteHostBinaries is more accurate.  Does this need to be env controlled do you think?

pkg/install/install.go
```diff
@@ -0,0 +1,467 @@
...
+
+	// The CNI network configuration to install.
+	CNINetworkConfig     string `envconfig:"CNI_NETWORK_CONFIG"`
+	CNINetworkConfigFile string `envconfig:"CNI_NETWORK_CONFIG_FILE"`
+
+	ServiceAccountToken []byte
+}
+
+func (c config) skipBinary(binary string) bool {
+	return strings.Contains(c.SkipCNIBinaries, binary)
```
robbrockbank, line 46: Should really split by comma and then compare. Probably doesn't matter too much, but as it stands "my-binary" would match "not-my-binary"

pkg/install/install.go
```diff
@@ -0,0 +1,467 @@
...
+func fileExists(file string) bool {
+	info, err := os.Stat(file)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !info.IsDir()
+}
+
+func mkdir(path string) {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
robbrockbank, line 73: This check isn't necessary, Mkdir is a no-op if it already exists.  It also errors if path is not a dir.

pkg/install/install.go
gianlucam76: Good catch.Yes it is currently controlled by environment variable in shell script. Will fix it. Regarding name, maintained the same name used now UPDATE_CNI_BINARIES

Dockerfile.amd64
gianlucam76: @robbrockbank  your suggestion is good. Changed this to full scratch. I verified cluster comes up just fine and @lmm verified RH validation is OK with that.

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+		if err := os.Remove(fmt.Sprintf("/host/etcd/cni/net.d/%s", oldName)); err != nil {
+			logrus.WithError(err).Warnf("Failed to remove %s", oldName)
+		}
+	}
+}
+
+// copyFile copies a file from src to dst. If src and dst files exist, and are
+// the same, then return success. Otherise, attempt to create a hard link
+// between the two files. If that fail, copy the file contents from src to dst.
+func copyFile(src, dst string) (err error) {
```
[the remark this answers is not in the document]
robbrockbank: I realize you are setting the file permissions as well, but presumably we could do that after the copy?

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+		// version to the host. Just check the timestamp on the certificate to see if it
+		// has been updated.  A bit hokey, but likely good enough.
+		filename := c.TLSAssetsDir + "/etcd-cert"
+		info, err := os.Stat(filename)
+		if err == nil {
+			statOutput := info.ModTime()
+			time.Sleep(10 * time.Second)
+			info, err = os.Stat(filename)
+			if err != nil {
+				continue
```
robbrockbank, line 223: Shouldn't this fail if we hit an error?  A failure here would be silent AFAICT

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+	// This prevents Kubernetes from restarting the pod repeatedly.
+	shouldSleep := getEnv("SLEEP", "true")
+	logrus.Infof("Done configuring CNI.  Sleep= %v", shouldSleep)
+	for shouldSleep == "true" {
+		// Kubernetes Secrets can be updated.  If so, we need to install the updated
+		// version to the host. Just check the timestamp on the certificate to see if it
+		// has been updated.  A bit hokey, but likely good enough.
+		filename := c.TLSAssetsDir + "/etcd-cert"
+		info, err := os.Stat(filename)
+		if err == nil {
```
robbrockbank, line 218: Should we be handling specific errors (namely not exists?)

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+		c.TLSAssetsDir = "/calico-secrets"
+	}
+
+	return c
+}
+
+func Install() error {
+	// Clean up any existing binaries / config / assets.
+	if err := os.Remove("/host/opt/cni/bin/calico"); err != nil {
+		logrus.WithError(err).Warnf("Error removing old plugin")
```
robbrockbank, line 102: We shouldn't write a warning log if the file doesn't actually exist.

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	// Write a CNI config file.
+	writeCNIConfig(c)
+
+	// Unless told otherwise, sleep forever.
+	// This prevents Kubernetes from restarting the pod repeatedly.
+	shouldSleep := getEnv("SLEEP", "true")
+	logrus.Infof("Done configuring CNI.  Sleep= %v", shouldSleep)
+	for shouldSleep == "true" {
```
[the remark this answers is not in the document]
robbrockbank: Could also replace this loop using something like fs-notify, there is a golang implementation of that, so you can watch the contents of a directory for changes.

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+	dirs := []string{"/host/opt/cni/bin", "/host/secondary-bin-dir"}
+	for _, d := range dirs {
+		if unix.Access(d, unix.W_OK) != nil {
+			logrus.Infof("%s is not writeable, skipping", d)
+			continue
+		}
+
+		// Iterate through each binary we might want to install.
+		// TODO: Include all the binaries we ship.
+		binaries := []string{"calico", "calico-ipam", "bandwidth", "flannel", "loopback", "host-local", "portmap", "tuning"}
```
robbrockbank, line 163: This deviates from the original script - in that the script just lists all files in the bin directory.  Any particular reason we don't do that here?

pkg/install/install.go
gianlucam76: Changed to if file does not exist then sleep, else do logic

pkg/install/install.go
gianlucam76: I added a Warn. I am not sure about failing. If I need to fail, should I simply leave the loop? Would that be OK? I though leaving the loop would be better.

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+	return def
+}
+
+func directoryExists(dir string) bool {
+	info, err := os.Stat(dir)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return info.IsDir()
```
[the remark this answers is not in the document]
gianlucam76: The IsNotExist, IsExist, IsPermission and IsTimeout functions in package os all return false when passed a nil error. Also tested locally to be sure.

pkg/install/install.go
gianlucam76: Changed to go over all files in  /opt/cni/bin/

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+	// Location on the host where CNI network configs are stored.
+	CNINetDir   string `envconfig:"CNI_NET_DIR" default:"/etc/cni/net.d"`
+	CNIConfName string `envconfig:"CNI_CONF_NAME"`
+
+	// Directory where we expect that TLS assets will be mounted into the calico/cni container.
+	TLSAssetsDir string `envconfig:"TLS_ASSETS_DIR"`
+
+	// SkipCNIBinaries is a comma-separated list of binaries. Each binary in the list
+	// will be skipped when installing to the host.
+	SkipCNIBinaries string `envconfig:"SKIP_CNI_BINARIES"`
```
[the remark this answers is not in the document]
gianlucam76: Added ShouldSleep to config struct and changed to bool.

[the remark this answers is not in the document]
caseydavenport: FWIW, the SLEEP env var isn't used any more. But happy to keep it included

Makefile
```diff
@@ -71,18 +71,34 @@ update-pins: update-libcalico-pin
...
 	$(MAKE) build ARCH=$*
 
+$(BIN)/calico-ipam: $(BIN)/install
+	cp $(BIN)/install $(BIN)/calico-ipam
+	
+$(BIN)/calico: $(BIN)/install
+	cp $(BIN)/install $(BIN)/calico
+	
+## Build the Calico network plugin and ipam plugins
+$(BIN)/install binary: $(SRC_FILES)
```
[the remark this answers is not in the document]
caseydavenport: Just means there are two ways to call the same target:

```
make bin/install
```

or 

```
make binary`
```

cmd/calico/calico.go
```diff
@@ -15,12 +15,31 @@
...
 	"github.com/projectcalico/cni-plugin/pkg/plugin"
 )
 
 // VERSION is filled out during the build process (using git describe output)
 var VERSION string
 
 func main() {
-	plugin.Main(VERSION)
+	// Use the name of the binary to determine which routine to run.
+	_, filename := path.Split(os.Args[0])
```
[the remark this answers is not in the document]
caseydavenport: We should only have a single copy of the binary in the image.

IIRC my intention was that the install script would _copy itself_ into the other binaries on the host (calico, calico-ipam) as needed.

pkg/install/install.go
```diff
@@ -0,0 +1,464 @@
...
+			}
+		} else {
+			time.Sleep(10 * time.Second)
+		}
+	}
+	return nil
+}
+
+func writeCNIConfig(c config) {
+	netconf := `{
```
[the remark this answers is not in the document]
caseydavenport: Yeah, that's our thing - users might be passing in all sorts of modifications to the default CNI config "template" as well, so sorta hard to change that right now.

Dockerfile.arm64
```diff
@@ -1,14 +1,12 @@
...
-ADD k8s-install/scripts/install-cni.sh /install-cni.sh
-ADD k8s-install/scripts/calico.conf.default /calico.conf.tmp
+LABEL maintainer "Casey Davenport <casey@tigera.io>"
+
+ADD bin/arm64 /opt/cni/bin/
 
 ENV PATH=$PATH:/opt/cni/bin
 WORKDIR /opt/cni/bin
-CMD ["/opt/cni/bin/calico"]
+CMD ["/opt/cni/bin/calico"]
```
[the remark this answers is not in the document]
caseydavenport: hm, I think it's just so that you could do `-v` and get the version easily. No need to keep it this.

Dockerfile.arm64
```diff
@@ -1,14 +1,12 @@
...
-LABEL maintainer "Casey Davenport <casey@tigera.io>"
+FROM scratch
 
 ADD licenses/ /licenses
 ADD LICENSE /licenses/
 
-ADD bin/arm64/ /opt/cni/bin/
-ADD k8s-install/scripts/install-cni.sh /install-cni.sh
-ADD k8s-install/scripts/calico.conf.default /calico.conf.tmp
+LABEL maintainer "Casey Davenport <casey@tigera.io>"
```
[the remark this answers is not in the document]
gianlucam76: This was already ready. Anyhow, I will change maintainer to be same in all Dockerfile.

cmd/calico/calico.go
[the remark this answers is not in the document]
gianlucam76: Is the intent then to change Makefile (from build: $(BIN)/install $(BIN)/calico $(BIN)/calico-ipam to build: $(BIN)/install) and then have install.go copy itself to calico-ipam calico just before iteration on all binaries so we can save ~50MB in the image?

Makefile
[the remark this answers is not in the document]
robbrockbank: TIL .... Thanks!

cmd/calico/calico.go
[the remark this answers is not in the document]
gianlucam76: image does not contain calico and calico-ipam anymore. At run time, install copies itself.

pkg/install/install.go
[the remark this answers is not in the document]
gianlucam76: Added a wrapper function which first invokes go-cp.CopyFile then copies file permissions.

Dockerfile.arm64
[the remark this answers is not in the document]
gianlucam76: Removed it.

Dockerfile.s390x
```diff
@@ -4,11 +4,14 @@ LABEL maintainer "LoZ Open Source Ecosystem (https://www.ibm.com/developerworks/
 
 ADD licenses/ /licenses
 ADD LICENSE /licenses/
+RUN mkdir -p /opt/cni/bin
+
+FROM scratch
+
+COPY --from=base /licenses /licenses
```
robbrockbank, line 11: Do we need this because it is a different license for s390?

gianlucam76: Good catch. We don't. I will update this

gianlucam76: It was there. I just moved around but left where it was.

FROM s390x/debian:9.8-slim

...
ADD licenses/ /licenses
ADD LICENSE /licenses/

...

neiljerram:
@caseydavenport @gianlucam76 @robbrockbank Do you know why we started respecting the `NODENAME` env var for substituting `__KUBERNETES_NODE_NAME__` in the CNI config?  I'm asking for a user report that behaviour seems to have changed from 3.15.2 to 3.16.0, and my own analysis so far indicates we made two changes:

1. It was `$(hostname)` in shell but now golang `os.Hostname()`
2. We also now respect an env var `NODENAME`, if that is defined.

Why (2)?

gianlucam76:
Thanks @neiljerram. That unfortunately appears to be a change I introduced with my commit (honoring first NODENAME). I see I went from

sed -i s/__KUBERNETES_NODE_NAME__/"${KUBERNETES_NODE_NAME:-$(hostname)}"/g $TMP_CONF

to
netconf = strings.Replace(netconf, "__KUBERNETES_NODE_NAME__", getEnv("NODENAME", nodename), -1)

So we first honor env var NODENAME if defined and go

If you guys want, I can push a PR and change it back to
 
netconf = strings.Replace(netconf, "__KUBERNETES_NODE_NAME__", nodename, -1)

nodename is already set to nodename, err := names.Hostname()

Regarding os.Hostname vs shell hostname, does that give any difference?

neiljerram:
Thanks @gianlucam76 !  So IIUC, it was an accidental change from honouring **KUBERNETES_NODE_NAME** to honouring **NODE_NAME** ?  If that's right, I'm happy to shepherd the PR to fix that.

> Regarding os.Hostname vs shell hostname, does that give any difference?

No, I don't think so.

neiljerram:
Thanks @gianlucam76 .  I've raised https://github.com/projectcalico/cni-plugin/pull/942 to correct this detail.

PR #915, opened 2020-07-16
CVEs

gianlucam76:
Leave only what's needed by install-cni shell script.

This is only temporary fix. The long term fix for master is to convert script to go binary.

### DIFF
--- a/Dockerfile.amd64
+++ b/Dockerfile.amd64
@@ -12,7 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1-407
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2-301.1592810506 AS base
+
+COPY k8s-install/scripts/clean.sh /clean.sh
+
+RUN touch /in-the-container
+
+RUN microdnf install hostname
+
+RUN mkdir /licenses
+ADD licenses/ /licenses
+ADD LICENSE /licenses/
+
+RUN microdnf update
+
+SHELL ["/bin/bash", "-c"]
+RUN  /clean.sh "/bin/sed" "/bin/mkdir" "/bin/cp" "/bin/mv" "/bin/cat" "/bin/touch" "/bin/chmod" \
+"/bin/grep"  "/bin/hostname" "/bin/basename" "/bin/base64"
+
+FROM scratch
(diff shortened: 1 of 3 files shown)


Dockerfile.amd64
```diff
@@ -12,7 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1-407
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.2-301.1592810506 AS base
+
+COPY k8s-install/scripts/clean.sh /clean.sh
+
+RUN touch /in-the-container
```
robbrockbank, line 19: What's the purpose of these various RUN parms?  Is it just to check the executables can run?  Couldn't we do that in some container verification post build?

k8s-install/scripts/clean.sh
```diff
@@ -0,0 +1,132 @@
...
+
+# Recursively find all that is needed
+loop=1
+while [ "$loop" -eq 1 ]; do 
+    loop=0
+    newDependencies=()
+    for library in "${libraries[@]}"; do
+        rlink=$(readlink "$library")
+        for l in $rlink; do
+            if [[ ! " ${libraries[@]} " =~ " /lib64/${l} " ]]; then
```
robbrockbank, line 96: I don't really grok the /lib64 vs /use/lib64 thing.  How come we don't need to worry about /usr/lib64 here?

k8s-install/scripts/clean.sh
```diff
@@ -0,0 +1,132 @@
+#!/bin/bash
```
robbrockbank, line 1: This script looks pretty neat.  Do you think this will likely be a more commonly used script - if so, I think it could benefit from beefing up the comments a little - especially in a lot of the nested routines.  The code itself looks fine, but a few more comments would go a long way.  That said, it's probably because my shell scripting is a bit rusty that I'm so slow reviewing  :-)

k8s-install/scripts/clean.sh
gianlucam76: One is a symlink of the other. But you are right, I should not count on that. I will fix this.

k8s-install/scripts/clean.sh
gianlucam76: Thanks. I think long term, script will disappear. But I am going to add more comments anyway.

PR #904, opened 2020-06-26
Add validate cni.conf to be valid json

johscheuer:
## Description

Fixes: https://github.com/projectcalico/cni-plugin/issues/903

This PR add jq and checks if the `cni.conf` is a valid json document.

The script will only run jq (for validation) if it's installed.

## Todos
- [ ] Tests
- [X] Documentation (I think we don't need some additional docs here)
- [X] Release note

## Release Note

```release-note
install-cni will now check if the cni.conf file is a valid json document
```

PR #918, opened 2020-07-22
Reintroduce CMD in Dockerfile

gianlucam76:
Needed to fix node build


Dockerfile.s390x
```diff
@@ -14,4 +14,5 @@ COPY --from=base /opt/cni/bin /opt/cni/bin
 ADD bin/s390x/ /opt/cni/bin/
 
 ENV PATH=$PATH:/opt/cni/bin
-WORKDIR /opt/cni/bin]
\ No newline at end of file
+WORKDIR /opt/cni/bin]
```
johscheuer, line 17: This is still broken see: https://github.com/projectcalico/cni-plugin/pull/904#issuecomment-661754566

PR #922, opened 2020-07-29
fix dualstack support for floating ips

fionera:
## Description
As stated in https://github.com/projectcalico/calico/issues/3838, currently calico throws an error if you try to use dualstack with floatingIPs because the InternalIP isnt mapped properly. This PR should fix it by using the same logic as in libcalico 
https://github.com/projectcalico/libcalico-go/blob/master/lib/backend/k8s/conversion/workload_endpoint_default.go#L162:L176

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note
```release-note
None required
```


fionera:
Sure will do :)

fionera:
@neiljerram is your test enough or should I add more?

neiljerram:
@fionera Sorry not to be clear.  I meant that just as a start, and as a way of indicating where would be a good place to test this.  But as the CI shows, it doesn't pass like this because the expectations of the following lines need adjusting.

neiljerram:
To be more precise, I think it just needs this adding into the WorkloadEndpointSpec at line 1915:
```
				IPNATs: []api.IPNAT{
					{
						InternalIP: "20.0.0.0",
						ExternalIP: "1.1.1.1",
					},
					{
						InternalIP: "fd80:20::45a7:85c0",
						ExternalIP: "2001:647f::21",
					},
				},
```

fionera:
Yeah I didnt test it locally since I couldnt get it to compile. I added the change tho and will try to fix it with the CI results

fionera:
@neiljerram I fixed the test now :) Anything else needed?

PR #924, opened 2020-08-06
Added Windows support

song-jiang:
## Description

This PR added Calico CNI, Calico CNI IPAM plugin on Windows nodes.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


.semaphore/semaphore.yml
```diff
@@ -0,0 +1,79 @@
...
+promotions:
+- name: Cleanup
+  pipeline_file: cleanup.yml
+  auto_promote:
+    when: "result = 'stopped'"
+
+global_job_config:
+  secrets:
+  # Key for pulling from private github repos.
+  - name: private-repo
```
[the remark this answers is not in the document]
song-jiang: We have some Windows CI scripts for WinFV which is in private repo for now.

pkg/k8s/k8s.go
```diff
@@ -59,11 +61,78 @@ func CmdAddK8s(ctx context.Context, args *skel.CmdArgs, conf types.NetConf, epID
 		"Namespace":        epIDs.Namespace,
 	})
 
+	d, err := dataplane.GetDataplane(conf, logger)
+	if err != nil {
+		return nil, err
+	}
+
 	logger.Info("Extracted identifiers for CmdAddK8s")
+	if runtime.GOOS == "windows" {
```
fasaxc, line 70: Rather than stubbing out several functions, why not replace this entire if block with `err := utils.CheckForSpuriousAdd()` and then have a since Windows/Linux-specific function?  I think that'd be a lot cleaner.

song-jiang: Will do it with another PR.

PR #927, opened 2020-08-12
File logging

mgleung:
## Description

Adds the option to configure the CNI plugin to output its logs to a file as well as `stderr`. This sets the default log location to be `/var/log/calico/cni/cni.log` with the log files rotated every 100 MB. 

This also cleans up some log messages that may be more useful in a debug capacity rather than an info one.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Add configurable file logging.
```


internal/pkg/utils/utils.go
```diff
@@ -696,7 +699,27 @@ func ConfigureLogging(logLevel string) {
...
+	// Set the log output to write to a log file if specified.
+	if logFilePath != "" {
+		// Create the path for the log file if it does not exist
+		err := os.MkdirAll(filepath.Dir(logFilePath), 0755)
+		if err != nil {
+			logrus.WithError(err).Errorf("Failed to create path for CNI log file: %v", filepath.Dir(logFilePath))
+		}
+
+		// Create file logger with log file rotation. Defaults to rotate once files hit 100 MB.
+		fileLogger := &lumberjack.Logger{
```
caseydavenport, line 712: Should we be setting a MaxBackups or MaxAge in this logger so that we don't built up log files indefinitely?

pkg/plugin/plugin.go
```diff
@@ -123,7 +123,7 @@ func cmdAdd(args *skel.CmdArgs) (err error) {
 		return fmt.Errorf("failed to load netconf: %v", err)
 	}
 
-	utils.ConfigureLogging(conf.LogLevel)
+	utils.ConfigureLogging(conf.LogLevel, conf.LogFilePath)
```
caseydavenport, line 126: One consequence of this change is that all of the logs will be intertwined in the same log file (e.g. if multiple CNI plugins are running in parallel). 

Can you check to see if we are including the container identifiers in _every log line_ produced by the plugin?

e.g., containerID, interface name, networkName.

internal/pkg/utils/utils.go
mgleung: Sure, I didn't add anything since I wasn't sure what a reasonable default might be, but I'll add something and make it configurable so it can be changed if needed.

pkg/plugin/plugin.go
mgleung: We include either a container ID or a handle ID in most of the logs. From what I have found, the logs that do not include that information don't seem to need it. ex:
```
Attempting to load block cidr=192.168.92.0/26 host="matt-bz-0j96-kadm-node-1"

PR #929, opened 2020-08-17
Fill in resource version and UID when doing a delete.

fasaxc:
## Description

Make sure that we delete the correct resource when cleaning up WEPs.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/k8s/k8s.go
+++ b/pkg/k8s/k8s.go
@@ -33,15 +33,16 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 
-	"github.com/projectcalico/cni-plugin/internal/pkg/utils"
-	"github.com/projectcalico/cni-plugin/pkg/dataplane"
-	"github.com/projectcalico/cni-plugin/pkg/types"
 	api "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	k8sconversion "github.com/projectcalico/libcalico-go/lib/backend/k8s/conversion"
 	calicoclient "github.com/projectcalico/libcalico-go/lib/clientv3"
 	cerrors "github.com/projectcalico/libcalico-go/lib/errors"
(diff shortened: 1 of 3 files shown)


pkg/k8s/k8s.go
```diff
@@ -475,44 +476,57 @@ func CmdDelK8s(ctx context.Context, c calicoclient.Interface, epIDs utils.WEPIde
...
+				UID:             &wep.UID,
+			},
+		); err != nil {
+			// Delete the WorkloadEndpoint object from the datastore, passing revision information from the
+			// queried resource above in order to prevent conflicts.
+			switch err := err.(type) {
+			case cerrors.ErrorResourceDoesNotExist:
+				// Log and proceed with the clean up if WEP doesn't exist.
+				logger.WithField("endpoint", wep).Info("Endpoint object does not exist, no need to clean up.")
+			case cerrors.ErrorResourceUpdateConflict:
```
caseydavenport, line 513: If we hit a UID conflict here do we actually want to retry?

e.g.,

- We do a GET
- pod is deleted / recreated with new UID
- We get conflict
- We do a GET (now retrieving the new UID)
- We properly delete the WEP this time (removing any IP addresses, which might actually be valid now?)

Honestly, seems _really_ unlikely since to be a problem it would require a CNI ADD occurring in between the retries. But in theory possible?

caseydavenport: I guess in KDD mode we're protected by the deletiontimestamp and the fact that we don't rely on the annotations (always fall back to the status fields), and in etcd mode this isn't an issue. So probably fine.

fasaxc: ErrorResourceUpdateConflict means the revision was wrong; not the UID.  If the UID is wrong then we get ErrorResourceDoesNotExist

PR #930, opened 2020-08-18
Fix that CNI plugin itself was only being included if make ut was run

fasaxc:
## Description

The upgrade-ipam init container still runs `calico-ipam` but if you do a clean build with `make clean image` then that file doesn't get included.

After a bit of head scratching, I found that the `ut` target was copying that file into place so Semaphore's `make ci` included it in the image whereas `make image` doesn't.

Possible that we want to fix the manifest instead but thought it made sense to make the build sane if that file is meant to be included.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ update-pins: update-libcalico-pin
 # Building the binary
 ###############################################################################
 BIN=bin/$(ARCH)
-build: $(BIN)/install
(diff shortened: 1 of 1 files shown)


fasaxc:
Rebased to pick up Song's fix.

PR #942, opened 2020-08-28
Revert accidental change from KUBERNETES_NODE_NAME to NODENAME

neiljerram:
For the hostname that we put into CNI config, we used to allow
KUBERNETES_NODE_NAME to override the system value, and
https://github.com/projectcalico/cni-plugin/pull/914 accidentally
changed that to NODENAME.  Here we revert that.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport:
Currently this is only targeting v3.17.0 - will need to be backported if we want to put this in a patch release.

caseydavenport:
@neiljerram are you planning on doing that?

neiljerram:
@caseydavenport Cherry-pick is here: https://github.com/projectcalico/cni-plugin/pull/943

Can you help with describing the impact of this, and so if it needs a release note?  I'm afraid I don't understand the chain of config here.

bootc:
As the reporter of this problem (in Slack: https://calicousers.slack.com/archives/C0BCA117T/p1598602400045300) the impact of this change as I see it is:

> Resolves an issue on nodes whose Kubernetes node name does not exactly match the system hostname

In effect, without this change, on any node where the Kubernetes node name (`kubectl get node`) is not exactly the same as the system hostname (`hostname` or `hostnamectl status`), pods fail to start entirely

PR #943, opened 2020-09-01
Merge pull request #942 from neiljerram/fix-node-name-var

neiljerram:
Revert accidental change from KUBERNETES_NODE_NAME to NODENAME

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


jhughes2112:
Question, was this the reason why (on AWS, for example) if the $HOSTNAME would return "ip-x-y-z-w" but the node name was joined as "ip-x-y-z-w.ec2.internal", nodes would fail partway through joining?  I had to work around this by forcing the short-name of AWS boxes to be the FQDN.  Can I revert that if I update?

bootc:
Yes, that sounds like the issue I ran into: not on AWS but the symptoms are similar. You'll find that pods are created and scheduled on the node but can't actually start. This only affects Calico 3.16.0 though; 3.15 and lower were not affected, and fixed in 3.16.1.

PR #948, opened 2020-09-04
Fix installation of etcd secrets into config file

caseydavenport:
## Description

We were comparing against the container filesystem, when we should have
been comparing against the host filesystem.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix population of etcd certificates in CNI config
```

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -353,23 +353,22 @@ func writeCNIConfig(c config) {
 
 	// Replace etcd datastore variables.
 	hostSecretsDir := c.CNINetDir + "/calico-tls"
-
-	etcdCertFile := fmt.Sprintf("%s/etcd-cert", hostSecretsDir)
-	if fileExists(etcdCertFile) {
+	if fileExists("/host/etc/cni/net.d/calico-tls/etcd-cert") {
+		etcdCertFile := fmt.Sprintf("%s/etcd-cert", hostSecretsDir)
 		netconf = strings.Replace(netconf, "__ETCD_CERT_FILE__", etcdCertFile, -1)
 	} else {
 		netconf = strings.Replace(netconf, "__ETCD_CERT_FILE__", "", -1)
 	}
 
-	etcdCACertFile := fmt.Sprintf("%s/etcd-ca", hostSecretsDir)
-	if fileExists(etcdCACertFile) {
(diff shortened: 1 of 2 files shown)


pkg/install/install.go
```diff
@@ -353,23 +353,22 @@ func writeCNIConfig(c config) {
 
 	// Replace etcd datastore variables.
 	hostSecretsDir := c.CNINetDir + "/calico-tls"
-
-	etcdCertFile := fmt.Sprintf("%s/etcd-cert", hostSecretsDir)
-	if fileExists(etcdCertFile) {
+	if fileExists("/host/etc/cni/net.d/calico-tls/etcd-cert") {
```
Apentacon, line 356: Is it OK to hard-code the certificate location? I think the correct question to ask here would be why line355 hostSecretsDir doesn't get correct directory in the first place? Also it wouldn't make sense to then leave the next line357 if you're already hard-coding cert location in the "if" part?

caseydavenport: In this case, I believe it is ok to hard code, unless I'm missing something. 

This code is a bit confusing because it's dealing with two different filesystem contexts. `hostSecretsDir` is the directory as understood by the host (e.g., `/etc/cni/net.d/calico-tls`). However, this code runs inside the CNI plugin container, which has that host directory mounted at `/host/etc/cni/net.d/calico-tls` no matter what the actual directory is on the host.

The bug here was that we were using `hostSecretsDir` in code that was running within the container's filesystem context, which is wrong because `hostSecretsDir` is only valid from the host's perspective. 

Here's another example of where we're hard coding it: https://github.com/projectcalico/cni-plugin/blob/4f3b7d3c73f67153eedbc4961df88db55bf649f8/pkg/install/install.go#L161

mgleung (approved):
LGTM

pkg/install/install.go
caseydavenport: Basically, locations within the container are always the same, because the mount point is always the same.. Locations on the host filesystem can be configured and so need to be variables.

In this case, we're hardcoding the filepath within the container, and using a variable for the host-side location, so it should be OK.

Apentacon: Thanks for explanation!

PR #955, opened 2020-09-23
Configure IPAM owner references via UID

caseydavenport:
## Description

Follow on from this PR: https://github.com/projectcalico/libcalico-go/pull/1303

Set the UID field based on the pod's UID so that IPAM allocations can
be garbage collected when the pod object is deleted.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/ipamplugin/ipam_plugin.go
+++ b/pkg/ipamplugin/ipam_plugin.go
@@ -109,7 +109,8 @@ func Main(version string) {
 
 type ipamArgs struct {
 	cnitypes.CommonArgs
-	IP net.IP `json:"ip,omitempty"`
+	IP             net.IP                        `json:"ip,omitempty"`
+	CALICO_POD_UID cnitypes.UnmarshallableString `json:"podUID,omitempty"`
 }
 
 func cmdAdd(args *skel.CmdArgs) error {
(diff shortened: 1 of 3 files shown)


pkg/ipamplugin/ipam_plugin.go
```diff
@@ -109,7 +109,8 @@ func Main(version string) {
 
 type ipamArgs struct {
 	cnitypes.CommonArgs
-	IP net.IP `json:"ip,omitempty"`
+	IP             net.IP                        `json:"ip,omitempty"`
+	CALICO_POD_UID cnitypes.UnmarshallableString `json:"podUID,omitempty"`
```
fasaxc, line 113: Do you have to use the non-standard naming convention here?

fasaxc (commented):
Looking good so far

pkg/ipamplugin/ipam_plugin.go
caseydavenport: Yeah, the code that parses this looks for env vars that match the field names in the struct.

I suppose I could either:
- Be consistent with the other env vars (what this PR does currently)
- Be consistent with the other go struct names (making the env var inconsistent)

I don't really have a strong preference either way.

PR #967, opened 2020-11-03
Fix error message in IP assignment which references the field it failed to set

mgleung:
## Description

Fixes https://github.com/projectcalico/calico/issues/3749 . This was causing a nil pointer exception since the BGP spec may not exist if the tunnel IP assignment fails.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fixes a bug that caused IPAM assignment to throw a nil pointer exception in cases where tunnel address IP assignment failed.
```


caseydavenport (approved):
ez

Issue #928, opened 2020-08-17
IPAM upgrade fails to unlock data store


caseydavenport:
Discovered in this Slack thread: https://calicousers.slack.com/archives/CPTH1KS00/p1597440347096800?thread_ts=1597088571.064200&cid=CPTH1KS00

> I see a logic flaw: the Migrate function removes the telltale directory before toggling the data store status back to true. Next time that function runs, the directory is gone, so it exits early without setting the data store status back to true.

## Expected Behavior

The upgrade-ipam script always sets the data store ready if it sets it to false.

## Current Behavior

Possible that the cluster never leaves `datastoreReady: false`.

## Possible Solution

- Just stop messing with the ready flag altogether - I am not sure it is actually working the way we intended for this migration.
- Make sure we only delete the host-local IPAM directory after we unlock the data store, so that if we fail we will try again.

seh:
Another possible solution is to set the ready flag before attempting to delete the sentinel directory tree. However, I'm not sure if _upgrade-ipam_ would handle things correctly if it started again, found that directory there, but all the migration steps apart from deleting the directory turned out to be no-ops. In other words, would _upgrade-ipam_ need to check whether it actually needs to do anything with these files, or would it just try to migrate them and deal with no changes occurring (because these steps already succeeded in a previous run). Of course, it may _never_ be able to delete the directory tree; if it fails the first time, it was probably due to lack of permission that won't have changed by the next attempt.

PR #968, opened 2020-11-03
Fix a bug that prevented failed attempts at unlocking the datastore d…

mgleung:
…uring IPAM migration from retrying

## Description

Fixes https://github.com/projectcalico/cni-plugin/issues/928 . This reorders the logic so that failures to set the datastore readiness will not delete the host-local IPAM directory whose presence allows for the logic to be retried again.

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Fixes a bug that prevented failed attempts at unlocking the datastore during IPAM migration from retrying.
```

### DIFF
--- a/pkg/upgrade/migrate.go
+++ b/pkg/upgrade/migrate.go
@@ -338,25 +338,28 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
 		log.Info("successfully released lock on host-local backend!")
 	}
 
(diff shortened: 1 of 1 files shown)


mgleung:
Tested this upgrading an installation from host-local IPAM to Calico IPAM with both the `datastoreReady` set to `true` and `false`.

PR #966, opened 2020-11-03
Fix issue moving files into place during install

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

PR #969, opened 2020-11-03
Automated cherry pick of #966: Fix issue moving files into place during install

caseydavenport:
Cherry pick of #966 on release-v3.16.

#966: Fix issue moving files into place during install

```release-note
Fix that installing CNI plugin would fail if binaries were in use
```


bcreane (commented):
looks good (to me). I'll let one of the core-maintainers weigh in for approval. thanks for addressing this so quickly. 🐬

PR #985, opened 2020-11-20
Add a file lock to serialise IPAM allocations.

fasaxc:
## Description

Acquire a best-effort host-wide lock to prevent multiple copies of the CNI plugin trying to assign
concurrently. AutoAssign is concurrency safe already but serialising the CNI plugins means that
we only attempt one IPAM claim at a time on the host's active IPAM block.  This reduces the load
on the API server by a factor of the number of concurrent requests.

## Todos
- [x] Set windows path for file lock
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Serialise IPAM updates using a file lock on disk in order to reduce load on the API server.  Concurrent updates are almost always to the same block so they only contend  rather than improving scale.
```


halfcrazy:
What about cni-plugin using a socket to connect to a daemon inside calico-node for ipam per host or per cluster.

fasaxc:
@halfcrazy that's a large re-architecture job.  It could be even better in performance than this approach since it could batch up requests and do them in one datastore update.  However, it's a lot more complex, especially the failure scenarios (block gets written but we fail to respond to some of the waiting CNI plugins, for example).  So, let's bank this increment and do the re-arch if it's shown to be needed later.

tghartland:
I tested this in the same setup I described in the slack, with a 2 core master node that can't handle the API load of creating many pods at once with clusterloader2.

Clusterloader2 is creating a deployment with a calculated number of replicas to ensure a set number of pods per node. It then measures for each pod the time from pod creation to when it first sees the pod as ready. (For reference, [this](https://user-images.githubusercontent.com/11710676/99823742-0883c500-2b55-11eb-86df-448ab4194705.png) is what I put in slack to show the effect compared to a larger master node that is able to handle the load).

I ran through my test of varying the number of pods created by clusterloader2, and then patched calico-node with the new CNI image, restarted the kubelets and ran through the test again.
The result:
lock_patch
There is a small improvement for the 50th percentile, but the 90th and 99th percentiles are kept in check much better.

The logging patches I had already added included one to log the number of retries and the total duration of [this loop](https://github.com/projectcalico/libcalico-go/blob/ddbe97145f70224b7b1e7a7b272d53876ca5cea7/lib/ipam/ipam.go#L379-L413).

With no lock, the number of retries and the time in the loop increase like this: https://gist.github.com/tghartland/eef5e1254cca52ab6b45edbae10ead29 (note: these logs are from another cluster, the times might not be directly comparable)

With this lock patch, there are no retries and so the time in the loop is always the time of a single attempt: https://gist.github.com/tghartland/c8177cd58e05a919fedc7e9a5f888e3d
It does seem like the reduced API load is improving the response time.

I'll try next week to figure out from the API server logs how many requests it is actually preventing, and how the response time is affected.
I'll also test this in a cluster with a larger master node and check that it isn't having any other side effects when the API server isn't a bottleneck.

pkg/ipamplugin/ipam_plugin.go
```diff
@@ -251,7 +255,16 @@ func cmdAdd(args *skel.CmdArgs) error {
 			assignArgs.HostReservedAttrIPv4s = rsvdAttrWindows
 		}
 		logger.WithField("assignArgs", assignArgs).Info("Auto assigning IP")
-		assignedV4, assignedV6, err := calicoClient.IPAM().AutoAssign(ctx, assignArgs)
+		autoAssignWithLock := func(calicoClient client.Interface, ctx context.Context, assignArgs ipam.AutoAssignArgs) ([]cnet.IPNet, []cnet.IPNet, error) {
```
caseydavenport, line 258: Think we'll also want a lock around the AssignIP call here: https://github.com/projectcalico/cni-plugin/pull/985/files#diff-2c8ea97e5c1bc5a024e2780d4b842e2b444124cf1e9670e18d91100405c7b4ddR179

fasaxc:
@tghartland Thanks for the analysis.  My back-of-envelope calculation is that

* without the patch: the number of API calls is something like `n*(n-1)` where `n` is the average number of concurrent CNI invocations
* with the patch, it should be exactly `n`

In a system with little load, the lock should always be free and there should be almost zero impact.

pkg/ipamplugin/ipam_plugin.go
```diff
@@ -285,6 +303,38 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	if path == "" {
+		path = ipamLockPath
+	}
+	err := os.MkdirAll(filepath.Dir(path), 0777)
+	if err != nil {
+		logrus.WithError(err).Error("Failed to make directory for IPAM lock")
+		// Fall through, still a slight chance the file is there for us to access.
+	}
+	ipamLock := flock.New(path)
+	err = ipamLock.Lock()
```
caseydavenport, line 322: If the lock is held forever (e.g., due to an error releasing the lock in the instance that is holding it) then I think we'll just block IPAM allocations on this node forever. Is that right?

Maybe we need some way to timeout held locks if a particular lock isn't released within a reasonable amount of time?

fasaxc: When the process exits, the lock will be released; a dead process can't hold a file lock.  I believe that we time out the CNI plugin with a panic in a background goroutine so that we can't block forever?

tghartland:
I did some processing of the API server logs from while there is a clusterloader2 test running and the API server becomes overloaded.

* 2 core master node
* 20 worker nodes
* creating 20 pods per node

Even in the image that I built for the CNI without the lock patch, I did include one change which was to set the kubernetes client user agent so that I could filter for those requests in the logs.

```diff
  // libcalico-go/lib/backend/k8s/k8s.go
  config.Burst = 100
+ config.UserAgent = "libcalico"
  cs, err := kubernetes.NewForConfig(config)
```

The logs then contain lines like
```
verb="GET" URI="/apis/crd.projectcalico.org/v1/ippools" latency=1.351026ms resp=200 UserAgent="libcalico" srcIP="..."
```
I'm assuming that the latency in this log is the time that the API server is measuring between receiving a request, and sending a response. I'll be plotting with this value, as well as the overall request rate.

In these plots there are two different parts to the load, the first is pods being created and then after a short idle time, there is another surge of requests for deleting those pods.

With no lock, the majority of requests to the API server comes from calico-ipam while creating pods. The request latency increases to around 400ms.
apilogs-nopatch

With the lock patch, only around half of the requests come from calico-ipam.
apilogs-patched

The request latency is still increased, but only to 200ms. (I tried the same thing with a 16 core master node, and it was able to keep the request latency at ~5ms even without the lock patch).

Looking at the plots, at the peak there is a difference of around 200 requests per second from calico during the pod creation.
The total number of requests sent from calico-ipam over the full duration is about 20,000 without the patch, and 10,000 with the patch.

caseydavenport (approved):
JGOC

PR #1008, opened 2021-01-07
Fix that we were returning an error when we shouldn't

caseydavenport:
## Description

Fixes https://github.com/projectcalico/calico/issues/4084

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
Fix bug that prevented pods from being torn down in some situations
```


mgleung (approved):
LGTM

Issue #990, opened 2020-12-08
ClusterInformation.datastoreReady stays false for k8s cluster with host-local ipam after node reboot 


jzhoucliqr:
With latest v3.17 release, for k8s cluster with `host-local` ipam-type, after reboot, the ClusterInformation.datastoreReady stays false never turns to true.

I think it might be caused by https://github.com/projectcalico/cni-plugin/pull/968/files, where below condition is not met. 
```
calicoDS.Status.UpdatedNumberScheduled == calicoDS.Status.DesiredNumberScheduled-calicoDS.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.IntVal
``` 

For a single node cluster case,  all 3 numbers are 1. Same happens to multi-node clusters. 

```
    updateStrategy:
      rollingUpdate:
        maxUnavailable: 1
      type: RollingUpdate
  status:
    currentNumberScheduled: 1
    desiredNumberScheduled: 1
    numberAvailable: 1
    numberMisscheduled: 0
    numberReady: 1
    observedGeneration: 1
    updatedNumberScheduled: 1
```

## Steps to Reproduce (for bugs)

1. setup a single node cluster with host-local ipam type
2. reboot the node
3.
4.

## Your Environment

* CNI plugin version (e.g. `/opt/cni/bin/calico --version`): v3.17.0
* Orchestrator version (e.g. kubernetes, mesos, rkt):
* CNI config:
* Operating System and version:
* Link to your project (optional):

caseydavenport:
@jzhoucliqr yeah, I agree this sounds likely to be a result of that change. We'll target a fix for a v3.17 patch release.

PR #1013, opened 2021-02-05
Fix datastore lock getting stuck during ipam migration

mgleung:
## Description

Removes the faulty logic that caused the datastore getting stuck as being locked since it is possible for multiple nodes to become ready between the unlocking logic's polling interval. Now, the datastore should become locked whenever a node begins its migration. This leaves a very small possibility in very specific cases that borrowed IPs may overlap between host-local and calico-ipam assignments but these cases would be fixed the next time that the pod restarts.

Fixes https://github.com/projectcalico/cni-plugin/issues/990

Verified this case works by setting up a cluster where the `maxUnavailable` number is set to the number of nodes and seeing that IPAM migration will set datastore readiness back to `true`

## Todos
- [x] Tests
- [x] Documentation
- [x] Release note

## Release Note

```release-note
Fix a bug that prevented datastore readiness from being reset during IPAM migration.
```

### DIFF
--- a/pkg/upgrade/migrate.go
+++ b/pkg/upgrade/migrate.go
@@ -344,7 +344,7 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
 	calicoDS, err := k8sClient.AppsV1().DaemonSets("kube-system").Get(context.Background(), "calico-node", metav1.GetOptions{})
 	if err != nil {
 		log.WithError(err).Error("failed to retrieve Calico daemonset to check if datastore readiness should be set to true")
-	} else if calicoDS.Status.UpdatedNumberScheduled == calicoDS.Status.DesiredNumberScheduled-calicoDS.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.IntVal {
+	} else if calicoDS.Status.UpdatedNumberScheduled >= calicoDS.Status.DesiredNumberScheduled-calicoDS.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable.IntVal {
 		log.Info("setting Calico datastore readiness to true...")
 		t := true
 		clusterInfo.Spec.DatastoreReady = &t
--- a/pkg/upgrade/migrate.go
+++ b/pkg/upgrade/migrate.go
@@ -338,23 +338,20 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
 		log.Info("successfully released lock on host-local backend!")
 	}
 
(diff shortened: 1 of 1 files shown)


mgleung:
Unofficial approval from @doublek: `Change looks good`

caseydavenport (approved):

pkg/upgrade/migrate.go
```diff
@@ -338,23 +338,20 @@ func Migrate(ctxt context.Context, c client.Interface, nodename string) error {
...
+	// that is already in use by host-local. Any pods stuck in this situation will be fixed once they are restarted.
+	// TODO: Re-enable datastoreReady if all of the nodes have been picked up for the rolling update.
+	// This logic has not been used because of the potential length of time it would bring normal operation
+	// of the cluster down for by locking the datastore until the update has been applied to all nodes.
+	// calicoDS, err := k8sClient.AppsV1().DaemonSets("kube-system").Get(context.Background(), "calico-node", metav1.GetOptions{})
+	// if calicoDS.Status.UpdatedNumberScheduled == calicoDS.Status.DesiredNumberScheduled {
+	log.Info("setting Calico datastore readiness to true...")
+	t := true
+	clusterInfo.Spec.DatastoreReady = &t
+	if _, err = c.ClusterInformation().Update(ctxt, clusterInfo, options.SetOptions{}); err != nil {
```
seh, line 351: Did you consider using a patch operation here instead of an update? This update can fail if another Kubernetes client changed the _ClusterInformation_ object since we acquired our copy of it (in the `clusterInfo` variable here). Ideally, we'd have our setting this value back to true succeed regardless of intervening changes from others. Patching would allow us that unconditional success.

mgleung: That's a good point and probably a cleaner way to do this than we have now. Right now, if we hit the error updating since the `ResourceVersion` is no longer the most current in the case you've outlined, we'll end up retrying the entire migration again until it works and passes (which isn't nearly as nice and clean as you've outlined with `Patch` https://github.com/projectcalico/cni-plugin/blob/master/pkg/ipamplugin/ipam_plugin.go#L97-L104). So I would expect us to work in that case if I'm understanding it correctly, it's just not nearly as nice or efficient as it could be. 

Also, it appears we haven't implemented `Patch` yet for any of our resource APIs (`ClusterInfo` client for reference https://github.com/projectcalico/libcalico-go/blob/master/lib/clientv3/clusterinfo.go) so we would need to update that first before we could use it. Definitely something we should be doing in the future, but for now, in order to get this out ASAP, we'll be skipping it.

PR #1074, opened 2021-05-07
Simplify cni installation

ScheererJ:
As the `calico` and `calico-ipam` are included in the container image next to the `install` binary, there is no need to copy them in the container.
It is also not necessary to have special logic to copy the `install` binary instead of the `calico` and `calico-ipam` binaries.
The resulting installation logic is easier to understand.

## Description

This pull request makes the `Install()` function easier to understand by removing unnecessary/redundant operations. It becomes easier to reason about what is happening when unnecessary operations are no longer present.

I tested the fix manually with manually build container images.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -169,14 +169,6 @@ func Install() error {
 		}
 	}
 
-	// Copy install to calico and calico-ipam
-	if err := copyFileAndPermissions("/opt/cni/bin/install", "/opt/cni/bin/calico"); err != nil {
-		logrus.WithError(err).Fatalf("Failed to copy install to calico")
-	}
-	if err := copyFileAndPermissions("/opt/cni/bin/install", "/opt/cni/bin/calico-ipam"); err != nil {
-		logrus.WithError(err).Fatalf("Failed to copy install to calico-ipam")
-	}
-
 	// Place the new binaries if the directory is writeable.
 	dirs := []string{"/host/opt/cni/bin", "/host/secondary-bin-dir"}
 	for _, d := range dirs {
@@ -193,11 +185,6 @@ func Install() error {
 		for _, binary := range files {
 			target := fmt.Sprintf("%s/%s", d, binary.Name())
 			source := fmt.Sprintf("/opt/cni/bin/%s", binary.Name())
-			if strings.Contains(binary.Name(), "calico") {
-				// For Calico binaries, we copy over the install binary. It includes the code
-				// for each, and the name of the binary determined which is executed.
-				source = "/opt/cni/bin/install"
-			}
 			if c.skipBinary(binary.Name()) {
 				continue
 			}


caseydavenport (changes requested):
Hm, I think the way this was intended to work was that we'd only have a single copy of the binary located at `install` and that we wouldn't include essentially the same code but in three differently named binaries (install, calico, calico-ipam). 

I think the correct fix is to remove the `calico` and `calico-ipam` binaries from the image and keep the copy logic here. The idea is that we only need a single copy of the binary and thus a smaller resulting image size.

ScheererJ:
Fair point. So you are saying that lines 196-200 should stay as is and not be removed, but instead the binaries should not make it into the container image in the first place. Right?

Could you please clarify why the first copy operations (lines 172-179) would still be required, though? It just copies the binaries inside the container, which seems rather useless when lines 196-200 stay anyway.

caseydavenport:
Ah! Sorry for the delay on this @ScheererJ - it got lost in my vast ocean of GitHub notifications.

So, looking closer at this code, it's definitely wonky. The way it is written now we definitely need lines 172-179, because the loop below uses the contents of `/opt/cni/bin` within the container to decide which binaries to copy to the host: https://github.com/projectcalico/cni-plugin/pull/1074/files#diff-3c83c1219d23236aa4c92d672faa69bb463e2001adb87833ca0f7434faded420R185

I think if anything, we might just need to drop lines 196-200. Then, the code will start by copying the `install` binary to `calico` and `calico-ipam` so the following loop will detect both and copy them to the host. 

Still a bit awkward.

ScheererJ:
@caseydavenport I am still a bit confused. Maybe you can enlighten me ;-) .

The cni container image contains all three binaries `calico`, `calico-ipam` and `install`, which are essentially the same. In https://github.com/projectcalico/cni-plugin/pull/1074#pullrequestreview-664599865 you hint to that removing `calico` and `calico-ipam` binaries from the container image would be the right way forward. Should I adapt the pull request to do so?

Staying with the status quo, i.e. all three binaries are included in the container image, lines 172-179 are somewhat strange as they simply overwrite the existing files in the running container. From my perspective, this would be useless iff the container image stays as is (with the three calico binaries).

Moving to the last part, lines 196-200 again seem strange iff either the three binaries stay in the container image or lines 172-179 stay in place. If lines 196-200 stay there is no need to have the three binaries in the container image and there is also no need for lines 172-179 as the `install` binary is anyway used as the source for the copy operation.

The only reasoning for keeping at least parts of it would be that you prefer the files to be hard-linked (not copied). Still, some of the steps would be redundant.

Personally, I would simply prefer to remove lines 172-179 and 196-200 to make the code easier to understand. I see the following alternatives:
- Keep all three binaries in the container image, but remove lines 172-179 <= my preference
- Only have the `install` binary in the container image, remove lines 172-179, but keep lines 196-200
- Only have the `install` binary in the container image, keep lines 172-179, but remove lines 196-200

Please comment as you are the expert.

caseydavenport:
Thanks for the thorough analysis - my comment was a bit hasty and on re-read I can see how it's confusing. 

I am happy to go with your preference on this PR, as I think it is strictly better than what exists today. Namely, it maintains the behavior of having all three binaries in the image, but removes redundant and unnecessary code from the install logic.

A further improvement on top of that, in my opinion, would be to remove the `calico` and `calico-ipam` binaries from the image, and adapt the install logic to explicitly copy them to the host using `install` as the source file. This will effectively reduce the size of the image to 1/3 of what it is today, which is a nice win and how the install logic was intended to work from the get-go. However, I think we should leave that out of this PR for now and bank the easy win of simplifying the logic first.

caseydavenport (approved):

caseydavenport:
Given all that, I think this PR is doing the right thing:
- Leaving the binaries in the image
- Removing the (unnecessary) logic to copy install -> calico, and install -> calico-ipam inside the image.
- Removing the (unnecessary) logic to copy install -> calico, and install -> calico-ipam on the host.

PR #1103, opened 2021-07-08
 Update call to AutoAssign() to use IPAMAssignments struct

coutinhop:
…ails on errors

## Description

Manually update libcalico-go pin and update call to AutoAssign() to use IPAMAssignments struct and output more details on errors

## Todos
- [x] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Added more detailed output of IP assignment errors
```


pkg/ipamplugin/ipam_plugin.go
```diff
@@ -272,19 +272,19 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 		if err != nil {
 			return err
 		}
+		logger.Infof("Calico CNI IPAM assigned addresses IPv4=%v IPv6=%v", v4Assignments, v6Assignments)
 
 		// Check if IPv4 address assignment fails but IPv6 address assignment succeeds. Release IPs for the successful IPv6 address assignment.
-		if num4 == 1 && len(assignedV4) != num4 {
-			if num6 == 1 && len(assignedV6) != 0 {
-				logger.Infof("Assigned IPv6 addresses but failed to assign IPv4 addresses. Releasing %d IPv6 addresses", len(assignedV6))
+		if num4 == 1 && v4Assignments != nil && len(v4Assignments.IPs) < v4Assignments.NumRequested {
```
mgleung, line 282: nit: I think if we're going to do comparisons against `v4Assignments.NumRequested` (which is the same thing as `num4`), we should do that comparison throughout the entire statement. Essentially all we're doing here is checking if we asked for an IP to be assigned and if it doesn't happen, then we need to error.

It might actually be easier to keep the comparison against `num4` just because the argument is used earlier in the call. If we do the comparison against `v4Assignments.NumRequested`, we'll need to make sure that that is assigned correctly in the library call that creates it whereas it's easier to track in the code what `num4` is without needing to cross check libcalico-go.

pkg/ipamplugin/ipam_plugin.go
```diff
@@ -311,21 +311,27 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 		if num4 == 1 {
-			if len(assignedV4) != num4 {
-				return fmt.Errorf("failed to request %d IPv4 addresses. IPAM allocated only %d", num4, len(assignedV4))
+			if err := v4Assignments.PartialFulfillmentError(); err != nil {
+				return fmt.Errorf("failed to request IPv4 addresses: %w", err)
 			}
-			ipV4Network := net.IPNet{IP: assignedV4[0].IP, Mask: assignedV4[0].Mask}
+			if len(v4Assignments.IPs) < num4 {
+				return fmt.Errorf("failed to request %d IPv4 addresses. IPAM allocated only %d", num4, len(v4Assignments.IPs))
+			}
```
mgleung, line 319: will we ever get to this case now? I thought the `PartialFullfillmentError` is only populated if this same condition was hit?

pkg/ipamplugin/ipam_plugin.go
coutinhop: Got it. Yeah, the tests in libcalico should assure that NumRequested is correct, but I was initially thinking it would be better to use it instead of num4/num6. But we'll need to use them, so I see your point in using num4/num6 everywhere. I'll make the changes.

pkg/ipamplugin/ipam_plugin.go
coutinhop: Right, I don't think we should, but I was trying to be safe. (Is that "too safe"? Should I remove it then?)

coutinhop: Will remove it!

pkg/ipamplugin/ipam_plugin.go
```diff
@@ -272,19 +272,19 @@ func cmdAdd(args *skel.CmdArgs) error {
 			defer unlock()
 			return calicoClient.IPAM().AutoAssign(ctx, assignArgs)
 		}
-		assignedV4, assignedV6, err := autoAssignWithLock(calicoClient, ctx, assignArgs)
-		logger.Infof("Calico CNI IPAM assigned addresses IPv4=%v IPv6=%v", assignedV4, assignedV6)
```
mgleung, line 276: I think it might be nice to keep this

coutinhop: I just moved it down to after the err check because v4Assignments/v6Assignments can be nil in this case, but I see it doesn't make sense because I later changed the logging to print the whole struct instead of just the IPs slice inside it. Will move it back, thanks!

mgleung (approved):
LGTM

PR #1104, opened 2021-07-12
windows: update EnsureVXLANTunnelAddr so we can call it from node

lmm:
## Description

This PR updates EnsureVXLANTunnelAddr so it can be called from node during startup.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


song-jiang (approved):
LGTM

PR #1116, opened 2021-07-15
create multi queue veth for xdp

arikachen:
Signed-off-by: chenqijun <chenqijun@corp.netease.com>

## Description

Now we use vpp + vcl to accelerate apps in containers, using xdp/afxdp to redirect packet from veth to vpp. 
vpp has more than one workers to process packet in our situation,  
only one vpp worker works when veth has one queue, it affect some performance.

It depend more recent kernel support.

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/dataplane/linux/dataplane_linux.go
+++ b/pkg/dataplane/linux/dataplane_linux.go
@@ -38,6 +38,7 @@ import (
 type linuxDataplane struct {
 	allowIPForwarding bool
 	mtu               int
+	queues            int
 	logger            *logrus.Entry
 }
 
@@ -45,6 +46,7 @@ func NewLinuxDataplane(conf types.NetConf, logger *logrus.Entry) *linuxDataplane
 	return &linuxDataplane{
 		allowIPForwarding: conf.ContainerSettings.AllowIPForwarding,
 		mtu:               conf.MTU,
+		queues:            conf.NumQueues,
 		logger:            logger,
 	}
 }
@@ -76,8 +78,10 @@ func (d *linuxDataplane) DoNetworking(
 	err = ns.WithNetNSPath(args.Netns, func(hostNS ns.NetNS) error {
 		veth := &netlink.Veth{
 			LinkAttrs: netlink.LinkAttrs{
-				Name: contVethName,
-				MTU:  d.mtu,
+				Name:        contVethName,
+				MTU:         d.mtu,
+				NumTxQueues: d.queues,
+				NumRxQueues: d.queues,
 			},
 			PeerName: hostVethName,
 		}
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -80,7 +80,7 @@ type NetConf struct {
 	} `json:"ipam,omitempty"`
 	Args                 Args                   `json:"args"`
 	MTU                  int                    `json:"mtu"`
-	NumQueues            int                    `json:"n_queues"`
+	NumQueues            int                    `json:"num_queues"`
 	Nodename             string                 `json:"nodename"`
 	NodenameFile         string                 `json:"nodename_file"`
 	IPAMLockFile         string                 `json:"ipam_lock_file"`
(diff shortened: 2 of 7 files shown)


caseydavenport:
@arikachen thanks for the PR - could you add some context to the description explaining your use-case?

/sem-approve

pkg/dataplane/linux/dataplane_linux.go
```diff
@@ -76,8 +78,10 @@ func (d *linuxDataplane) DoNetworking(
 	err = ns.WithNetNSPath(args.Netns, func(hostNS ns.NetNS) error {
 		veth := &netlink.Veth{
 			LinkAttrs: netlink.LinkAttrs{
-				Name: contVethName,
-				MTU:  d.mtu,
+				Name:        contVethName,
+				MTU:         d.mtu,
+				NumTxQueues: d.queues,
+				NumRxQueues: d.queues,
```
neiljerram, line 84: What was the previous default for these fields, when we weren't setting them explicitly?

I guess it was 1, and so by default this PR doesn't change that, but adds the capability to do that by configuring `n_queues`.  Is that right?

neiljerram (commented):
Code looks good, but like @caseydavenport I'd also like more description of the motivation for this.  I also dropped one specific comment.

arikachen:
@caseydavenport @neiljerram 
thanks for you reply.  I update description in pr.

pkg/dataplane/linux/dataplane_linux.go
arikachen: yes

neiljerram:
Thanks @arikachen .  I guess the XDP -> VPP path only works for the ingress direction.  Does that mean we should only change NumRxQueues?

arikachen:
> Thanks @arikachen . I guess the XDP -> VPP path only works for the ingress direction. Does that mean we should only change NumRxQueues?

Yes, it is. but kernel limit the queue num with peer tx queue num.
```
if (dev->real_num_rx_queues < peer->real_num_tx_queues) {
            NL_SET_ERR_MSG_MOD(extack, "XDP expects number of rx queues not less than peer tx queues");
            err = -ENOSPC;
            goto err;
        }
```

arikachen:
@neiljerram Thanks.  I have test with increase rx queue,  only one rx queue works.
 ```
# ethtool -S eth0
NIC statistics:
     peer_ifindex: 351
     rx_queue_0_xdp_packets: 10570419
     rx_queue_0_xdp_bytes: 1526253129
     rx_queue_0_xdp_drops: 0
     rx_queue_1_xdp_packets: 0
     rx_queue_1_xdp_bytes: 0
     rx_queue_1_xdp_drops: 0
# ethtool -l eth0
Channel parameters for eth0:
Pre-set maximums:
RX:             2
TX:             1
Other:          0
Combined:       1
Current hardware settings:
RX:             2
TX:             1
Other:          0
Combined:       1
```

**Comment by neiljerram** [comment]
@arikachen Thanks for your last explanatory comment.  Let's see if we can get the CI to pass now...

**Review by neiljerram — APPROVED** [review]
LGTM - @caseydavenport do you have any concerns?

File: pkg/types/types.go [hunk 2]
```diff
@@ -80,6 +80,7 @@ type NetConf struct {
 	} `json:"ipam,omitempty"`
 	Args                 Args                   `json:"args"`
 	MTU                  int                    `json:"mtu"`
+	NumQueues            int                    `json:"n_queues"`
```
neiljerram, line 83: One of my colleagues has asked if we can change this JSON name to "num_queues", for a little more immediate clarity.

neiljerram (changes requested):
@arikachen Would you mind making one small naming change?  Then I think we can merge this.

arikachen:
> @arikachen Would you mind making one small naming change? Then I think we can merge this.

@neiljerram thanks.

neiljerram (approved):

PR #1119, opened 2021-07-26
Add HostPort support in the gRPC dataplane

AloysAugustin:
## Description
This PR adds support for the HostPort / HostIP fields in the gRPC dataplane.

This depends on and https://github.com/projectcalico/libcalico-go/pull/1471 to compile.

This PR doesn't include tests yet, but we'll add them if you agree on the approach 🙂 .

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```


caseydavenport (commented):
I don't think I have any problem adding this to the DoNetworking call - seems like a logical replacement of the `portmap` plugin to do this in CNI, so architecturally we have some evidence it works.

Either way we may also want to coordinate enhancing Felix to be capable of this in order to reconcile the rules if needed, but that is more of a nice-to-have than a requirement I think.

caseydavenport:
CC @mgleung - we'll need to figure out if we want to use this for the iptables implementation as well.

caseydavenport:
Whoops, didn't realize this didn't merge already!

PR #1146, opened 2021-08-20
fix cluster.server placeholder in writeKubeconfig

wangwang:
## Description

fix cluster.server placeholder in writeKubeconfig, remove unnessary square brackets or the server may like 'https://[169.254.0.1]:443'

## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -429,7 +429,7 @@ kind: Config
 clusters:
 - name: local
   cluster:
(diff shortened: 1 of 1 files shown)


caseydavenport:
@wangwang the square brackets are necessary for IPv6 - I think the existing code is correct. 

Is this in response to an issue you were seeing on your cluster?

PR #1148, opened 2021-08-21
Windows installer for Hostprocess

jsturtevant:
## Description

I've trying out different cni's with [Windows HostProcess containers](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/) and started to try out calico. Initially was working through powershell scripts but feel like using the same type of approach Linux uses is better.  

This is a really rough draft and requires a few things:
- [ ] yaml files from https://gist.github.com/jsturtevant/523368ba72abbbc253d13433b7eb32d6  
- [ ] updates to go-client https://github.com/kubernetes/kubernetes/pull/104490

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
Adds support for Windows HostProcess containers
```


Dockerfile-windows
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2015-2019 Tigera, Inc. All rights reserved.
```
jsturtevant, line 1: The dockerfile-windows is named this way due to the way the makefile loads arches based on docker files: https://github.com/projectcalico/go-build/blob/296c6f3239d6a1f42c691c5b09c9d36f0f617107/Makefile.common#L24

It was tricky to figure out what going on here. There is a probably a better way to do this?

Makefile
```diff
@@ -126,32 +131,43 @@ $(BIN)/install binary: $(SRC_FILES)
...
 	# Need amd64 builds tagged as :latest because Semaphore depends on that
 	docker tag $(CNI_PLUGIN_IMAGE):latest-$(ARCH) $(CNI_PLUGIN_IMAGE):latest
 endif
 	touch $@
 
-.PHONY: fetch-cni-bins
-fetch-cni-bins: $(BIN)/flannel $(BIN)/loopback $(BIN)/host-local $(BIN)/portmap $(BIN)/tuning $(BIN)/bandwidth
+$(DEPLOY_CONTAINER_MARKER_WINDOWS): Dockerfile-windows build 
+	docker buildx create --name img-builder --use --platform windows/amd64 
+	GO111MODULE=on docker buildx build --platform windows/amd64 --output=type=registry --pull -t $(CNI_PLUGIN_IMAGE):latest-$(ARCH)-windows --build-arg GIT_VERSION=$(GIT_VERSION) -f Dockerfile-windows .
```
jsturtevant, line 166: There might be a better way to build the windows image but I've used buildx to build on linux.  

One restriction is it requires pushing the image as to a registry at build time. There are a few other restrictions but the way it works now there aren't any issues.

Dockerfile-windows
lmm: Yeah... this is something we'll need to figure out. Apart from the dockerfile arches thing, right now we don't have a good way to build Windows images because Semaphore CI doesn't support Windows yet.

lmm:
Thanks for doing this @jsturtevant ! I'll try take a look in the next day or so.

(@song-jiang , fyi in case you didn't already see this.)

jsturtevant:
Thanks! This is really dependent on  kubernetes/kubernetes#104490 (I've put the basic changes on a branch and updated the mod file to test it here)

I also don't really like the way it needs to append the `CONTAINER_SANDBOX_MOUNT_POINT` to all the files and would be open to other solutions. We are exploring some other ways to handle this on the Windows side that might make it cleaner but it is several months away if we are successful. You can read a little more about the specifics here https://github.com/kubernetes/enhancements/pull/2865

Dockerfile-windows
jsturtevant: since these are mostly executable it's easy enough to build them on Linux using buildx but they can only be exported to a registry, not to the file system.

jsturtevant:
The init container is working, I've pushed an image to `jsturtevant/calico-cni:hostprocess-amd64-windows`. A nightly build of containerd and hcsshim is required as of now. We have one at https://github.com/kubernetes-sigs/sig-windows-tools/releases/tag/windows-containerd-nightly for convenience

k8s:

```
NAME                                                                  READY   STATUS                  RESTARTS           AGE
calico-node-windows-cl2ql                                             0/1     Pending                 0                  0s                                                               
calico-node-windows-cl2ql                                             0/1     Pending                 0                  0s                                                               
calico-node-windows-cl2ql                                             0/1     Init:0/1                0                  0s                                                               
calico-node-windows-cl2ql                                             0/1     Init:0/1                0                  13s                                                              
calico-node-windows-cl2ql                                             0/1     PodInitializing         0                  20s                                                              
calico-node-windows-cl2ql                                             1/1     Running                 0                  22s
```

output on the host where cni was installed. 
```
PS C:\Users\capi> cd C:\etc\cni\net.d\ 
PS C:\etc\cni\net.d> ls 


    Directory: C:\etc\cni\net.d


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----         9/1/2021   6:15 PM           1004 10-calico.conflist
-a----         9/1/2021   6:15 PM           2794 calico-kubeconfig


PS C:\etc\cni\net.d> cat .\10-calico.conflist 
{
  "name": "Calico",
  "windows_use_single_network": true, 

  "cniVersion": "0.3.1",
  "type": "calico",
  "mode": "vxlan",
  "nodename": "win-conta-7x9wh",      

  "vxlan_mac_prefix": "0E-2A",        
  "vxlan_vni": 4096,
  "mtu": 1350,
  "policy": {
    "type": "k8s"
  },

  "log_level": "info",

  "capabilities": {"dns": true},      
  "DNS":  {
    "Search":  [
      "svc.cluster.local"
    ]
  },

  "datastore_type": "kubernetes",

  "kubernetes": {
    "kubeconfig": "C:/C/c9bc9c6d94a20079a90740081e9f1678ca819fc28803e44c09fc008d7fb6ba66/etc/cni/net.d/calico-kubeconfig" 
  },

  "ipam": {
    "type": "calico-ipam",
    "subnet": "usePodCidr"
  },

  "policies":  [
    {
      "Name":  "EndpointPolicy",
      "Value":  {
        "Type":  "OutBoundNAT",
        "ExceptionList":  [
          "10.96.0.0/12"
        ]
      }
    },
    {
      "Name":  "EndpointPolicy",
      "Value":  {
        "Type":  "SDNROUTE",
        "DestinationPrefix":  "10.96.0.0/12", 
        "NeedEncap":  true
      }
    }
  ]
}
```

Issue #1165, opened 2021-09-17
Misleading logs


joy717:
https://github.com/projectcalico/cni-plugin/blob/f6cd1b86daec16b04e095b04abf0af3bffb73a89/pkg/k8s/k8s.go#L596

`len(unallocated) > 0` means that we do released Addrs. 

## Expected Behavior

should log correctly.

## Current Behavior

wrong logic for logging.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* CNI plugin version (e.g. `/opt/cni/bin/calico --version`):
* Orchestrator version (e.g. kubernetes, mesos, rkt):
* CNI config:
* Operating System and version:
* Link to your project (optional):

PR #1166, opened 2021-09-17
fix misleading logs

joy717:
## Description

fix misleading logs for [#1165](https://github.com/projectcalico/cni-plugin/issues/1165)

## Todos
- [ ] Tests
- [ ] Documentation
- [x] Release note

## Release Note

```release-note
None required
```


caseydavenport:
The existing logic is correct I believe. The `unallocated` slice returned by `ReleaseIPs` contains any IP addresses that we asked to release, but weren't actually allocated.

PR #1189, opened 2021-10-11
Use CGO to enable boringssl

caseydavenport:
## Todos
- [ ] Tests
- [ ] Documentation
- [ ] Release note

## Release Note

```release-note
None required
```

### DIFF
--- a/Dockerfile.arm64
+++ b/Dockerfile.arm64
@@ -1,19 +1,3 @@
-# Copyright (c) 2021 Tigera, Inc. All rights reserved.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -69,6 +69,14 @@ ifeq ($(BUILDARCH),amd64)
 	ETCD_CONTAINER=quay.io/coreos/etcd:$(ETCD_VERSION)
 endif
 
+# We need CGO to leverage Boring SSL.  However, the cross-compile doesn't support CGO yet.
+# Currently CGO can be enbaled in ARM64 and AMD64 builds.
```
neiljerram, line 73: nit: enbaled

neiljerram: Also, comment doesn't match the code, which is only for amd64.

neiljerram (changes requested):
Appears to be confusion here about arm64.  If we can't yet do CGO for arm64, probably best not to change the arm64 dockerfile at all.

PR #1217, opened 2021-11-24
change err site

Icarus9913:
Change the err handling site. 
Once we met the err, we should handle it firstly and not process other logical steps.


fasaxc (changes requested):
I don't think this change is right; the IPAM allocation can partially fail, returning some IPs and an error.  It's not an idiomatic Go API but it gives the caller the information they may need to clean up a messy failure.

